PR #1, opened 2015-08-05
Generate datatypes with sat-datatype, fix dependencies between projects

Duhemm:
Now it compiles on my machine, given a previously locally-published version of `ivyProj` that contains the changes introduced by sbt/sbt#2106. This is due to this repository containing the changes made to `ComponentCompiler`. These changes depend on changes made to `ivyProj`, which is not included in this repository, and that have not been published as of 0.13.9-M3. (see [here in the diff](https://github.com/sbt/incrementalcompiler/pull/1/files#diff-0ecdbc5a001d52fb34f5eafb7cd1aaa6R10))

What should I do about these changes? Roll them back until `ivyProj` get republished with the changes introduced in sbt/sbt#2106?


.gitignore
```diff
@@ -0,0 +1 @@
+target/
```
eed3si9n: Use the global gitignore.

PR #2, opened 2015-08-13
New testing framework for the incremental compiler

Duhemm:
The tests that are included in this PR are juste examples. I'll start porting some of the scripted tests if this new format LGTY!


internal/compile-inc/src/test/scala/sbt/inc/IncrementalCompilerTest.scala
```diff
@@ -0,0 +1,254 @@
...
+import java.io.File
+import scala.compat.Platform.EOL
+
+import sbt.IO
+import sbt.Path
+
+import xsbt.{ CompilationFailedException, TestAnalyzingCompiler }
+
+object IncrementalCompilerTest {
+  implicit class FileOP(val content: String)
```
jsuereth, line 21: So we lift strings into FileOPs?

jsuereth:
SO, to help be maven frinedly, I think we need to remove the `test->test` dependnecy.

Instead, create a new "incrementalCOmpilerTestFramework" project which has your test-harness code and then depend on that with `test->compile`. 

Otherwise, I love the new test DSL, looks great!

internal/compile-inc/src/test/scala/sbt/inc/IncrementalCompilerTest.scala
Duhemm, line 21: Yes, this allows us to write something like `"A.scala" -> delete` to express file deletion.

Duhemm:
@jsuereth Would it be enough to just change the dependency marked `test->test` to `test->compile`?

Duhemm:
@jsuereth @eed3si9n Any news on this PR?

jsuereth:
yeah, this LGTM assuming we enable TravisCI.

Duhemm:
I've rebased this PR and changed a few things regarding the splitting of the compiler interface. These new tests use classes defined in the compiler bridge. Because there were 2 projects, we would have needed to depend on either one based on the value of `scalaVersion`. Instead I merged the two projects in one, and made `scalaSource` depend on the value of  `scalaVersion`.

PR #3, opened 2015-08-27
Compiler interface logging (sbt/sbt#2177)

Duhemm:
Port of sbt/sbt#2177


jsuereth:
LGTM.

Again, cna we get a travisCI configuration before mergin though?

PR #4, opened 2015-08-27
Include private members in API hash of traits. (sbt/sbt#2160)

Duhemm:
This is a port of sbt/sbt#2160.

I've included both commits, particularly the second that simply added back overloads to restore the binary compatibility. It's not required here, should I remove it?


jsuereth:
LGTM.

This repos hould have travisCI running, but does not.  Please do not merge until TravisCI report success.

Duhemm:
Triggering Travis

eed3si9n:
Given that Travis is now enabled: https://travis-ci.org/sbt/incrementalcompiler/builds/78530783, this should be ok to merge.

PR #5, opened 2015-08-31
Update to sbt/io 1.0.0-M1, use scala.sys.process rather than sbt process, setup Travis

Duhemm:
Build for 2.11 will fail until sbt/util-*_2.11 gets released


dwijnand:
Travis boostrap.

Duhemm:
Most of the changes are formatting or fixes to warnings.
The important changes are in:
- build.sbt
- project/*
- JavacProcessLogger.scala (should there be overloads for the old methods?)
- AggressiveCompile.scala (same question)
- ScalaInstance.scala (I tried to get rid of some deprecated methods, but it looks like we need to keep some deprecated overloads)

PR #6, opened 2015-09-03
Split compiler interfaces

Duhemm:


interface/src/main/java/xsbti/compile/ScalaInstance.java
```diff
@@ -17,16 +17,16 @@
...
-	/**@deprecated Only `jars` can be reliably provided for modularized Scala. (Since 0.13.0) */
-	@Deprecated
+	/**deprecated Only `jars` can be reliably provided for modularized Scala. (Since 0.13.0) */
+	//Deprecated
 	File compilerJar();
 
-	/**@deprecated Only `jars` can be reliably provided for modularized Scala. (Since 0.13.0) */
-	@Deprecated
+	/**deprecated Only `jars` can be reliably provided for modularized Scala. (Since 0.13.0) */
+	//Deprecated
```
Duhemm, line 29: These have been de-deprecated only so that I can compile the rest using `-Xfatal-warnings`. They have been deprecated for a while but are still used internally.

Duhemm:
@eed3si9n I've rebased it. Is it okay to disable 2 tests for Scala pre-2.11 the way I did in https://github.com/Duhemm/incrementalcompiler/commit/3ea463b6e8e9efe08e31735a94495a05340b152b or should we simply not share the tests?

eed3si9n:
@Duhemm What you did looks fine.

PR #7, opened 2015-09-08
Update sbt modules, migrate to scalatest

Duhemm:
- Update sbt/util to 1.0.0-M3
- Update sbt/librarymanagement to 1.0.0-M2

Also, migrate the tests from specs2 to scalatest.


eed3si9n:
Awesome

eed3si9n:
Actually scalatest not fully converted. I should've reviewed more carefully..

gkossakowski:
What was the reason to moe from specs2 to scalatest?

eed3si9n:
ScalaTest seems to be more stable and depends less on community libraries.

gkossakowski:
Ok. I'll port my tests for class-based deps to scalatest.

PR #10, opened 2015-09-09
Retrieve version of compiler interface from properties file

Duhemm:
Previously we used the version of the component manager. However, these
two versions do not need to be in sync, which would make us try to
retrieve the wrong version of the compiler interface sources.


eed3si9n:
LGTM

PR #12, opened 2015-09-11
[Port of sbt/sbt#2197] Fixes #2196, Specify the compiler bridge module in a setting

Duhemm:
See sbt/sbt#2197

PR #13, opened 2015-09-11
Download "xsbti" component using Ivy

Duhemm:
The first four commits of this PR are the same as #12.

Previously we were relying on `componentID` to have it available. Because this `componentID` has been removed from the this project's build, sbt crashes when trying to compile the compiler interface.

This PR introduces logic to get `interface.jar` using Ivy. This way, we definitely get rid of `componentID`.


eed3si9n:
LGTM

PR #17, opened 2015-09-17
Adds object Doc 

eed3si9n:
This change adds `object Doc` to provide API point to call JavaDoc with Pickling-based caching.
So (d83279f021a4eb66528455d23974571665d9508a) is the main change.
This is intended to partially fix sbt/sbt#1614 (partially because I haven't implemented it for Scaladoc yet).

/review @Duhemm, @dwijnand, @jsuereth


jsuereth:
LGTM

PR #24, opened 2015-10-20
Configure classfile manager

Duhemm:
`ClassfileManager` is now an interface, and `xsbti.compile.Options` must provide a way to create a new `ClassfileManager`. This will allow users of the API provided by `sbt.internal.inc.IC` to specify the class file manager that they want to use.

`object ClassfileManager` is renamed to `DefaultClassfileManager`.

This change will help to port Scala IDE to this API (see https://github.com/scala-ide/scala-ide/blob/master/org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/EclipseSbtBuildManager.scala#L199-L207).


Duhemm:
Superseded by #25

PR #29, opened 2015-11-12
Get compiler bridge using Ivy in IncrementalCompilerSpec

Duhemm:


Duhemm:
The problem is that the compiler bridge will not get recompiled after being modified. I don't think that we want the compiler bridge to be recompiled every time that we run `IncrementalCompilerSpec`, but how to decide when to recompile it?

eed3si9n:
I am ok with recompiling the compiler bridge used by the unit test.

PR #30, opened 2015-11-13
Refactor incrementalcompiler API

eed3si9n:
sbt always had incrementalcompiler API in xsbti (Java interface) called
[`IncrementalCompiler<Analysis,
ScalaCompiler>`](https://github.com/sbt/incrementalcompiler/blob/c99b2e0
a395ad6be551704e6ff2f9c22c4a2278f/internal/compiler-interface/src/main/j
ava/xsbti/compile/IncrementalCompiler.java) used together with [`
Inputs<Analysis,ScalaCompiler>`](https://github.com/sbt/incrementalcompi
ler/blob/c99b2e0a395ad6be551704e6ff2f9c22c4a2278f/internal/compiler-inte
rface/src/main/java/xsbti/compile/Inputs.java), etc.
Over the course of time, the Java API became out of sync with the
incremental compiler implementation in `IC` object as features were
added such as:
- name hashing
- new Java compiler (sbt/sbt#1702)
- bytecode enhancement support (sbt/sbt#1714)

This change is intended to fold these changes back into natural places,
like making `previousResult` part of the `Inputs`.
sbt-datatype is used to generate Java datatypes, which should help us
maintain bincompat going forward.
The centerpiece of this API is:

``` scala
def compile(in: Inputs, log: Logger): CompileResult
```

The intent is to set up incremental compilation using only the publicly
exposed datatypes.
One of the glaring exception is the setup of `AnalyzingCompiler`, which
takes 50+ lines of acrobatics to create.
This is demonstrated in the newly added [incrementalcompiler/src/test/scala/sbt/inc/IncrementalCompilerSpec.scala](https://github.com/sbt/incrementalcompiler/blob/topic/api_point/incrementalcompiler/src/test/scala/sbt/inc/IncrementalCompilerSpec.scala).

/review @Duhemm, @dwijnand, @jsuereth


internal/compiler-interface/src/main/datatype/incremental.json
```diff
@@ -138,6 +138,266 @@
...
+      "namespace": "xsbti.compile",
+      "target": "Java",
+      "type": "record",
+      "doc": [
+        "Compilation options."
+      ],
+      "fields": [
+        {
+          "name": "classpath",
+          "type": "java.io.File []",
```
Duhemm: Arrays are represented using `java.io.File*`

internal/compiler-interface/src/main/java/xsbti/compile/IncrementalCompiler.java
```diff
@@ -67,5 +57,5 @@
 	* @param targetJar Where to create the output jar file containing the compiled classes.
 	* @param instance The ScalaInstance to compile the compiler interface for.
 	* @param log The logger to use during compilation. */
-	void compileInterfaceJar(String label, File sourceJar, File targetJar, File interfaceJar, ScalaInstance instance, Logger log);
+	// void compileInterfaceJar(String label, File sourceJar, File targetJar, File interfaceJar, ScalaInstance instance, Logger log);
```
Duhemm, line 60: Will `newScalaCompiler` and `compileInterfaceJar` be restored or do you intend to remove them?

eed3si9n, line 60: We need some mechanism to provide `AnalyzingCompiler`, otherwise this API is not useful.

dwijnand:
I defer to @Duhemm :)

PR #31, opened 2015-11-16
Fix comparison of `MiniSetup`.

Duhemm:
Previously `CompilationSetup`s were compared using an instance of
`Equiv[CompileSetup]` that was defined in `CompilationSetup`'s companion
object.

`CompilationSetup` has been renamed to `MiniSetup`, but there's no
longer a `MiniSetup` companion object because `MiniSetup` is created
using sbt-datatype. We now have to import the instance of
`Equiv[MiniSetup]` to use.


eed3si9n:
LGTM

PR #35, opened 2015-12-07
Uses URI to write out inc_compile text

eed3si9n:


dwijnand:
Out of interest, what's this for?

PR #37, opened 2015-12-20
Add extensible metadata into Setup

eed3si9n:
Setup now contains an array of `(String, String)`:

``` json
        {
          "name": "extra",
          "type": "xsbti.T2<String, String>*"
         }
```

This can be used to invalidate previous analysis for various purposes including sbt version changes.

/review @Duhemm, @dwijnand


internal/incrementalcompiler-core/src/main/scala/sbt/internal/inc/MiniSetupUtil.scala
```diff
@@ -19,13 +20,20 @@ object MiniSetupUtil {
...
-        sameNameHasher
+        sameNameHasher &&
+        sameExtra
     }
   }
+  implicit val equivPairs: Equiv[Array[T2[String, String]]] = new Equiv[Array[T2[String, String]]] {
+    def equiv(a: Array[T2[String, String]], b: Array[T2[String, String]]): Boolean =
+      ((a.toList filter { x => !(x.get1 startsWith "info.") }) sameElements
+        (b.toList filter { x => !(x.get1 startsWith "info.") }))
+  }
```
Duhemm, line 36: Two `extra`s are considered equivalent if they have the same mapping for all keys that do not start with `info.`?
So the two following `extra`s are equivalent?

``` scala
val e1 = Array(("info.foo", "foo"))
val e2 = Array(("info.foo", "bar"), ("info.baz", "hello"))
```

Why?

eed3si9n, line 36: When the metadata is not equivalent we will invalidate the previous compilation result, but there might be a situation where we would like to track some information without invalidating the previous.

Duhemm:
LGTM

PR #38, opened 2015-12-22
Port scripted tests from sbt/sbt

eed3si9n:
It doesn't cover all capability of `build.sbt` and flexibility, but it can do basic compile and file copying type of tests.

/review @Duhemm, @dwijnand


dwijnand:
Is this duplicating code with the sbt repo?

eed3si9n:
@dwijnand Yes. The core scripted engine is currently duplicated. I can split it out into util.

dwijnand:
If deduping into util is feasable, I think it would be preferable. Thanks.

Duhemm:
I suggest we merge this as is and open an issue with a few points:
- [ ] Have the core scripted engine in `sbt/util`
- [ ] Reuse `ScalaInstance`
- [ ] Implement `run`
- [ ] Don't show the whole stack trace when a pending test fails

I see there are tests that are marked pending because they depend on a structure with subprojects or because they use compiler plugins. I guess these tests will be ported to sbt/sbt 1.0's build?

LGTM

eed3si9n:
Yea. Let's do that so you can work on them.

PR #40, opened 2015-12-23
Use scripted core from sbt/util

Duhemm:


eed3si9n:
Closing in favor of #41

PR #41, opened 2015-12-24
Use scripted core from sbt/util

eed3si9n:
Continuation of #40


eed3si9n:
@Duhemm could you go over the pending scripted tests and find tests that can be added back because `run` exist?

Duhemm:
@eed3si9n Sure

Duhemm:
@eed3si9n I addressed your comments about the `run` task. Is it worth redirecting `System.out` and `System.err` to the logger?

eed3si9n:
>  Is it worth redirecting System.out and System.err to the logger?

As long as the test works, I don't care much about the side effects.

Duhemm:
@eed3si9n I surprisingly finally succeeded at reproducing the issue on my machine, and to "fix it" by sleeping after each file command.

eed3si9n:
@Duhemm o wow.

dwijnand:
I merged #57, which disabled `source-dependencies/nested-case-class`, which you said you'd then re-enable here, so here's me reminding you.

eed3si9n:
I'm going to merge this as is for now in the hopes to merge other pending pull requests.

PR #43, opened 2015-12-24
Use real local to retrieve compiler bridge during test. 

### LINKED ISSUE #36: include Ivy Local into resolving compiler interface and compiler bridge
Currently it requires `publishM2` instead of `publishLocal`.

A variation of this issue is that currently it requires `publishM2` of util because util-interface is needed for compile-interface.
## steps
1. `publishLocal` util with SHA1 version.
2. change incrementalcompiler's build to rely on the SHA1 version of util.
3. run test on incrementalcompiler after `publishM2`ing the incrementalcompiler JARs
## problem

One problem is that when the compiler fails to find the interface, it seems to keep going, and then it fails to compile the bridge:

``` scala
/var/folders/7g/l96y5q310h90xsz109qqjpmc0000gn/T/sbt_10b7c08a/xsbt/CompilerInterface.scala:6: error: object Logger is not a member of package xsbti
import xsbti.{ AnalysisCallback, Logger, Problem, Reporter, Severity }
       ^
```
## expectation

I can `publishLocal` both util and incrementalcompiler.

eed3si9n:
Fixes #36

/review @Duhemm

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -241,8 +241,8 @@ lazy val publishBridgesAndTest = Command.args("publishBridgesAndTest", "<version
   val version = args mkString ""
   val compilerInterfaceID = compilerInterface.id
   val compilerBridgeID = compilerBridge.id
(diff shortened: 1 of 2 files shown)


Duhemm:
LGTM. Thanks!

PR #48, opened 2016-01-17
FPORT: Assorted incremental compiler bug fixes

dwijnand:
This is the non-release notes, non-scripted tests part of forward porting https://github.com/sbt/sbt/pull/2343.

Cherry-picking this was harder than most of the others I'm doing in sbt/sbt#2352, so @adriaanm and @romanowski, as well as perhaps @gkossakowski and @Duhemm, a review here wouldn't hurt at all. Thanks, sorry for the inconvenience.

### DIFF
--- a/internal/compiler-bridge/src-2.10/main/scala/xsbt/Dependency.scala
+++ b/internal/compiler-bridge/src-2.10/main/scala/xsbt/Dependency.scala
@@ -96,12 +96,9 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile {
   }
 
   private abstract class ExtractDependenciesTraverser extends Traverser {
-    protected val depBuf = collection.mutable.ArrayBuffer.empty[Symbol]
-    protected def addDependency(dep: Symbol): Unit = { depBuf += dep; () }
-    def dependencies: collection.immutable.Set[Symbol] = {
-      // convert to immutable set and remove NoSymbol if we have one
-      depBuf.toSet - NoSymbol
-    }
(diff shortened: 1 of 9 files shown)


eed3si9n:
@dwijnand Thanks for doing this. You need https://github.com/sbt/sbt/pull/2358 to get the test to pass.

eed3si9n:
```
[error] /home/travis/build/sbt/incrementalcompiler/internal/incrementalcompiler-core/src/main/scala/sbt/internal/inc/Incremental.scala:24: not found: type Logger
[error]   class PrefixingLogger(val prefix: String)(orig: Logger) extends Logger {
[error]                                                                   ^
```

dwijnand:
Ok reproduced my test failure on Travis: https://travis-ci.org/sbt/incrementalcompiler/jobs/102993289#L1573

```
[info] - should extract source dependencies from public members *** FAILED ***
[info]   Set('E, 'A, 'G, 'B, 'C, 'D) did not equal Set('E, 'A, 'G, 'B, 'D) (DependencySpecification.scala:27)
```

This assertion was specifically changed in sbt/sbt#2358. Can someone, perhaps @gkossakowski, explain why this test is failing here now?

dwijnand:
Btw the above test failure is only under 2.11.7 which might be of interest.

Duhemm:
The issue is that the changes have not been applied to the 2.11 version of the compiler bridge. I'm taking care of this! Thanks for starting :wink:

dwijnand:
Thanks @Duhemm

Duhemm:
Superseded by #53

PR #50, opened 2016-01-17
Fix DependencySpecification test.

dwijnand:
Forward port of sbt/sbt#2358.

This is a fixup of 0f616294c4e713dc415f5dc3ae7aef257decb228.
That commit assumed that dealiasing is being done for types referred in
self type. It was changed to not do that but the test wasn't updated.
Unfortunately, that mistake slipped by during PR review because unit tests
of compileInterface were not ran (see #2358).


eed3si9n:
This commit should be added into #48

dwijnand:
Thanks, cherry-picked and pushed.

PR #52, opened 2016-01-17
Upgrade to latest sbt/librarymanagement

dwijnand:
Brings in among other things:
- sbt/librarymanagement#17 FPORT: Make JCenter opt in
- sbt/librarymanagement#18 FPORT: Make sbt aware of Dotty 
- sbt/librarymanagement#19 FPORT: Maven compatibility changes (intransitive warnings and "configuration not public") 
- sbt/librarymanagement#20 FPORT: Refix snapshots


eed3si9n:
I think we should use M6, or whatever the next number is for librarymanagement to avoid getting evicted.
I'm not too sure if Ivy can order SHA1 commit versions correctly when there's a conflict.
I use that for incrementalcompiler since it's used only by the root projects like sbt and zinc.

PR #51, opened 2016-01-17
FPORT: Consider signatures of methods before and after erasure in ExtractAPI

dwijnand:
Forward port of sbt/sbt#2261.

PR #54, opened 2016-01-19
FPORT: Consider signatures of methods before and after erasure in ExtractAPI

Duhemm:
Forward port of sbt/sbt#2261.

(Supersedes #51, thanks @dwijnand!)


Duhemm:
This makes no sense. Removing `hello` and hoping that the problem came from Travis.

dwijnand:
:+1: Travis passed.

Duhemm:
Note: We need to re-enable scripted test `source-dependencies/value-class` after this PR gets merged.

eed3si9n:
Merged

PR #57, opened 2016-01-25
FPORT: Fix sbt/sbt#2411

Duhemm:


Duhemm:
The tests will fail because support for `run` in scripted tests is implemented in #41, but this PR fixes the spurious failures that we've seeing with Travis in #41.

I'll disable the test `source-dependencies/nested-case-class` here and re-enable it in #41 after this PR gets merged.

dwijnand:
LGTM, merging as it's a forward port.

PR #67, opened 2016-02-23
FPORT: Do not rely on filename of package objects

Duhemm:
Forward port of sbt/sbt#2438 (Reopening because of "branch not found" on Travis)


eed3si9n:
Somehow Travis CI is confused about where to get the branch from.

```
The command "eval git clone --depth=50 --branch=fport/package-object-filename https://github.com/sbt/incrementalcompiler.git sbt/incrementalcompiler" failed. Retrying, 2 of 3.
```

PR #73, opened 2016-02-27
FPORT: Avoid CCE when scalac internally uses compileLate

dwijnand:
Scripted test half of forward porting sbt/sbt#2453.

Porting to here rather than in sbt/sbt (sbt/sbt#2487).

### DIFF
--- a/incrementalcompiler/src/sbt-test/apiinfo/source-path/build.sbt
+++ b/incrementalcompiler/src/sbt-test/apiinfo/source-path/build.sbt
@@ -1,0 +1,3 @@
+scalaVersion := "2.11.7"
+
+scalacOptions in Compile ++= "-sourcepath" :: (baseDirectory.value / "srcpath").toString :: Nil
--- a/incrementalcompiler/src/sbt-test/apiinfo/source-path/src/main/scala/test.scala
+++ b/incrementalcompiler/src/sbt-test/apiinfo/source-path/src/main/scala/test.scala
@@ -1,0 +1,17 @@
+object Test {
+  // When the refchecks compiler phase checks if Tuple2 has the
+  // @deprecated annotation, it forces the info for the
+  // `@deprecatedInheritance` annotation (the only annotation on
+  // class Tuple2.
+  //
+  // Because we are compiling with `-sourcpath` that contains a
+  // source file for `deprecatedInheritance`, this triggers a
+  // `compileLate` of that file (which basically runs all previous
+  // compiler phases on that file.)
+  //
+  // `compileLate` assumes that all of the phases are subclasses
+  // of `GlobalPhase`, rather than just `Phase`. This triggers a
+  // `ClassCastException` when it encounters SBT's custom
+  // API phase.
+  new Tuple2("", "")
+}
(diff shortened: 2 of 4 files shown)


dwijnand:
The failure is unrelated to the addition ("remove-test-b" failed). Restarting.

Duhemm:
I think scripted tests in `incrementalcompiler` don't support `build.sbt`...

We should have a way to support passing some configuration to the incremental compiler.

dwijnand:
Really? Because there's a bunch of them:

```
09:15:16 $ find incrementalcompiler -name '*.sbt'
incrementalcompiler/src/sbt-test/source-dependencies/abstract-type-override/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/backtick-quoted-names/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/canon/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/compactify/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/continuations/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/export-jars/changes/build2.sbt
incrementalcompiler/src/sbt-test/source-dependencies/ext/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/inherited-dependencies/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/inherited-deps-java/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/inherited-macros/macro-client/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/inherited_type_params/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/java-mixed/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/java-mixed/changes/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/less-inter-inv/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/less-inter-inv-java/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/relative-source-error/changes/absolute.sbt
incrementalcompiler/src/sbt-test/source-dependencies/relative-source-error/changes/relative.sbt
incrementalcompiler/src/sbt-test/source-dependencies/resident-java/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/restore-classes/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/same-file-used-names/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/transitive-memberRef/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/type-alias/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/typeref-only/build.sbt
incrementalcompiler/src/sbt-test/source-dependencies/value-class/build.sbt
```

Duhemm:
Yes, the tests are either pending, or the build.sbt define `check-iterations` or `check-same` (they have been re-implemented in [IncHandler.scala](https://github.com/sbt/incrementalcompiler/blob/1.0/internal/incrementalcompiler-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala#L58-L68))

Note that some tests set `recompileAllFraction`, which could be useful to avoid false positives.

eed3si9n:
We should add a custom command to replace the flag added in build.sbt.

dwijnand:
Why do scripted tests in incrementalcompiler not support .sbt files?

Duhemm:
Because these tests do not start a complete sbt instance like scripted in sbt/sbt does. After all, here were only interested in testing the incremental compiler. I've added support for new commands and `IncOptions` in #76.

dwijnand:
That makes sense. Thanks.

dwijnand:
I've created #77 to fix the general bug of scripted tests that use build.sbt.

Would it be possible to merge this forward port for now, deferring the remaining fixes to #77?

PR #74, opened 2016-02-27
Move scripted tests into position

dwijnand:


dwijnand:
Uhm.. Travis isn't running on this.. I'll try bouncing the PR.

dwijnand:
For reference the logs show:

> Running source-dependencies / fbounded-existentials

and

> Running source-dependencies / trait-private-object

So I think we're good. Merge? Bump?

eed3si9n:
LGTM

Issue #83, opened 2016-03-17
Scripted doesn't compile Java sources


gkossakowski:
The scripted framework doesn't support compiling Java sources. They are simply ignored (and not registered in Analysis object).

This blocks porting some of the Java-specific scripted tests from class-based-dependencies branch.

PR #86, opened 2016-03-19
Class dependencies

gkossakowski:
This PR implements class-based dependency tracking. The basic idea is simple: let's track dependencies at class level instead of at source level. Dependencies on inner classes are tracked separately. Also, APIs (to detect which dependencies has changed) are tracked for each class separately.

The detailed list of what has been done and what's left to be done was tracked in sbt/sbt#1104. I'm pasting it here for convenience.

----

### Tasks

Class-based dependency tracking can be implemented in two phases. The third phase is dedicated to testing on large projects.
#### Invalidate by inheritance based on class dependencies (phase 1)

_~~Tasks marked as completed are completed only in a prototype implementation. No changes has been merged to sbt yet.~~_
- [x] add tracking of declared classes in source files; we want to invalidate classes but at the end we need className -> srcFile relation because we can recompile source files only
- [x] add dependency tracking of top-level (not owned by any other class) imports; we'll assign dependencies introduced by those imports to the first top level class/object/trait declared in the source file (see notes below for details)
- [x] track dependencies at class level (and use `declaredClasses` to immediately map them back to files until the rest of the algorithm is implemented)
- [x] add tracking of children of sealed parents; we should track this in `API` object corresponding to sealed parent; this will enable us to perform proper invalidation when a new case (class) introduced to a sealed hierarchy
- [x] add per (top level) class tracking of api hashes; tracking of api hashes per each (even nested) class separately will be done in phase 2
- [x] handle dependencies coming from local and anonymous classes (see discussion below)
- [x] switch invalidation of dependencies by inheritance to invalidate at class level instead of at source level (would fix #2320)
- [x] distinguish between source and binary class names. Introduce a new relation `binaryClassName: Relation[String, String]` and use it whenever the conversion is needed.
#### Track name hashes per class (phase 2)
- [x] refactor tracking of APIs to be done at class level instead of at source level
- [x] extract APIs of all classes (including inner ones) separately
- [x] track name hashes at class level instead of being at source level (see also: #2319)
- [x] implement the rest of invalidation logic (member reference, name hashing) based on class names (get rid of mapping through `declaredClasses` in almost entire algorithm) (would fix #2320)

Only once the last bullet is merged we'll see improvement to incremental compilation when big number of nested classes is involved (e.g. like in Scala compiler's case).
#### Testing, bug fixing and benchmarking (phase 3)
- [x] test with the Scalatest repo
- [x] test with the scala repo using the sbt build
- [x] test with the specs2 repo
- [ ] fix handling of anonymous and local classes defined in Java (similar logic will have to be implemented as for handling local Scala classes)
- [ ] index classes only declared (not inherited) in source files
- [ ] benchmark `;clean;compile` performance
- [x] simplify classpath and Analysis instance lookup in incremental compiler (I think number of classpath lookups can be reduced now)
#### Merge changes upstrem, prepare for shipping (phase 4)
- [x] determine the location where this work will be merged into
### Benefits

The improved dependency tracking delivers up to 40x speedups of incremental compilation in scenarios tested. Check benchmarking results here: https://github.com/sbt/sbt/issues/1104#issuecomment-193002190

The speedups are caused by fixing two main issues:
1. The #2319 would be fixed once name hashes are tracked per class. This way introduction of a new class and members coming with it would not affect source files dependent (by member ref) on already existing classes.
2. Effects of adding members (like methods) to a class would affect only classes that inherit from that class. At the moment, adding a member to a class that nobody inherits from can trigger invalidation of all descendants of all classes defined in the same source file (see #2320 for a specific scenario).

The 1. is likely to be triggered by any code base that uses more than one class defined in a single source file. The 2. is affecting code bases with big number of nested classes that are inherited.
One example is Scala compiler itself. Even with name hashing, we invalidate too much and code edit cycle becomes long whenever a new member is introduced.

----

### Known issues
- Java-specific tests are ignored (see #83)
- Java Anonymous and local classes are not handled properly leading to either crashes or undercompilation (see the task above)
- asSeenFrom is not honored correctly for inherited classes (they are extracted only as seen from declaration site) which leads to under compilation and missing typechecking errors (https://gitter.im/scala/contributors?at=56ec053bbb4a1731739afb6f)
- handling of private members in traits has been reverted due to sbt/sbt#2490
#### Backwards compatibility

This PR affects backwards compatibility in several ways:
- support for Scala 2.7, 2.8 and 2.9 has been dropped
- support for `Analysis.groupBy` operation has been dropped (it's not clear who needs it)
- the API of `Analysis` object is very different now because most methods take classNames as arguments instead of file names


gkossakowski:
Review by @Duhemm and @adriaanm 

Realistically speaking, I don't expect you to review all commits. I'd recommend you going through the list of tasks that I implemented and just check if you understand why they were done. If something is not clear, ask me.

If the motivation coming from tasks is clear then commits will be easy to understand both now and in the future.

internal/incrementalcompiler-core/src/main/scala/sbt/internal/inc/MemberRefInvalidator.scala
```diff
@@ -53,65 +53,59 @@ import xsbt.api.APIUtil
  * of regular members then we'll invalidate sources that use those names.
  */
```
smarter, line 46: The whole comment for this class needs to be updated since it describes the strategy as being implemented as `T => Set[File]` whereas now it's `String => Set[String]`.

internal/incrementalcompiler-core/src/main/scala/sbt/internal/inc/Relations.scala
```diff
@@ -196,35 +265,38 @@ trait Relations {
...
-      ("used names", "File:String")
+      FFRelationDescriptor("products", _.srcProd),
+      FFRelationDescriptor("binary dependencies", _.binaryDep),
+      // NameHashing doesn't provide direct dependencies
+      FFRelationDescriptor("direct source dependencies", r => Relations.emptySource.internal),
+      // NameHashing doesn't provide direct dependencies
+      FSRelationDescriptor("direct external dependencies", r => Relations.emptySource.external),
+      // NameHashing doesn't provide public inherited dependencies
+      FFRelationDescriptor("public inherited source dependencies", r => Relations.emptySource.internal),
+      // NameHashing doesn't provide public inherited dependencies
```
Duhemm: If NameHashing is becoming the only option, we should be able to get rid of `direct` and `publicInherited` relations

internal/compiler-bridge/src-2.10/main/scala/xsbt/Dependency.scala
```diff
@@ -33,59 +33,150 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile {
   import global._
 
   def newPhase(prev: Phase): Phase = new DependencyPhase(prev)
-  private class DependencyPhase(prev: Phase) extends GlobalPhase(prev) {
+  private class DependencyPhase(prev: Phase) extends Phase(prev) {
```
Duhemm, line 36: Isn't this undoing the changes introduced in sbt/sbt#2435 / #73 ?

internal/incrementalcompiler-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -106,21 +137,22 @@ private[inc] abstract class IncrementalCommon(log: sbt.util.Logger, options: Inc
 
       new APIChanges(apiChanges)
     }
-  def sameSource[T](src: T, a: Source, b: Source): Option[APIChange[T]] = {
+  def sameClass(className: String, a: AnalyzedClass, b: AnalyzedClass): Option[APIChange] = {
     // Clients of a modified source file (ie, one that doesn't satisfy `shortcutSameSource`) containing macros must be recompiled.
```
Duhemm: The comment should read `shortcutSameClass` now that it has been renamed

Duhemm:
It looks like `ClassToAPISpecification.scala` has been deleted in b0a8a91aaa5b6f55714986f7bf389ec8aa091ec0

What this intended?

gkossakowski:
> It looks like ClassToAPISpecification.scala has been deleted in b0a8a91

It's a broken conflict resolution that I've missed. Great catch! I pushed a commit that moves ClassToAPI and Analyze specifications.

I also pushed commits that fix local inheritance dependencies.

internal/incrementalcompiler-core/src/main/scala/sbt/internal/inc/MemberRefInvalidator.scala
gkossakowski, line 46: I've pushed a commit that updates docs.

gkossakowski:
Thanks for spotting all the missing articles @smarter. I pushed updated version of that commit.

internal/incrementalcompiler-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
gkossakowski: Fixed.

internal/incrementalcompiler-core/src/main/scala/sbt/internal/inc/Relations.scala
gkossakowski: Good idea. I've removed them and while I was at it, I removed `declaredClasses` relations too (that wasn't used for anything).

smarter:
The comment on top of `ExtractAPI.scala` is no longer accurate, it says:

```
 * This class depends on instance of CallbackGlobal instead of regular Global because
 * it has a call to `addInheritedDependencies` method defined in CallbackGlobal.
```

But the call to `addInheritedDependencies` is gone.

gkossakowski:
I added a small refactoring of the code that performs classpath and Analysis lookups. My motivation was streamline bits of code that are passing `entry` and `forEntry` functions.

I updated `ExtractAPI` per @smarter's comment.

Is this good to go for you?

smarter:
LGTM in the limited testing I did, however I'm still a bit concerned by the performance regressions on scala/scala (https://github.com/sbt/sbt/issues/1104#issuecomment-199360458), would be good to know why this happens and if this happens in other big projects too.

smarter:
Also the tests seem to be failing, no idea why

Duhemm:
LGTM!

gkossakowski:
The test failure was a Travis glitch.

eed3si9n:
Merged!

gkossakowski:
Excellent news! I'll work on remaining issues. Just as a convenience, here they are:
- handling of anonymous and local classes defined in Java (similar logic will have to be implemented as for handling local Scala classes)
- Java-specific tests are ignored (see #83)
- asSeenFrom is not honored correctly for inherited classes (they are extracted only as seen from declaration site) which leads to under compilation and missing typechecking errors (https://gitter.im/scala/contributors?at=56ec053bbb4a1731739afb6f)

internal/compiler-bridge/src-2.10/main/scala/xsbt/ExtractUsedNames.scala
```diff
@@ -57,57 +84,81 @@ class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) ext
...
+          usedNameInImportSelector(selector.name)
+          usedNameInImportSelector(selector.rename)
+        }
+      // TODO: figure out whether we should process the original tree or walk the type
+      // the argument for processing the original tree: we process what user wrote
+      // the argument for processing the type: we catch all transformations that typer applies
+      // to types but that might be a bad thing because it might expand aliases eagerly which
+      // not what we need
+      case t: TypeTree if t.original != null =>
+        t.original.foreach(traverse)
```
retronym, line 137: I believe this line (and the one above for macro originals) is a performance regression. Both `foreach` and `traverse` will recurse into the tree. See https://github.com/sbt/zinc/pull/187

Either `t.original.foreach(handleTreeNode)` or simply `traverse(t.original)` seems to be the right approach.

jvican: FTR, fixed in https://github.com/sbt/zinc/pull/193.

PR #89, opened 2016-03-27
Add minimal .gitignore

smarter:


eed3si9n:
Where does this end? Let's all use global gitignore file and be done with it - https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

smarter:
I'm fine with that too, just noticed that this repo didn't have one like sbt/sbt does.

SethTisue:
my two cents: for stuff like `.DS_Store` files, Emacs backup files, etc, a global .gitignore is a good solution. I don't think it's a good way to handle directories containing build products; in a Scala context a directory named `target` is _probably_ ignorable, but not certainly, and in other contexts the probability drops.

gkossakowski:
I'm with @SethTisue on this one. I don't think requiring a global gitignore to include `target` is a good idea. Unfortunately, the name is too generic.

dwijnand:
For myself, as a Scala engineer, I've opted for globally ignoring `target/` (cos life's too short to micro this all the time).

But given that this directory is a direct consequence of there being an sbt build committed to the repo, I agree to this change. That's the test: if it's a derivative of committed files then ignore, if not no (therefore no .DS_Store, *.swp, *.iml, etc).

eed3si9n:
ok. Thanks all for the input.

PR #94, opened 2016-04-07
Overhaul ClassfileManager to pessimistically clone

stuhood:
ClassfileManager is supposed to protect against errors occurring during compilation which would prevent analysis from being written.

But with its current API, it is only able to do that job with the explicit cooperation of all tools involved. For example, if javac or scalac decided to overwrite a file before raising an exception, or analysis of classfiles fails in `sbt.internal.inc.classfile.Analyze`, then the post-facto call to `generated` is too late to back things up. This is not just a theoretical concern: the Analyze object can _currently_ fail when partial classpaths are in play. Will post a separate patch to improve that, but it's certainly not going to be the last time we see an exception.

Instead, this review changes the API of ClassfileManager to allow for pessimistically cloning the class directory into the configured backup directory, and restoring it in case of failure.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/ClassfileManager.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/ClassfileManager.scala
@@ -3,89 +3,118 @@ package internal
 package inc
 
 import sbt.io.IO
+import sbt.io.Path._
+
 import java.io.File
 import collection.mutable
-import xsbti.compile.{ IncOptions, DeleteImmediatelyManagerType, TransactionalManagerType }
+import xsbti.compile.{ IncOptions, DeleteImmediatelyManagerType, Output, SingleOutput, MultipleOutput, TransactionalManagerType }
 
 /**
- * During an incremental compilation run, a ClassfileManager deletes class files and is notified of generated class files.
+ * During an incremental compilation run, a ClassfileManager backs up classfiles, and
+ * can restore them in case of failure.
+ *
  * A ClassfileManager can be used only once.
  */
 trait ClassfileManager {
   /**
-   * Called once per compilation step with the class files to delete prior to that step's compilation.
-   * The files in `classes` must not exist if this method returns normally.
-   * Any empty ancestor directories of deleted files must not exist either.
+   * Called before any changes to classfiles have been made.
    */
-  def delete(classes: Iterable[File]): Unit
+  def begin(): Unit
 
-  /** Called once per compilation step with the class files generated during that step.*/
-  def generated(classes: Iterable[File]): Unit
+  /**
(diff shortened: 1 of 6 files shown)


eed3si9n:
> Instead, this review changes the API of ClassfileManager to allow for pessimistically cloning the class directory into the configured backup directory, and restoring it in case of failure.

Would there be performance implications for projects that has hundreds of files, especially on Windows?

stuhood:
> Would there be performance implications for projects that has hundreds of files, especially on Windows?

The performance implications should be proportional to the total size of the compile. On OSX, for one of our worst-case targets containing 3849 classfiles:
- `cp -R` takes ~3 seconds with a cold filesystem cache, and then steadies out at 1.7 seconds
- `jar c0f` takes about 0.5 second
- `tar cf` takes about 0.1 second (with COPYFILE_DISABLE)

...I can have this jar up these files instead? 

But additionally, this is an (optional) correctness guard which currently doesn't work. It seems fairly important that it actually works if people choose to enable it.

gkossakowski:
Have you considered pushing calls to `generated` before `Analyze` has run? I think ClassfileManager could be notified as soon as class files are written by hooking into Scala compiler for Scala class files. For Java class files you can implement a custom version of [JavaFileManager](https://docs.oracle.com/javase/7/docs/api/javax/tools/JavaFileManager.html) that records what class file are being written.

stuhood:
@gkossakowski : The JavaFileManager approach would only work for un-forked compilers, so I think that would require 3 different strategies. I don't really have the resources for something that ambitious unless the pessimistic approach looks like an actual performance bottleneck.

gkossakowski:
The pessimistic approach has time complexity proportional with the size of
the project. In the incremental compiler we would like to have most of the
cost dependent on the size of the incremental change.

Meanwhile, I'd make the pessimistic class file manager an opt-in option.

W dniu środa, 13 kwietnia 2016 Stu Hood notifications@github.com
napisał(a):

> @gkossakowski https://github.com/gkossakowski : The JavaFileManager
> approach would only work for un-forked compilers, so I think that would
> require 3 different custom classes. I don't really have the resources for
> something that ambitious unless the pessimistic approach looks like an
> actual performance bottleneck.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> https://github.com/sbt/zinc/pull/94#issuecomment-209570544

## 

Grzegorz Kossakowski
twitter: @gkossakowski http://twitter.com/gkossakowski
github: @gkossakowski http://github.com/gkossakowski

stuhood:
I think I agree with the objections. Moving the classfile recording before javac analysis makes sense, so may open a PR to do that instead.

PR #107, opened 2016-05-01
SafeLazy: explicitly turn by-name parameter into Function0

smarter:
Dotty does not support eta-expanding lazy vals, it's clearer to do it by hand.


eed3si9n:
LGTM

PR #115, opened 2016-05-02
Bring back `DefaultShowAPI#apply(SourceAPI)`

smarter:
Looks like it got lost in the class-based-dependencies merge in b0a8a91aaa5b6f55714986f7bf389ec8aa091ec0

Also remove `DefaultShowAPI#apply(ClassLike)`, it is not needed since `ClassLike` is a subclass of `Definition` and there is an `apply` overload for `Definition`.


smarter:
Ah nevermind, `SourceAPI` has been replaced by a sequence of `ClassLike`

PR #116, opened 2016-05-03
Fix API extraction of inherited classes

gkossakowski:
### Refactor ClassLike and add ClassLikeDef

This change refactors definition.json to make `ClassLike` a non-recursive type. Before the introduction of class-based dependency tracking, an api of a nested class has been stored as a member of the enclosing class so a class could have other class as a member. This was expressed by the fact that `ClassLike` was recursive: its Structure had a collection of `Definition` to represents the members and `ClassLike` was a subtype of `Definition`.
With introduction of class-based dependency tracking, each class has been extracted and stored separately. The inner class would still be stored as a member of outer class but members of inner classes were skipped.
An empty inner class was stored just to mark the fact that there's a member with a given name which is important for name hashing correctness when there's no dependency on a class directly but a rename could introduce one.

Storing an empty class was a hack and this commit fixes the type hierarchy by introducing the following changes:
- introduce ClassDefinition that is a subtype of Definition; all members
  of a class are subtypes of ClassDefinition (ClassLike is not a subtype
  of ClassDefinition)
- change Structure to refer to ClassDefinition instead of Definition for
  members
- move ClassLike higher up in type hierarchy so its a direct subtype of
  Definition
- introduce ClassLikeDef which represents an inner class as a member of
  the outer class; ClassLikeDef carries only information about class
  declaration itself but not about its members and that is enforced
  statically

NameHashing has been simplified because it doesn't have to keep track of the entire path for definitions it hashes. Hashes of names are tracked individually per class so location is simply name of the class and it's type (we want to distinguish between objects and classes).

NameHashingSpecification has been refactored to not rely on nested classes for testing the desired scenarios. The semantics of tests has been preserved even if a different API structure is used in tests.

---
### Fix naming of inherited classes in ExtractAPI

For an inherited class, ExtractAPI would form a (source) class name by calling `Symbol.className` on the inherited class. However, that created a name of a class as seen at declaration site and not at inheritance site. Let's consider an example:

``` scala
class A { class AA }
class B extends A
```

Before this change, ExtractAPI would create an API representation of `AA` twice: once seen from A, and then the second time seen from B as an inherited member. However, in both cases it would use `A.AA` as a name. This commit fixes naming so an inherited representation of `AA` has a name `B.AA`.

This commit also clarifies how classes declared in package objects are named. If you have:

``` scala
package pkg1.pkg2
package object pkg3 {
  class Foo
}
```

then the fully qualified name of the class corresponding to `pkg3` package object is `pkg1.pkg2.pkg3.package`. The full name of the `Foo` class is `pkg2.pkg2.pkg3.Foo`.

### DIFF
--- a/internal/compiler-bridge/src-2.10/main/scala/xsbt/ClassName.scala
+++ b/internal/compiler-bridge/src-2.10/main/scala/xsbt/ClassName.scala
@@ -20,6 +20,21 @@ trait ClassName {
    */
   protected def className(s: Symbol): String = pickledName(s)
 
+  /**
+    * Create a (source) name for the class symbol `s` with a prefix determined by the class symbol `in`.
+    *
+    * If `s` represents a package object `pkg3`, then the returned name will be `pkg1.pkg2.pkg3.package`.
+    * If `s` represents a class `Foo` nested in package object `pkg3` then the returned name is `pkg1.pkg2.pk3.Foo`.
+    */
+  protected def classNameAsSeenIn(in: Symbol, s: Symbol): String = atPhase(currentRun.picklerPhase.next) {
+    if (in.isRoot || in.isRootPackage || in == NoSymbol || in.isEffectiveRoot)
+      s.simpleName.toString
+    else if (in.isPackageObjectOrClass)
+      in.owner.fullName + "." + s.name
+    else
+      in.fullName + "." + s.name
+  }
+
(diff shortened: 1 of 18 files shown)


gkossakowski:
Review by @smarter or @eed3si9n

eed3si9n:
This does affect source-dependencies/as-seen-from-b?

PR #117, opened 2016-05-03
Classpath lookup performance fix

### LINKED ISSUE #100: Performance regression in the classpath lookup
The performance of classpath lookup has regressed in both sbt and zinc. The performance is lost due to a repeated calls to `FileValueCache.get` that performs a filesystem call to see if a corresponding file has changed and the cache entry needs to be invalidated.

The cached values are partially applied function `DefinesClass` that is an alias for `File => String => Boolean`. The `FileValueCache` caches `String => Boolean` functions based on a given file. The file here corresponds to a classpath element (e.g. directory or a jar). In sbt 0.13.11, the values from the cache are fetched in this code:

``` scala
/**
   * Returns a function that searches the provided class path for
   * a class name and returns the entry that defines that class.
   */
  def entry(classpath: Seq[File], f: DefinesClass): String => Option[File] =
    {
      val entries = classpath.toStream.map { entry => (entry, f(entry)) }
      className => entries.collect { case (entry, defines) if defines(className) => entry }.headOption
    }
```

Here, the argument `f` is a function roughly of the shape `(file: File) => FileValueCache.get(file)`. The first line looks up (lazily because it's a Stream) cached functions from the `FileValueCache`. The returned function in the second line closes over `entries`. The end result is that `entry` returns a function that closes over a series of functions that were looked up from `FileValueCache`.

Now, in zinc and sbt 1.0 the function `f` passed to the `entry` method is roughly of the shape `(file: File) => (className: String) => FileValueCache.get(file)(className)`. When you apply this function with a file in the `entries` definition, you'll just get another function `g` that _closes_ over `file` value but doesn't call the cache yet. The cache is called only when `className` is passed to `g` - that's is the cache is hit on every classpath lookup. The delayed evaluation of the `File` argument was introduced twice: once on the zinc side and second time on the sbt side. In both cases it's due to wrapping to accommodate to Java API.

The regression in zinc occurred at https://github.com/sbt/zinc/commit/9c75003b3129400b1ee3b8a16d1792132f776594#diff-f07388b6f3520add59fb517e7d3f530aR38.
The sbt regression was introduced at https://github.com/sbt/sbt/commit/20d0c41473f515cf94ebbc5ac9d59e2602284b76#diff-6373e7f7122325e753b75fe1cc76ff5fR878

gkossakowski:
Check the individual commits for details.

This fixes #100 but doesn't bring the much needed refactoring that I mentioned in #100.

### DIFF
--- a/internal/compiler-bridge/src-2.10/main/scala/xsbt/API.scala
+++ b/internal/compiler-bridge/src-2.10/main/scala/xsbt/API.scala
@@ -14,8 +14,6 @@ object API {
 final class API(val global: CallbackGlobal) extends Compat {
   import global._
 
-  @inline def debug(msg: => String) = if (settings.verbose.value) inform(msg)
-
   def newPhase(prev: Phase) = new ApiPhase(prev)
   class ApiPhase(prev: Phase) extends GlobalPhase(prev) {
     override def description = "Extracts the public API from source files."
@@ -25,7 +23,7 @@ final class API(val global: CallbackGlobal) extends Compat {
         val start = System.currentTimeMillis
         super.run
         val stop = System.currentTimeMillis
-        debug("API phase took : " + ((stop - start) / 1000.0) + " s")
+        debuglog("API phase took : " + ((stop - start) / 1000.0) + " s")
       }
 
     def apply(unit: global.CompilationUnit): Unit = processUnit(unit)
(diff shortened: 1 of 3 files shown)


eed3si9n:
LGTM pending Travis

PR #120, opened 2016-05-11
Rework DefinesClass lookup.

gkossakowski:
Before this commit, DefinesClass meant two things:
1. Function String => Boolean that determined whether a class with
    a given binary name is defined by a classpath entry the function
    instance is associated with
2. A function File => (String => Boolean) that created function from 1.
    for a given classpath entry

After this commit, `DefinesClass` means only the function from 1.

The operation from 2. is moved to newly introduced PerClasspathEntryLookup
interface. That interface also includes lookup of Analysis that corresponds
to a given classpath entry.

Thanks to this change, conversions between Java-defined function-like
objects and Scala functions is largely removed. These conversions were
error prone as shown by #100. As a bonus, the refactoring from this commit
makes it a bit easier to implement sbt/sbt#2525.

### DIFF
--- a/internal/compiler-interface/src/main/java/xsbti/compile/DefinesClass.java
+++ b/internal/compiler-interface/src/main/java/xsbti/compile/DefinesClass.java
@@ -1,12 +1,14 @@
 package xsbti.compile;
 
 /**
-* Determines if an entry on a classpath contains a class.
-*/
-public interface DefinesClass
-{
- 	/**
-	* Returns true if the classpath entry contains the requested class.
-	*/
-	boolean apply(String className);
+ * Determines if a classpath entry contains a class.
+ *
+ * The corresponding classpath entry is not exposed by this interface. It's tied
+ * to it by a specific class that implements this interface.
+ */
+public interface DefinesClass {
+    /**
+     * Returns true if the classpath entry contains the requested class.
+     */
+    boolean apply(String className);
 }
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Locate.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Locate.scala
@@ -7,11 +7,12 @@ package inc
 
 import java.io.File
 import java.util.zip.{ ZipException, ZipFile }
+
+import xsbti.compile.{ DefinesClass, PerClasspathEntryLookup }
+
 import Function.const
 
 object Locate {
-  type DefinesClass = File => String => Boolean
-
   /**
    * Right(src) provides the value for the found class
    * Left(true) means that the class was found, but it had no associated value
@@ -36,9 +37,9 @@ object Locate {
    * Returns a function that searches the provided class path for
    * a class name and returns the entry that defines that class.
    */
-  def entry(classpath: Seq[File], f: DefinesClass): String => Option[File] =
+  def entry(classpath: Seq[File], lookup: PerClasspathEntryLookup): String => Option[File] =
     {
-      val entries = classpath.toStream.map { entry => (entry, f(entry)) }
+      val entries = classpath.toStream.map { entry => (entry, lookup.definesClass(entry)) }
       className => entries.collect { case (entry, defines) if defines(className) => entry }.headOption
     }
   def resolve(f: File, className: String): File = if (f.isDirectory) classFile(f, className) else f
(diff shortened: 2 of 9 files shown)


internal/compiler-interface/src/main/java/xsbti/compile/DefinesClass.java
```diff
@@ -1,12 +1,14 @@
...
+ * Determines if a classpath entry contains a class.
+ *
+ * The corresponding classpath entry is not exposed by this interface. It's tied
+ * to it by a specific class that implements this interface.
+ */
+public interface DefinesClass {
+    /**
+     * Returns true if the classpath entry contains the requested class.
+     */
+    boolean apply(String className);
```
stuhood, line 13: Boolean isn't the best API for this, as confirming that something is defined will usually take a similar amount of time to just fetching it (similar to `Map.get` vs `Map.contains+Map.apply`).

Would prefer to see `Maybe`/`Option` here, or removal of this API.

internal/zinc-core/src/main/scala/sbt/internal/inc/Locate.scala
```diff
@@ -50,32 +51,38 @@ object Locate {
...
-      }
-      val entries = try { jar.entries.map(e => toClassName(e.getName)).toSet } finally { jar.close() }
-      entries.contains _
+  private object FalseDefinesClass extends DefinesClass {
+    override def apply(binaryClassName: String): Boolean = false
+  }
+
+  private class JarDefinesClass(entry: File) extends DefinesClass {
+    import collection.JavaConversions._
+    private val jar = try { new ZipFile(entry, ZipFile.OPEN_READ) } catch {
```
stuhood, line 68: AFAICT, this will unnecessarily hold a reference to the ZipFile in memory.

Maybe:

```
private val entries = {
  val jar = ???
  try { jar.entries... }
}
```

internal/zinc-core/src/main/scala/sbt/internal/inc/Locate.scala
```diff
@@ -50,32 +51,38 @@ object Locate {
...
 
   def toClassName(entry: String): String =
     entry.stripSuffix(ClassExt).replace('/', '.')
 
   val ClassExt = ".class"
 
-  def directoryDefinesClass(entry: File): String => Boolean =
-    className => classFile(entry, className).isFile
+  private class DirectoryDefinesClass(entry: File) extends DefinesClass {
+    override def apply(binaryClassName: String): Boolean = classFile(entry, binaryClassName).isFile
```
stuhood, line 84: While you're in here...

the recursive creation of files in `def subDirectory` is likely causing filesystem operations, as each `new File` call executes normalization steps.

internal/compiler-interface/src/main/java/xsbti/compile/DefinesClass.java
eed3si9n, line 13: Since the calling side looks like this,

``` scala
val entries = classpath.toStream.map { entry => (entry, lookup.definesClass(entry)) }
```

and if there's frequent enough call, it might make sense to avoid instantiation of `Maybe`s.
If we're going to change it, I would either rename `apply` to some verb phrase or use [`xsbti.F1`](https://github.com/sbt/util/blob/1.0/internal/util-interface/src/main/java/xsbti/F1.java).

internal/zinc-core/src/main/scala/sbt/internal/inc/Locate.scala
eed3si9n, line 68: I like the suggestion. We shouldn't need to hold on to zip file since the entries are converted to `Set[String]` anyway right?

internal/zinc-core/src/main/scala/sbt/internal/inc/Locate.scala
eed3si9n, line 84: @stuhood Do you have a suggestion for improving `subDirectory`? If so, could you send a separate PR or post to sbt-dev to discuss it?

eed3si9n:
This PR has been open for a while, so before we lose it I rebased against 1.0 and opened a new PR - https://github.com/sbt/zinc/pull/130, which I intend to merge when Travis passes.

PR #128, opened 2016-06-01
Minimal API for `ScalaCompiler`, expose in `Compilers`

Duhemm:
We only had an API for a `JavaCompiler`, and an empty interface for
`Compilers`. This commit adds an API for `ScalaCompiler` and exposes
both `JavaCompiler` and `ScalaCompiler` in `Compilers`.


eed3si9n:
LGTM

PR #130, opened 2016-06-08
Rework DefinesClass lookup.

eed3si9n:
This is a continuation of #120. It's rebased against 1.0, and added a minor refactoring.

Before this commit, DefinesClass meant two things:
1. Function String => Boolean that determined whether a class with
    a given binary name is defined by a classpath entry the function
    instance is associated with
2. A function File => (String => Boolean) that created function from 1.
    for a given classpath entry

After this commit, `DefinesClass` means only the function from 1.

The operation from 2. is moved to newly introduced PerClasspathEntryLookup
interface. That interface also includes lookup of Analysis that corresponds
to a given classpath entry.

Thanks to this change, conversions between Java-defined function-like
objects and Scala functions is largely removed. These conversions were
error prone as shown by #100. As a bonus, the refactoring from this commit
makes it a bit easier to implement sbt/sbt#2525.

/cc @gkossakowski

### DIFF
--- a/internal/compiler-interface/src/main/datatype/incremental.json
+++ b/internal/compiler-interface/src/main/datatype/incremental.json
@@ -320,18 +320,10 @@
       ],
       "fields": [
         {
-          "name": "analysisMap",
-          "type": "xsbti.F1<java.io.File,xsbti.Maybe<xsbti.compile.CompileAnalysis>>",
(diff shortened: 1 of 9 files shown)


eed3si9n:
LGTM pending Travis

gkossakowski:
Thanks for picking it up for me!

PR #131, opened 2016-06-09
(WIP) Adds FileWatch

eed3si9n:
Source: https://github.com/eed3si9n/zinc/tree/wip/watcher

This adds new Java interfaces `xsbti.compile.FileWatch` and
`xsbt.compile.FileChanges`, which are intended for proactive file
change detection using native OS hooks or other means such as IDEs.
### motivation

On Windows, especially it seems, there is a large overhead for running
compile task without changing any sources. From profiling we discovered
that 60% of the time is taken up by
`sbt.inc.IncrementalCommon.changedInitial`. This further breaks down to
[line 134 -
137](https://github.com/sbt/sbt/blob/v0.13.11/compile/inc/src/main/scala
/sbt/inc/IncrementalCommon.scala#L134-L137) spending 7%, 38%, 20%, and
34% respectively.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: scala
 jdk:
-  - openjdk7
+   - openjdk7
 scala:
    - 2.11.7
 
--- a/build.sbt
+++ b/build.sbt
@@ -160,7 +160,7 @@ lazy val zincCore = (project in internalPath / "zinc-core").
   dependsOn(zincApiInfo, zincClasspath, compilerBridge % Test).
   settings(
     testedBaseSettings,
-    libraryDependencies ++= Seq(sbtIO, utilLogging, utilRelation),
+    libraryDependencies ++= Seq(sbtIO, utilLogging, utilRelation, barbaryWatchService),
     // we need to fork because in unit tests we set usejavacp = true which means
     // we are expecting all of our dependencies to be on classpath so Scala compiler
     // can use them while constructing its own classpath for compilation
--- a/internal/compiler-interface/src/main/datatype/incremental.json
+++ b/internal/compiler-interface/src/main/datatype/incremental.json
@@ -327,6 +327,13 @@
           ]
         },
         {
+          "name": "fileWatch",
+          "type": "xsbti.compile.FileWatch",
+          "doc": [
+            "Detects change in file system."
+          ]
+        },
+        {
           "name": "skip",
           "type": "boolean",
           "doc": [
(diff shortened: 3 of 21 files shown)


jvican:
Can we close this @eed3si9n ? It's better to have the PR list empty before we start getting contributions, it sends another message to the people that will open new PRs. We can reopen when you want to follow up on this.

eed3si9n:
> Can we close this @eed3si9n ? It's better to have the PR list empty before we start getting contributions, it sends another message to the people that will open new PRs.

That makes no sense. I don't think I've ever cared if there's an open PR that says "(WIP)" or not when I thought about sending someone a PR.

dwijnand:
Consider switching to https://github.com/takari/directory-watcher.

romanowski:
IIRC Intellij has pretty effective watch service that we may be able to use (I think it comes from open sourced part of Intellij platform)

CC @niktrop

niktrop:
@romanowski 
Yes, we have one. But it uses many of platform specific classes, so it will be a nontrivial task to extract and use it.

jvican:
After the improvements to file watch in sbt/sbt, I see no reason why we should keep this open. Should we close?

eed3si9n:
This PR was about incorporating FileWatch in to the change detection of IncrementalCommon. I think it still stands.

jvican:
Okay, I struggled to see the benefit of this since I'm used to `~`. This option is clearly for those that don't use the file watching in their build tools. I think it's a useful improvement.
  
Do you want to follow up on this or do you want me to create a ticket and leave it open so that others can get this to the finish line?

jvican:
@eed3si9n ping.

eed3si9n:
> Okay, I struggled to see the benefit of this since I'm used to ~. This option is clearly for those that don't use the file watching in their build tools. I think it's a useful improvement.

Yes. Other build tools like IntelliJ calling Zinc, would still need the list of changed files.

> Do you want to follow up on this or do you want me to create a ticket and leave it open so that others can get this to the finish line?

Yea. It would be great if someone can pick this up.

dwijnand:
@eed3si9n This is an ancient PR at this point. It conflicts with lots of changes in the target branches. Could you create a new issue (or append to an existing one) with the changes intended by this pull request and cross-link?

That way we can get this off the PR backlog.

jvican:
Let's close this one until it's rebased and submitted again for review :smile:

eatkins:
I am actually working on a PR for sbt that I believe will resolve this issue by using external hooks to minimize the file system access within zinc.

PR #132, opened 2016-06-26
Java compilation refactoring and fixes

eed3si9n:
This is sort of a continuation of #128

There has been weird indirection around how Java compilation is wired up probably to keep bincompat during 0.13. @Duhemm's #128 started to clean that up, but I ran into an issue when I tried to adjust sbt to actually use it because `JavaTools` is what's needed not `JavaCompiler`.
Making `IncrementalCompilerJavaTools` [extend `JavaCompiler`](https://github.com/sbt/zinc/pull/128/commits/76a1a172641f443b8a4d6b9a70bb624c55955f62#diff-e8d61b91c2bfc9b46d463ed2e85d8aebR50) I don't think was the right choice. Instead now `JavaTool`, `JavaTools`, `JavaCompiler`, and `Javadoc` are all under `xsbti` as Java interfaces.

During testing I've also caught a Java error parsing bug, which was committed separately - 09933e8

/review @dwijnand @Duhemm


internal/compiler-interface/src/main/java/xsbti/compile/JavaTool.java
```diff
@@ -0,0 +1,26 @@
+package xsbti.compile;
+
+import java.io.File;
+import xsbti.Logger;
+import xsbti.Reporter;
+
+/**
+ * An interface for on of the tools in the java tool chain.
```
dwijnand: s/on/one/

internal/compiler-interface/src/main/java/xsbti/compile/JavaTool.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * We assume the following is true of tools:
+ * - The all take sources and options and log error messages
```
dwijnand: s/The/They/

internal/compiler-interface/src/main/java/xsbti/compile/JavaTool.java
```diff
@@ -0,0 +1,26 @@
...
+ * - They return success or failure.
+ */
+public interface JavaTool {
+  /**
+   * This will run a java compiler / or other like tool (e.g. javadoc).
```
dwijnand: I think we should drop the slash

eed3si9n: Sure. I just copy pasted from the existing comment including the above typos.

internal/compiler-interface/src/main/java/xsbti/compile/JavaTool.java
```diff
@@ -0,0 +1,26 @@
...
+import xsbti.Logger;
+import xsbti.Reporter;
+
+/**
+ * An interface for on of the tools in the java tool chain.
+ *
+ * We assume the following is true of tools:
+ * - The all take sources and options and log error messages
+ * - They return success or failure.
+ */
```
dwijnand, line 17: It looks like they also take a reporter. But also, why duplicate this info here?

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavaCompiler.scala
```diff
@@ -134,10 +57,21 @@ object JavaCompiler {
...
+      }
+      val augmentedClasspath = if (cpOptions.autoBoot) classpath ++ Seq(scalaInstance.libraryJar) else classpath
+      val javaCp = ClasspathOptionsUtil.javac(cpOptions.compiler)
+      (new CompilerArguments(scalaInstance, javaCp))(Array[File](), augmentedClasspath, Some(target), options)
+    }
 }
 
 /** Interface we use to document java code. This is a tag over the raw JavaTool interface. */
-trait Javadoc extends JavaTool {}
+// trait Javadoc extends JavaTool {}
```
dwijnand: I think we should remove this now

eed3si9n: Yea. I think I forgot to delete that line.

dwijnand:
Minor details. LGTM, nice work.

internal/compiler-interface/src/main/java/xsbti/compile/JavaTool.java
eed3si9n, line 17: I can wordsmith around what's intended to communicate here.

Issue #152, opened 2016-07-14
sbt.inc.IncrementalTest disabled


stuhood:
`sbt.inc.IncrementalTest`, which contains a large collection of tests (in an easy to debug harness/format) is currently mostly disabled.

The primary issue is that `sbt.inc.TestAnalysisCallback` (from the same file) has gone out of sync with the private AnalysisCallback in `sbt.inc.internal`. Since the AnalysisCallback impl is completely private, adjusting its implementation to be _directly_ usable for tests would avoid duplication and prevent losing sync again.

PR #153, opened 2016-07-14
Re-enable IncrementalTest as ignored.

stuhood:
See https://github.com/sbt/zinc/issues/152: this re-enables IncrementalTest as ignored for now. More work will be necessary in order to sync up the AnalysisCallback implementations, but there are some passing tests, and this will help to minimize compiler-checkable bitrot.

### DIFF
--- a/internal/zinc-core/src/test/scala/sbt/inc/IncrementalCompilerTest.scala
+++ b/internal/zinc-core/src/test/scala/sbt/inc/IncrementalCompilerTest.scala
@@ -27,7 +27,9 @@ object IncrementalCompilerTest {
                                                              |Mark this scenario as passing to clear this failure.""".stripMargin)
 
   /** Exception thrown when a step has failed. */
-  case class FailedStepException(state: ScenarioState, step: Step, cause: String) extends Exception {
+  case class FailedStepException(override val getMessage: String) extends Exception
+
+  object FailedStepException {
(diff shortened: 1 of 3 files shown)


eed3si9n:
These are commented out because we opted for using scripted tests - https://github.com/sbt/zinc/tree/1.0/zinc/src/sbt-test

jvican:
What's the status of this @eed3si9n ? Should this PR be closed since the decision to favour scripted tests was already taken?

eed3si9n:
Yea. Let's close it for now.

romanowski:
@eed3si9n if we don't want this tests anymore can we remove it from codebase with all its utils? I don't think that having whole file commented out and some dead code will help us anywhere. We can keep link to this (or any other) revision in e.g. comment somewhere if for some reason this code would be needed (but IMHO we can just remove all dead code).

@jvican  @stuhood WDYT?

eed3si9n:
Yea. I'm fine with that. We can always dig it out of the git history.

stuhood:
+1

On Feb 9, 2017 12:58 PM, "eugene yokota" <notifications@github.com> wrote:

Yea. I'm fine with that. We can always dig it out of the git history.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/sbt/zinc/pull/153#issuecomment-278771669>, or mute the
thread

PR #155, opened 2016-07-16
Bump to sbt-datatype 0.2.2

eed3si9n:
`*.scala` changes are due to parameter order change, which now lists from `super`'s field first, like `name`.

/review @dwijnand, @Duhemm


dwijnand:
LGTM

Issue #88, opened 2016-03-24
Spurious change to abstract modifier


gkossakowski:
``` scala
// Global.scala
class Global

// Analyzer.scala
trait Analyzer extends AnyRef with StdAttachments {
  val global : Global
}

// StdAttachments.scala
trait StdAttachments { self: Analyzer =>
  type MacroContext = Context { val universe: self.global.type }
}

// Context.scala
abstract class Context {
  val universe: Global
}

// Macros.scala
trait Macros { self: Analyzer =>
  val _openMacros = foo[MacroContext]()
  def foo[A](xs: A*): Foo[A] = null
}
class Foo[A]
```
1. `compile` (compile everything)
2. add whitespace to `Macros.scala` and run `compile`; observe:

```
[diff] -val _openMacros: this#Foo[this#Context {abstract val universe: Macros.this#global}]
[diff] +val _openMacros: this#Foo[this#Context {override abstract val universe: Macros.this#global}]
```

gkossakowski:
I have not succeeded at minimizing this further yet. For example, annotating `_openMacros`'s type explicitly makes the problem go away. The same goes if I change RHS to `(List[MacroContext](): List[MacroContext]))` and leave val without type annotation (!!!)

gkossakowski:
Possibly related to sbt/sbt#726 that was also about spurious changes to the override flag.

gkossakowski:
I managed to minimize it a bit more by getting rid of dependency on `List`. It turns out that for some reason repeated parameters are involved. If I change `foo` to be `def foo[A](): Foo[A] = null` then the bug goes away. 😳

gkossakowski:
I've encountered this problem in scala/scala repo. The workaround I found is to explicitly annotate return type of a `var`:

``` patch
diff --git a/src/compiler/scala/tools/nsc/typechecker/Macros.scala b/src/compiler/scala/tools/nsc/typechecker/Macros.scala
index bcf9e01..e6b14bd 100644
--- a/src/compiler/scala/tools/nsc/typechecker/Macros.scala
+++ b/src/compiler/scala/tools/nsc/typechecker/Macros.scala
@@ -496,7 +496,7 @@ trait Macros extends MacroRuntimes with Traces with Helpers {
   /** Keeps track of macros in-flight.
    *  See more informations in comments to `openMacros` in `scala.reflect.macros.whitebox.Context`.
    */
-  var _openMacros = List[MacroContext]()
[...]

PR #159, opened 2016-08-10
Split API info into another zip entry

eed3si9n:
It's been reported that for large code bases with thousands of files,
the Analysis object reaches the size of 100MB, mostly from the API info
blob.

This changes SameAPI.apply to use only the API hash, and thereby
removing the necessity to load previous API info during the incremental
compilation.

To prevent needless reading of API info, the companions map is now
stored into another zip entry, which is read lazily.

/cc @gkossakowski, @romanowski

### DIFF
--- a/internal/compiler-interface/src/main/datatype/other.json
+++ b/internal/compiler-interface/src/main/datatype/other.json
@@ -16,7 +16,7 @@
 				},
 				{
 					"name": "api",
-					"type": "Companions"
+					"type": "lazy Companions"
 				},
 				{
 					"name": "apiHash",
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalNameHashing.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalNameHashing.scala
@@ -21,7 +21,7 @@ private final class IncrementalNameHashing(log: sbt.util.Logger, options: IncOpt
   protected def invalidatedPackageObjects(invalidatedClasses: Set[String], relations: Relations,
     apis: APIs): Set[String] = {
     transitiveDeps(invalidatedClasses, logging = false)(relations.inheritance.internal.reverse) filter {
-      className => APIUtil.hasPackageObject(apis.internalAPI(className))
+      _.endsWith(".package")
     }
   }
 
--- a/internal/zinc-persist/src/main/scala/sbt/internal/inc/FileBasedStore.scala
+++ b/internal/zinc-persist/src/main/scala/sbt/internal/inc/FileBasedStore.scala
@@ -17,9 +17,9 @@ object FileBasedStore {
   private val analysisFileName = "inc_compile.txt"
   private val companionsFileName = "api_companions.txt"
 
-  def apply(file: File): AnalysisStore = new FileBasedStore0(file)
+  def apply(file: File): AnalysisStore = new FileBasedStoreImpl(file)
 
-  private final class FileBasedStore0(file: File) extends AnalysisStore {
+  private final class FileBasedStoreImpl(file: File) extends AnalysisStore {
     val companionsStore = new FileBasedCompanionsMapStore(file)
 
     def set(analysis: CompileAnalysis, setup: MiniSetup): Unit = {
@@ -30,11 +30,8 @@ object FileBasedStore {
           outputStream.putNextEntry(new ZipEntry(analysisFileName))
           TextAnalysisFormat.write(writer, analysis, setup)
           outputStream.closeEntry()
-          val apis = analysis match { case a: Analysis => a.apis }
-          val internal = apis.internal map { case (k, v) => k -> v.api }
-          val external = apis.external map { case (k, v) => k -> v.api }
           outputStream.putNextEntry(new ZipEntry(companionsFileName))
-          TextAnalysisFormat.writeCompanionMap(writer, internal, external)
+          TextAnalysisFormat.writeCompanionMap(writer, analysis match { case a: Analysis => a.apis })
           outputStream.closeEntry()
       }
     }
--- a/internal/zinc-persist/src/main/scala/sbt/internal/inc/FileBasedStore.scala
+++ b/internal/zinc-persist/src/main/scala/sbt/internal/inc/FileBasedStore.scala
@@ -6,48 +6,67 @@ package internal
 package inc
 
 import java.io._
-import java.util.zip.ZipEntry
+import java.util.zip.{ ZipInputStream, ZipEntry }
 import sbt.io.{ IO, Using }
+import xsbti.SafeLazy
 import xsbti.compile.{ CompileAnalysis, MiniSetup }
+import xsbti.api.Companions
+import scala.util.control.Exception.allCatch
 
 object FileBasedStore {
-  def apply(file: File): AnalysisStore = new AnalysisStore {
+  private val analysisFileName = "inc_compile.txt"
+  private val companionsFileName = "api_companions.txt"
 
-    private val entryFileName = "cache.txt"
+  def apply(file: File): AnalysisStore = new FileBasedStore0(file)
+
+  private final class FileBasedStore0(file: File) extends AnalysisStore {
+    val companionsStore = new FileBasedCompanionsMapStore(file)
 
     def set(analysis: CompileAnalysis, setup: MiniSetup): Unit = {
       if (!file.getParentFile.exists()) file.getParentFile.mkdirs()
       Using.zipOutputStream(new FileOutputStream(file)) {
         outputStream =>
-          val reader = new BufferedWriter(new OutputStreamWriter(outputStream, IO.utf8))
-          outputStream.putNextEntry(new ZipEntry(entryFileName))
-          TextAnalysisFormat.write(reader, analysis, setup)
+          val writer = new BufferedWriter(new OutputStreamWriter(outputStream, IO.utf8))
+          outputStream.putNextEntry(new ZipEntry(analysisFileName))
+          TextAnalysisFormat.write(writer, analysis, setup)
+          outputStream.closeEntry()
+          val apis = analysis match { case a: Analysis => a.apis }
+          val internal = apis.internal map { case (k, v) => k -> v.api }
+          val external = apis.external map { case (k, v) => k -> v.api }
+          outputStream.putNextEntry(new ZipEntry(companionsFileName))
+          TextAnalysisFormat.writeCompanionMap(writer, internal, external)
           outputStream.closeEntry()
       }
     }
 
     def get(): Option[(CompileAnalysis, MiniSetup)] =
-      try {
-        Some(getUncaught())
-      } catch {
-        case _: IOException =>
-          readOldFormat()
-        case _: Exception => None
-      }
+      allCatch.opt(getUncaught())
 
     def getUncaught(): (CompileAnalysis, MiniSetup) =
       Using.zipInputStream(new FileInputStream(file)) {
         inputStream =>
-          inputStream.getNextEntry()
+          lookupEntry(inputStream, analysisFileName)
           val writer = new BufferedReader(new InputStreamReader(inputStream, IO.utf8))
-          TextAnalysisFormat.read(writer)
+          TextAnalysisFormat.read(writer, companionsStore)
       }
+  }
 
-    def readOldFormat(): Option[(CompileAnalysis, MiniSetup)] = try
-      Some(Using.fileReader(IO.utf8)(file) { reader => TextAnalysisFormat.read(reader) })
-    catch {
-      case _: Throwable => None
+  private def lookupEntry(in: ZipInputStream, name: String): Unit =
+    Option(in.getNextEntry) match {
+      case Some(entry) if entry.getName == name => ()
+      case Some(entry)                          => lookupEntry(in, name)
+      case None                                 => sys.error(s"$name not found in the zip file")
     }
 
+  private final class FileBasedCompanionsMapStore(file: File) extends CompanionsStore {
+    def get(): Option[(Map[String, Companions], Map[String, Companions])] =
+      allCatch.opt(getUncaught())
+    def getUncaught(): (Map[String, Companions], Map[String, Companions]) =
+      Using.zipInputStream(new FileInputStream(file)) {
+        inputStream =>
+          lookupEntry(inputStream, analysisFileName)
+          val reader = new BufferedReader(new InputStreamReader(inputStream, IO.utf8))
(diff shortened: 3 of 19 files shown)


internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -85,25 +85,22 @@ private[inc] abstract class IncrementalCommon(log: sbt.util.Logger, options: Inc
...
         case APIChangeDueToMacroDefinition(src) =>
           wrappedLog.debug(s"Public API is considered to be changed because $src contains a macro definition.")
         case apiChange: NamesChange =>
           val src = apiChange.modifiedClass
           val oldApi = oldAPIMapping(src)
           val newApi = newAPIMapping(src)
-          val apiUnifiedPatch = apiDiff.generateApiDiff(src.toString, oldApi.api, newApi.api, contextSize)
-          wrappedLog.debug(s"Detected a change in a public API ($src):\n$apiUnifiedPatch")
+          wrappedLog.debug(s"""Detected a change in a public API ($src):
+                              |old hash = ${oldApi.apiHash}, new hash = ${newApi.apiHash}""".stripMargin)
```
smarter: Wait, does that mean that API diffs are no longer printed? They are _extremely_ useful when debugging/developing incremental compilation

internal/compiler-interface/src/main/datatype/other.json
```diff
@@ -16,7 +16,7 @@
 				},
 				{
 					"name": "api",
-					"type": "Companions"
+					"type": "xsbti.Maybe<Companions>"
```
gkossakowski: Why it becomes optional now?

gkossakowski:
As far as I understand, this change moves away zinc from operating on api objects. It switches solely to api hashes. It's pretty change big so the commit message should be expanded accordingly. The questions I think commit message should answer are:
- what size of the project makes API info file so large that it's a significant performance hit to load it?
- what about removed API diffs? API diffs are the key identifying tricky problems in incremental compiler, e.g.: https://github.com/sbt/zinc/issues/88 (and http://www.scala-sbt.org/0.13/docs/Understanding-Recompilation.html#Debugging+an+interface+representation)

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
eed3si9n: The previous API info is still around, just not loaded in during the compilation, so we might be able to do API diffing after the compilation is done.

internal/compiler-interface/src/main/datatype/other.json
eed3si9n: When I save to `inc_compile` I throw this away, and it thaws back as `Maybe.nothing()` value here, which should help in both time and space.

eed3si9n:
@gkossakowski Replaced the commit to based on the review.
The subprojects I'm looking at contains 1000+ source files, which then could further expand via code generation, macro, compiler plugin etc. Also it's not uncommon for Play projects to contain dozens of subprojects, which on their own could expand as teams grow.

I've brought back the API diffing, so all you have to do is enable the flag and load the full API info, which can be done via `getWithAPI` on `AnalysisStore` vs plain `get`.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -94,16 +95,24 @@ private[inc] abstract class IncrementalCommon(log: sbt.util.Logger, options: Inc
...
           wrappedLog.debug(s"Public API is considered to be changed because $src contains a macro definition.")
         case apiChange: NamesChange =>
           val src = apiChange.modifiedClass
           val oldApi = oldAPIMapping(src)
           val newApi = newAPIMapping(src)
-          val apiUnifiedPatch = apiDiff.generateApiDiff(src.toString, oldApi.api, newApi.api, contextSize)
-          wrappedLog.debug(s"Detected a change in a public API ($src):\n$apiUnifiedPatch")
+          (m2o(oldApi.api), m2o(newApi.api)) match {
+            // The API info now is optional. The caller is responsible for loading the previous Analysis
+            // with the full API info using the info stored in an APIInfoStore.
```
smarter: I think this should be done automatically all the time when debug is enabled

gkossakowski:
I did a simple test of compiling an empty class to see how much space we spend on API representation. The number I got is roughly 3.8kb. To get to 100mb of API representation you need around 25k top level classes. The API representation compresses very well using standard zip compression. I get 13.5x contraction in size.

Having said that, splitting API into a separate storage is fine. I think @smarter's latest feedback is spot on and should be addressed.

eed3si9n:
@gkossakowski @smarter Moved the "with-or-without API info" logic into the `AnalysisStore`.

romanowski:
I agree with @gkossakowski  about simple zip compression - in our we get around 10% of file caches. See my PR: #160

IMO correct way to split APIs to another files is to extract them from AnalyzedClass to separate lazy entry (map) in Analysis itself.

eed3si9n:
ok. Turned it into lazy instead of optional as per @romanowski's suggestion:

``` diff
                {
                    "name": "api",
 -                  "type": "Companions"
 +                  "type": "lazy Companions"
                },
```

This does make the whole thing more "automatic" as read-API-or-not-to-read decision is deferred to the usage point.

internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisStore.scala
```diff
@@ -6,9 +6,11 @@ package internal
 package inc
 
 import xsbti.compile.{ CompileAnalysis, MiniSetup }
+import xsbti.Maybe
 
 trait AnalysisStore {
   def set(analysis: CompileAnalysis, setup: MiniSetup): Unit
+  /** Calls `getWithAPI` if API debug is enabled, calls `getWithoutAPI` otherwise. */
```
romanowski: This comment is outdated

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalNameHashing.scala
```diff
@@ -21,7 +21,7 @@ private final class IncrementalNameHashing(log: sbt.util.Logger, options: IncOpt
   protected def invalidatedPackageObjects(invalidatedClasses: Set[String], relations: Relations,
     apis: APIs): Set[String] = {
     transitiveDeps(invalidatedClasses, logging = false)(relations.inheritance.internal.reverse) filter {
-      className => APIUtil.hasPackageObject(apis.internalAPI(className))
+      _.endsWith(".package") // APIUtil.hasPackageObject(apis.internalAPI(className))
```
romanowski: Please remove commented code

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileBasedStore.scala
```diff
@@ -7,17 +7,53 @@ package inc
...
+      Using.fileWriter(IO.utf8)(analysisFile) { writer => TextAnalysisFormat.write(writer, analysis, setup) }
     }
 
     def get(): Option[(CompileAnalysis, MiniSetup)] =
       try { Some(getUncaught()) } catch { case _: Exception => None }
+
     def getUncaught(): (CompileAnalysis, MiniSetup) =
-      Using.fileReader(IO.utf8)(file) { reader => TextAnalysisFormat.read(reader) }
+      {
+        val (analysis0: Analysis, setup) = Using.fileReader(IO.utf8)(analysisFile) { reader => TextAnalysisFormat.read(reader) }
```
romanowski: I think all this logic should be moved to  APIsF's read and write methods (in TextAnalysisFormat.scala) not to construct twice AnalyzedClass instance.

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileBasedStore.scala
```diff
@@ -7,17 +7,53 @@ package inc
...
+              k -> (new AnalyzedClass(v.compilation, v.name,
+                SafeLazy(apiinfo.get.external(k).api), v.apiHash, v.nameHashes, v.hasMacro))
+          }
+        )
+        val analysis = analysis0.copy(apis = apis)
+        (analysis, setup)
+      }
+  }
+
+  private final class FileBasedAPIInfoStore0(file: File) {
```
romanowski: Why 0?

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileBasedStore.scala
```diff
@@ -7,17 +7,53 @@ package inc
...
 import xsbti.compile.{ CompileAnalysis, MiniSetup }
+import xsbti.api.{ AnalyzedClass, Lazy }
 
 object FileBasedStore {
-  def apply(file: File): AnalysisStore = new AnalysisStore {
+  def apply(analysisFile: File, apiInfoFile: File): AnalysisStore =
+    new FileBasedStore0(analysisFile, apiInfoFile)
+
+  private final class FileBasedStore0(analysisFile: File, apiInfoFile: File) extends AnalysisStore {
+    val apiInfoStore = new FileBasedAPIInfoStore0(apiInfoFile)
```
romanowski: I think this is great candidate for 2nd entry in zip... ( #160 )

internal/zinc-persist/src/main/scala/sbt/internal/inc/TextAnalysisFormat.scala
```diff
@@ -266,6 +279,28 @@ object TextAnalysisFormat {
     }
   }
 
+  private[this] object FullAPIsF {
```
romanowski: I think we should store there only companions - we don't want to store compilations and name hashes twice (especially name hashes are heavy)

internal/zinc-persist/src/main/scala/xsbt/api/AnalyzedClassFormat.scala
```diff
@@ -3,20 +3,34 @@
...
-      val oin = new ObjectInputStream(new InputWrapperStream(in))
-      try { oin.readObject.asInstanceOf[AnalyzedClass] } finally { oin.close() }
-    }
-  def writes(out: Output, src: AnalyzedClass): Unit = {
-    val oout = new ObjectOutputStream(new OutputWrapperStream(out))
-    try { oout.writeObject(src) } finally { oout.close() }
-  }
+object AnalyzedClassFormats {
+  // This will throw out API information intentionally.
+  def lightFormat(implicit ev0: Format[Compilation], ev1: Format[NameHashes]): Format[AnalyzedClass] =
```
romanowski: If we store just companions in 2nd store then we can use single format

internal/zinc-persist/src/main/scala/xsbt/api/CompanionsFormat.scala
```diff
@@ -0,0 +1,16 @@
+package xsbt.api
+
+import xsbti.api._
+import sbinary._
+
+object CompanionsFormat extends Format[Companions] {
```
romanowski, line 6: If we move all compatnions to single store then we can serialized whole map not single companion one by one. I think we should let Java optimized our serialization

eed3si9n:
@romanowski Thanks for the detailed review. I did adopt most of your suggestions, and pushing the responsibility down to TextAnalysisFormat did make the imple much cleaner. I also made the Companions storage another zip entry.

internal/zinc-persist/src/main/scala/xsbt/api/CompanionsFormat.scala
eed3si9n: I want to rely less on Java serialization.. eventually I want to turn all these into JSON so we can cross machine boundaries safely.

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileBasedStore.scala
```diff
@@ -6,48 +6,67 @@ package internal
...
 
 object FileBasedStore {
-  def apply(file: File): AnalysisStore = new AnalysisStore {
+  private val analysisFileName = "inc_compile.txt"
+  private val companionsFileName = "api_companions.txt"
 
-    private val entryFileName = "cache.txt"
+  def apply(file: File): AnalysisStore = new FileBasedStore0(file)
+
+  private final class FileBasedStore0(file: File) extends AnalysisStore {
```
romanowski: maybe `ZippedStore`?

eed3si9n: The whole point is that I'm mimicking the `object FileBaseStore` so I think that's fine. It's only so the stack trace looks nicer.

romanowski: I don't want to nitpick but I just don't like 0s in class names :)

eed3si9n: How about `Impl`?

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileBasedStore.scala
```diff
@@ -6,48 +6,67 @@ package internal
...
-          TextAnalysisFormat.write(reader, analysis, setup)
+          val writer = new BufferedWriter(new OutputStreamWriter(outputStream, IO.utf8))
+          outputStream.putNextEntry(new ZipEntry(analysisFileName))
+          TextAnalysisFormat.write(writer, analysis, setup)
+          outputStream.closeEntry()
+          val apis = analysis match { case a: Analysis => a.apis }
+          val internal = apis.internal map { case (k, v) => k -> v.api }
+          val external = apis.external map { case (k, v) => k -> v.api }
+          outputStream.putNextEntry(new ZipEntry(companionsFileName))
+          TextAnalysisFormat.writeCompanionMap(writer, internal, external)
```
romanowski: can we pass whole APIs class here and extract this at the last possible place (CompanionsF IIRC)?

It should make code nicer.

eed3si9n: Sure.

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileBasedStore.scala
romanowski: :+1:

romanowski:
Except minor comments above LGTM

PR #164, opened 2016-08-25
Add callback allowing to switch to single files compilation mode

pkukielka:

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -74,8 +74,7 @@ object Incremental {
       log.debug("All initially invalidated classes: " + initialInvClasses + "\n" +
         "All initially invalidated sources:" + initialInvSources + "\n")
       val analysis = manageClassfiles(options) { classfileManager =>
-
-        incremental.cycle(initialInvClasses, initialInvSources, sources, binaryChanges, previous, doCompile, classfileManager, 1)
+        incremental.cycle(initialInvClasses, initialInvSources, sources, binaryChanges, lookup, previous, doCompile, classfileManager, 1)
       }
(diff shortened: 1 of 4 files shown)


romanowski:
LGTM

Duhemm:
Am I right that the goal is to bypass incremental recompilation entirely and recompile only the files that were modified? How is this different from [AntStyle incremental compilation](https://github.com/sbt/zinc/blob/1.0/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalAntStyle.scala)?

pkukielka:
@Duhemm 
You are right about the goal, but we cannot use AntStyle builder. 
This pull request (as several others before and after it) is aiming to bring zinc to IntelliJ (you will be able to choose zinc as default compiler). So we need normal incremental compilation (including name hashing) which only sometimes need to be switched to single files compilation (namely when someone uses 'Compile'). Moreover if after compiling single file user will hit 'Make', zinc should have correct analysis in place and continue compilation from there.

romanowski:
Of course Intellij needs to handle all files that should be recompiled due to possible changes in compiled source (that is why `incrementallyInvalided` is passed to hook.

Duhemm:
@pkukielka I see, thanks for the explanations. The only thing that worries me is that the incremental compiler is starting to have 2 different ways of being configured: `IncOptions` and `ExternalHooks`. Any reason not to put this setting in particular in `IncOptions`?

romanowski:
External hooks are part of IncOptions - however this is Java code (with possible nulls) so that is why it is also accessible by Lookup instance.

https://github.com/sbt/zinc/blob/1.0/internal/compiler-interface/src/main/datatype/incremental.json#L145-L149
https://github.com/sbt/zinc/blob/1.0/internal/compiler-interface/src/main/java/xsbti/compile/IncOptionsUtil.java#L87-L99

eed3si9n:
LGTM

PR #165, opened 2016-08-26
Speed up external dependencies lookup

romanowski:
Previously in order to find Analysis for given class name zinc fist have to enumerated whole classpath (each class in each jar or classes directory). In huge workspace it might take over 40 sec – and this was overhead for each compilation (even the smallest).

Now zinc searches for given analysis in all analysis ordered in the same way as classpath so whole classpath enumeration is skipped for searching for changed sources. Now overhead is barely  noticeable.

This might not detect correct changes if class x.X was previously present only in ouput of depended project but later class with same binary name was added to raw classes directory that is earlier on classpath. This is true if this situation happens when jar content is changed (so x.X is added there) however this should be capture as change in binary dependencies.

From reasons above I deiced to keep old behavior under flag.

### DIFF
--- a/internal/compiler-interface/src/main/datatype/incremental.json
+++ b/internal/compiler-interface/src/main/datatype/incremental.json
@@ -146,6 +146,14 @@
           "name": "externalHooks",
           "type": "xsbti.compile.ExternalHooks",
           "doc": "External hooks that allows clients e.g. IDEs to interacts with compilation internals"
+        },
+        {
+          "name": "analysisOnlyExtDepLookup",
+          "type": "boolean",
+          "doc": [
+            "Ignore classpath entries without Analysis when searching for external dependencies.",
+            "It might miss changes in classes from jars external class directories."
+          ]
         }
       ]
     },
--- a/internal/zinc-persist/src/test/scala/sbt/inc/TextAnalysisFormatSpecification.scala
+++ b/internal/zinc-persist/src/test/scala/sbt/inc/TextAnalysisFormatSpecification.scala
@@ -54,7 +54,6 @@ object TextAnalysisFormatTest extends Properties("TextAnalysisFormat") {
     TextAnalysisFormat.write(writer, analysis, commonSetup)
 
     val result = writer.toString
-    // println(result)
 
     val reader = new BufferedReader(new StringReader(result))
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 target/
+.idea
--- a/internal/compiler-interface/src/main/java/xsbti/compile/IncOptionsUtil.java
+++ b/internal/compiler-interface/src/main/java/xsbti/compile/IncOptionsUtil.java
@@ -102,6 +102,10 @@ public class IncOptionsUtil {
     return true;
   }
 
+  public static boolean defaultAnalysisOnlyExtDepLookup() {
+    return false;
+  }
+
   public static IncOptions defaultIncOptions() {
     IncOptions retval = new IncOptions(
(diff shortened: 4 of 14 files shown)


gkossakowski:
Is that related to sbt/sbt#2525 ?

romanowski:
Yes. Basically I needed to 'compile' when nothing (or almost nothing) is changed as main goal.
We are using zinc as builder in Intellij so we can reuse some data that Intellij stores.

Initally our overhead was >200 sec for fist and >150 for follwing no-op compilation (lower bound is around 7-10 sec).

After removing all integration problems and introduce full analysis caching I go down to ~160/120 sec

After that I consume hooks (https://github.com/sbt/zinc/pull/161):
- Intellij tracking changed sources for me
- I ignore products since Intellij also ignores them
- I am caching modification dates for binary deps for single compilation (so I check e.g. scala-libary.jar once)

This reduce times to 100/60 sec.

This changes allows me to go down to 55/15 sec - those are times I needed.

internal/compiler-interface/src/main/datatype/incremental.json
```diff
@@ -146,6 +146,14 @@
           "name": "externalHooks",
           "type": "xsbti.compile.ExternalHooks",
           "doc": "External hooks that allows clients e.g. IDEs to interacts with compilation internals"
+        },
+        {
+          "name": "fastExtDependenciesLookup",
```
eed3si9n: I think we need a better name than "fast", since you might come up with even faster ideas, and then we'd have to call it "superFast".

internal/zinc-persist/src/test/scala/sbt/inc/TextAnalysisFormatSpecification.scala
```diff
@@ -51,12 +51,16 @@ object TextAnalysisFormatTest extends Properties("TextAnalysisFormat") {
     TextAnalysisFormat.write(writer, analysis, commonSetup)
 
     val result = writer.toString
-    // println(result)
 
     val reader = new BufferedReader(new StringReader(result))
 
     val (readAnalysis: Analysis, readSetup) = TextAnalysisFormat.read(reader)
 
+    println(result.take(commonHeader.size))
```
eed3si9n: Was this intentional?

romanowski: no, for sure not :)

fommil:
How does this behave in sbt if the project is using `exportJars`?

romanowski:
@fommil I don't know how this is exactly implemented in sbt but in theory sbt should provide lookup that will point to Analysis for exported jar - so behaviour shouldn't change. If Analysis is not found for exported jars so this is bug in zinc - sbt integration.
I never explicitly check if I am working with class directory or jar. I only rely on presence of Analysis for given classpath entry.

eed3si9n:
I wonder if you can make it safer without sacrificing on the perf gains you made. For example can you store SHA1 hash of the JAR files in Analysis and skip on if they haven't changed since the last `compile`? If it's unable to detect the change in library dependencies, I don' think we should have this on by default.

romanowski:
That is why I create this issue: https://github.com/sbt/zinc/issues/166 since this is exacly our problem here.
I will switch `analysisOnlyExtDepLookup` to false by default - so we can merge it now.

eed3si9n:
> I will switch analysisOnlyExtDepLookup to false by default

Sounds good.

romanowski:
@eed3si9n can we merge it now? :)

eed3si9n:
Merged

PR #171, opened 2016-08-31
Fix split zip reading

eed3si9n:


romanowski:
I should check before creating my PR :) 

Can also fix typo here: https://github.com/romanowski/zinc/blob/da0712dfe07b84df3dcd579708ca33b89a8b2056/internal/zinc-persist/src/main/scala/sbt/internal/inc/TextAnalysisFormat.scala#L291-L292 ?

except that LGTM and I will close my PR

eed3si9n:
Fixed the typo.

Issue #166, opened 2016-08-26
Zinc should detect shadowing classes from binary deps by classes from ext deps 


romanowski:
During implementation of https://github.com/sbt/zinc/pull/165 I noticed that if I have project (proj1) with classpath `Seq(scalaJars, proj1_out, proj2_out, jar1.jar)` and jar1.jar contains class foo.Bar and I create class foo.Bar (I know I shouldn't :) ) in proj2_out then zinc will not recompile any sources from proj1 that reference to foo.Bar (but it should).

Other way (change from ext to binary) is (was) supported but is slows down searching for changed classes (see https://github.com/sbt/zinc/pull/165 for more details).

romanowski:
We are creating unit tests for this - PR will be created soon

gkossakowski:
> Other way (change from ext to binary) is (was) supported but is slows down searching for changed classes (see #165 for more details).

What does this mean? Does the "other way" refer to behavior before #165 got merged?

romanowski:
Tests added in https://github.com/sbt/zinc/pull/175 clearly shows that is happens only with analysisOnlyExtDepLookup set to true. I am thinking on reverting https://github.com/sbt/zinc/pull/165

eed3si9n:
Added a test in #179 using actual files, and it seems to be happening without the flag.

eed3si9n:
So we want correctness in terms of respecting classpath, but we also want to avoid looking up classpath all the time. Suppose we have the following classpath:

```
lib/scala-library.jar, sub1/target, sub2/target, lib/sample-binary_2.11-0.1.jar
```

From here I can check whether they have an Analysis associated.

```
None, Some(a1), Some(a2), None
```

And then, we can check the last timestamp. Suppose shadowing of a class happened in `sub2/target` since the last `sub1` compilation it would be:

```
None, Some(t), Some(t + 10s), None
```

This can prompt the duplication check between `sub2/target`'s product and the `lib/sample-binary_2.11-0.1.jar`. If no compilation has happened, then it would be:

```
None, Some(t), Some(t), None
```

This also allows for bump up of JAR version, which still would be:

```
None, Some(t), Some(t), None
```

Can this work?

eed3si9n:
To detect whether shadowing had effect on `sub1`, I'd now need to track `libraryClassNames: Relation[File, String]`. (The relationship named `binaryClassName` exists for different purpose.)

romanowski:
[...]

PR #176, opened 2016-09-20
Revert #165 but keep some of the changes

eed3si9n:
Revert #165 but keep some of the changes
Ref #166

I'm looking into sbt/sbt#2525 changes, so I'm going to roll back the core changes in #165 first.

/review @romanowski


zinc/src/test/scala/sbt/inc/LookupSpec.scala
```diff
@@ -93,17 +67,19 @@ class LookupSpec extends BaseIncCompilerSpec {
...
-    import setup._
+  // it should "find correct analysis for class file" in mockedProjectContext { setup =>
+  //   import setup._
 
-    lookup.lookupAnalysis(project1.P1P2.classFile) shouldEqual project1.analysis
-    lookup.lookupAnalysis(project2.P1P2.classFile) shouldEqual project2.analysis
-    lookup.lookupAnalysis(project2.P2.classFile) shouldEqual project2.analysis
-    lookup.lookupAnalysis(project1.P1J2.classFile) shouldEqual project1.analysis
-    lookup.lookupAnalysis(project1.P1C2.classFile) shouldEqual project1.analysis
+  //   val lookup = new LookupImpl(config)
```
romanowski: Why this is commented out? It should work.

eed3si9n: It doesn't seem to work. 
https://github.com/eed3si9n/zinc/blob/92fd371d052236be2cdfe04755c3130b90cf920f/zinc/src/test/scala/sbt/inc/BaseIncCompilerSpec.scala#L106
VirtualProject uses `out` as `classpathEntry`, but this test is passing along `project1.P1P2.classFile`. For this to work `entryLookup` implementation should be changed to inspect the `Analysis` too.

eed3si9n: Added 9dabad4bec0ec02777a238311ee552c72a84a751 to fix this.

lightbend-cla-validator:
Hi @tgodzik,

eed3si9n:
I've ported the new tests from #175 minus the flag.

zinc/src/test/scala/sbt/inc/LookupSpec.scala
romanowski: https://github.com/sbt/zinc/blob/1.0/internal/zinc-core/src/main/scala/sbt/internal/inc/Lookup.scala#L33 It says that this is class file.

When class foo.A is used in code that comes from directoty (not jar) IncrementalCommon.externalBinaryModify calls lookupAnalysis with class file not directory. 

Tested in zinc-intellij integration.

eed3si9n: I think we agree. I just changed the test's implementation of Analysis lookup for make sure it can be looked up with class file.

romanowski: I don't think 9dabad4 is a proper fix. PerClasspathEntryLookup is documented to work with cp entry but Lookup should work with classes. 

I think we should revert only lines 35-36 in LookupImpl.scala from my implementation.

eed3si9n: In this PR I just wanted to revert #165 but keep the tests. In #177  we can fix the `LookupImpl` and `PerClasspathEntryLookup`.

zinc/src/test/scala/sbt/inc/BaseIncCompilerSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  val analysis = in.newSource(this)
+}
+
+class DirectorySetup(val baseDir: File) {
+  private var myEntries = Seq.empty[ClasspathEntry]
+
+  def register(entry: ClasspathEntry) = myEntries :+= entry
+
+  def classpath = myEntries.map(_.classpathEntry)
```
eed3si9n: I wonder if we're making our lives more difficult than necessary by allowing arbitrary ordering of Analysis and external JARs.

romanowski: I think we should support any order of classpath - so unless we are not forcing classpath order we need to support each possibility

eed3si9n: I was talking to @gkossakowski, and his recommendation is to invalidate everything when we detect change in classpath. Do you do anything that changes the classpath frequently?

romanowski: TBH it is pretty common case - e.g. update library version change.

stuhood:
Yea, that's a common case.

eed3si9n:
Added more comment on this here - https://github.com/sbt/zinc/issues/166#issuecomment-249277024

PR #179, opened 2016-09-22
Add MultiProjectIncrementalSpec to test shadowing detection

eed3si9n:
Ref #166 
This is on top of #176

Instead of the virtual project, this uses actual files and JAR to test the shadowing detection.
## steps

There are two subprojects sub1 and sub2. sub1 depends on sub2. The test consists of three phases:
1. sub1 is compiled with `src/Depender.scala`, and uses empty `sub2/target/` directory and `lib/sample-binary_2.11-0.1.jar` as the classpath. `Depender` calls `test.pkg.Ext1.x`, which is defined in the JAR file.
2. sub1 is compiled next with `src/Depender.scala` and `src/Depender2.scala`, and uses empty `sub2/target/` directory and `lib/sample-binary_2.11-0.1.jar` as the classpath. `Depender2` calls `test.pkg.Ext2.x`, which is defined in the JAR file.
3. Next, sub2 is compiled with `src/Ext1.scala`.
4. sub1 is compiled again with `src/Depender.scala`, `src/Depender2.scala`, and `src/Good.scala`. This should detect that `test.pkg.Ext1` is now shadowed by `sub2/target`, and thus `src/Depender.scala` must be invalidated.
## problem

ff75cd1c5512817c4a921fcc52a6eeb5196625ce demonstrates that current implementation does not detect this shadowing.


gkossakowski:
Should you mark your test as pending and then mark it as passing in #180? Otherwise, LGTM.

PR #181, opened 2016-09-27
Use customized fileManager to notify when .class files are generated

peiyuwang:
This is a follow up PR for #94 for the Java use case. See [1] for the original discussion.

Transaction rollback handling is still through `ClassFileManager`, except now writes to generated class files can be reported through `WriteReportingFileManager` incrementally. 

Changes summary:
- `WriteReportingFileManager`, `WriteReportingFileObject` and `WriteReportingFileObject` for recording `.class` file writes.
- Merge two `ClassFileManager` interfaces (one java and one scala) into one `ExternalHooks. ClassFileManager` so it can be added to `JavaTool.run` interface
- Everything else is to pass `ClassFileManager` to javac, mostly straightforward, only one worth mentioning is move `doCompile` one level closer to where `ClassFileManager` transaction is created, from `IncrementalCompile` to `Incremental`, in order to take the extra parameter. Because each compile cycle needs to report its own `AnalysisCallback`, an `AnalysisCallback.Builder` is also introduced.

Testing:
- travis-ci
- Have a local integration test that injects an exception (d8ae7068737474e0a86f11265fef17a06be6435), verified the expected behavior.
- Did another local integration in twitter repo for an actual failure in Analysis, observed rollback actually happened.

TODOs:
- Do this also for scalac, this can be a separate change.
- Deprecate forked java usage, this is again a different change [2]

[1] https://github.com/sbt/zinc/pull/94#issuecomment-209325666
[2] https://groups.google.com/d/topic/sbt-dev/j4uFWvRW-OA/discussion

cc @stuhood @eed3si9n @gkossakowski


zinc-compile/src/test/scala/inc/DocSpec.scala
```diff
@@ -16,7 +16,7 @@ class DocSpec extends UnitSpec {
     IO.withTemporaryDirectory { cacheDir =>
       IO.withTemporaryDirectory { out =>
         val javadoc = Doc.cachedJavadoc("Foo", cacheDir, local)
-        javadoc.run(List(knownSampleGoodFile), Nil, out, Nil, log, reporter)
+        javadoc.run(List(knownSampleGoodFile), Nil, out, Nil, null, log, reporter)
```
eed3si9n: Our public API shouldn't expect null.

zinc/src/main/scala/sbt/internal/inc/javac/AnalyzingJavaCompiler.scala
```diff
@@ -35,11 +38,13 @@ final class AnalyzingJavaCompiler private[sbt] (
...
    * @param output   The output configuration for this compiler
    * @param callback  A callback to report discovered source/binary dependencies on.
+   * @param classfileManager The component that manages generated class files.
    * @param reporter  A reporter where semantic compiler failures can be reported.
    * @param log       A place where we can log debugging/error messages.
    * @param progressOpt An optional compilation progress reporter.  Where we can report back what files we're currently compiling.
    */
-  def compile(sources: Seq[File], options: Seq[String], output: Output, callback: AnalysisCallback, reporter: Reporter, log: Logger, progressOpt: Option[CompileProgress]): Unit = {
+  def compile(sources: Seq[File], options: Seq[String], output: Output, callback: AnalysisCallback,
+    classfileManager: ClassFileManager, reporter: Reporter, log: Logger, progressOpt: Option[CompileProgress]): Unit = {
```
eed3si9n, line 46: If this is not something that changes between the compilations, it might make sense to put the classfileManager in the constructor parameter along with thing like `scalaInstance`.

internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java
```diff
@@ -1,12 +1,31 @@
...
  */
 public interface ExternalHooks {
     public static interface Lookup {}
 
-    public static interface ClassFileManager{}
+    /**
+     * During an incremental compilation run, a ClassfileManager deletes class files and is notified of generated class files.
+     * A ClassfileManager can be used only once.
+     */
+    public static interface ClassFileManager {
```
eed3si9n: Let's move this class out of `ExternalHooks` if it's being used outside of ExternalHooks.

internal/compiler-interface/src/main/java/xsbti/compile/JavaTool.java
```diff
@@ -21,9 +21,11 @@
    *
    * @param sources  The list of java source files to compile.
    * @param options  The set of options to pass to the java compiler (includes the classpath).
+   * @param classFileManager The manager for generated class files.
    * @param reporter The reporter for semantic error messages.
    * @param log      The logger to dump output into.
    * @return true if no errors, false otherwise.
    */
-  boolean run(File[] sources, String[] options, Reporter reporter, Logger log);
+  boolean run(File[] sources, String[] options, ExternalHooks.ClassFileManager classFileManager,
```
eed3si9n: It is necessary to add `classFileManager` as a parameter here? If the majority of the time it's passed null and if it doesn't change that much, I feel like it belongs more to the setup of the surrounding class, unless I'm not understanding something.

eed3si9n (changes requested):
I'm fine with pushing ClassFileManager to Java layer, but I'm not sure about some the details.

zinc/src/main/scala/sbt/internal/inc/javac/AnalyzingJavaCompiler.scala
peiyuwang: iiuc `classfileManager` lives within the scope of incremental compile run of multiple cycles: it will start with an empty set of generated class files and collect them.

https://github.com/sbt/zinc/blob/4bc93500d2ce98da15f61fdc291aacceee4f440e/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala#L76-L78

Therefore its lifespan is not attached to constructor but to each incremental compile call.

zinc-compile/src/test/scala/inc/DocSpec.scala
peiyuwang: One option I considered is to have a `noop` `ClassFileManager` to pass around when no work is to be done.

Two options to do this: 
- in `ClassfileManager` similar to the existing predefined `ClassFileManager`:`deleteImmediately`
- As a default option in `compiler-interface` along with `ClassFileManager` is defined

I slightly prefer the former, just to keep the interface _interface_, that will introduce a dependency of `zincCore` to `zinc-compile` (to use `noop` in `DocSpec`), is that ok?

peiyuwang: Fixed by introducing `NoopClassFileManager`, since it's used only by test, placed it under `zinc-compile-core/../test`. Similarly `CollectingClassFileManager` is also used only for testing.

internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java
peiyuwang: Done, moved out of `ExternalHooks`

internal/compiler-interface/src/main/java/xsbti/compile/JavaTool.java
peiyuwang: Ditto on this, see my earlier reply.

A new `classFileManager` is needed per incremental compile (shared by multiple cycles)

In a non-test environment, majority of time `classFileManager` is not null, it is `deleteImmediately` since transaction is only opted-in (by turning on `TransactionalManagerType` `IncOption`, in which case `deleteImmediately.generated` call is no-op.

[1] https://github.com/sbt/zinc/blob/057f52e4f6791c03c3f39cfe1e658f8f72c575da/internal/zinc-core/src/main/scala/sbt/internal/inc/ClassfileManager.scala#L54

eed3si9n (approved):

dwijnand:
Just FYI @eed3si9n, merging this broke https://jenkins.scala-sbt.org:8592/job/sbt-standalone-build/40/console

PR #183, opened 2016-10-16
Adding column number information into compiler output

VlachJosef:
Currently compiler output lacks column number information easily
accessible for cli tools.

This change is adding column number just behind line number

```
[error] /Users/me/column/src/main/scala/Column.scala:5:31: type mismatch;
```

Initial PR was against 0.13.13 branch https://github.com/sbt/sbt/pull/2785, but after discussion was moved here.


dwijnand (approved):

eed3si9n (approved):

ornicar:
Unfortunately the `extraLoggers` don't get access to the column number. This breaks tools that use an extra logger to jump to the line and position of an error.

Is there a way around this?

console output: `LilaWsServer.scala:26:39: ';' expected but ']' found.`
extra logger output: `LilaWsServer.scala:26: ';' expected but ']' found.`

PR #186, opened 2016-10-20
Move NoopClassFileManager out of test for other non-javac use cases

peiyuwang:
This should fix https://ci.twitter.biz/job/pants-performance-compile-gizmoduck-server-tests/2843/consoleFull

> a48811eb72c45ef97990036ebe345065994d0919/main/src/main/scala/sbt/Defaults.scala:856: not enough arguments for method run: (sources: List[java.io.File], classpath: List[java.io.File], outputDirectory: java.io.File, options: List[String], classFileManager: xsbti.compile.ClassFileManager, log: sbt.util.Logger, reporter: xsbti.Reporter)Unit.
> [sbt] [error] Unspecified value parameter reporter.
> [sbt] [error]           javadoc.run(srcs.toList, cp, out, javacOptions.value.toList, s.log, reporter)

Test:
- https://travis-ci.org/peiyuwang/zinc/builds/169135437 is green
- Also locally tested sbt by running `sbt test`, previously broken with above error

### DIFF
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/NoopClassFileManager.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/NoopClassFileManager.scala
@@ -1,0 +1,14 @@
+package sbt.internal.inc
+
+import java.io.File
+
+import xsbti.compile.ClassFileManager
+
+/**
+ * A noop ClassFileManager that applies to non-javac tools.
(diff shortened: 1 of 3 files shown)


peiyuwang:
https://github.com/sbt/sbt/pull/2791 is up for the downstream side

eed3si9n (approved):

PR #187, opened 2016-10-20
Zinc extract used names is very sluggish

wpopielarski:
Compilation of shapeless with Zinc is very slow. Potential reason lays in
multiple visit in already traversed tree.

Please look at least at test. This file should be processed in about 5 sec. but actually takes about 5 mins.


internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
```diff
@@ -86,16 +86,14 @@ class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) ext
...
-        handleClassicTreeNode(tree)
-        handleMacroExpansion(original)
-        super.traverse(tree)
-      case _ =>
-        handleClassicTreeNode(tree)
+    override def traverse(tree: Tree): Unit = {
+      if (!inspectedTrees.contains(tree)) {
+        if (handleTree(tree))
+          inspectedTrees += tree
         super.traverse(tree)
```
eed3si9n, line 95: Does this handle macro expansion?

eed3si9n (changes requested):

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
wpopielarski: sorry @eed3si9n it's not. Going to fix. Looks like there is no test to discover it.

retronym:
I would suggest to incorporate https://github.com/sbt/sbt/pull/2754 into Zinc (I've lodged https://github.com/typesafehub/zinc/issues/105 to track this work). When I profiled the slowness of Shapeless with SBT + Scala 2.12, I initially was looking at ways to optimize name extraction but eventually found out that the slowness was (mostly) due to the poor interaction with JIT.

retronym:
Actually, I see now that https://github.com/sbt/zinc/commit/33e9a507874415d4c851b8583c3387a4623b7b33 appears to introduce exponential performance by (accidentally?) introducing `tt.original.foreach(traverse)` instead of `tt.original.foreach(handleNode)`. Deeply nested ASTs in `TypeTree#original` (as you would see in long HLists types) will lead you into the problem.

AFAICT, this problem is only introduces in sbt/zinc, not in the incremental compiler in SBT 0.13.x

eed3si9n:
@retronym Thanks for the detective work.

eed3si9n:
I opened https://github.com/sbt/zinc/pull/193 as a replacement for this PR.

wpopielarski:
superseded by #193

Issue #185, opened 2016-10-19
Allow optionally turning off customized file manager


peiyuwang:
#181 uses a VFS approach to protect `.class` files generation inside a transaction. This in theory should work for all files generated by javac, however, while it works for `.class` files, which is the majority case, `resource` (could be projects using annotation processor) and `.java` (An example is jmh that first generates `.java` files, then compiles them into `.class` files) files currently fail under this approach, see stack trace in [1].

The root cause is a current javac limitation, [somewhere in `JavacFileManager`](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b27/com/sun/tools/javac/file/JavacFileManager.java#638) there is a hard coded `instanceof` check, so when `FileObject` or `JavaFileObject` are wrapped into `ForwardingFileObject`/`ForwardingJavaFileObject`, instanceof check would fail because the check is on the forwarding object, not its wrapped object.

```
public boolean isSameFile(FileObject a, FileObject b) {
       nullCheck(a);
       nullCheck(b);
       if (!(a instanceof BaseFileObject))
           throw new IllegalArgumentException("Not supported: " + a);
       if (!(b instanceof BaseFileObject))
           throw new IllegalArgumentException("Not supported: " + b);
       return a.equals(b);
   }
```

Assume upstream change (I opened a request internally at Twitter) will take some time, we need some workaround to turn off the customized manager if project uses jmh for example.

A few options to introduce this option:
- Introduce a new boolean inc option. Downside this eventually will be removed, likely will touch a few layers in order to pass to `LocalJavaCompiler`
- Reuse the existing inc option `classfileManagerType`, so we only invoke VFS for `TransactionalManagerType`. Downside is `ClassFileManager` likely needs to a new method to optionally wrap an existing FileManager into `ForwardingFM`
- Pass an environment variable w/o touching any interfaces.

The third option seems the cleanest for such an use case. @eed3si9n what do you think? any other ideas?

[1] Compiling a jmh project

```
java.lang.RuntimeException: java.lang.IllegalArgumentException: Not supported: sbt.internal.inc.javac.WriteReportingJavaFileObject@1cfd1875
    at com.sun.tools.javac.main.Main.compile(Main.java:559)
    at com.sun.tools.javac.api.JavacTaskImpl.doCall(JavacTaskImpl.java:129)
    at com.sun.tools.javac.api.JavacTaskImpl.call(JavacTaskImpl.java:138)
[...]

PR #188, opened 2016-10-25
Optionally turning on customized file manager

peiyuwang:
This addresses #185 

Implementation detail:

Adding an additional parameter is a little clunky, rather, this PR combines existing param `classFileManager` and the new boolean `useCustomizedFileManager` into `IncToolOptions` because both are for inc compile used for java tools.

As a by product, by using `Maybe<ClassFileManager>` allows to kill `NoopClassFileManager`.

If this looks ok I will make the downstream change in https://github.com/sbt/sbt, probably replace https://github.com/sbt/sbt/pull/2791


peiyuwang:
@eed3si9n can you review this? mostly done with testing this change (along with the earlier #181 ), ready to ask for a new release if this PR goes in

eed3si9n:
So the title of this PR should be flipped to optionally turning "on" if the default is off?

peiyuwang:
yea, :) fixed

eed3si9n:
Merged! Thanks for the contribution :)

peiyuwang:
cool thanks!

I will update https://github.com/sbt/sbt/pull/2791

PR #189, opened 2016-10-25
Forward port of sbt/sbt#2767

avdv:
This avoids an NPE when accessing position info in case `sourcePath` or `sourceFile` are `null`. See sbt/sbt#2766 for details.


stuhood:
Thanks for this.

PR #190, opened 2016-10-31
Add props to define source dependencies on io/util/lm

dwijnand:
Use either -Dsbtio.path / -Dsbtutil.path / -Dsbtlm.path on the command line
or sbtio.path= / sbtutil.path= / sbtlm.path in project/local.properties


dwijnand:
I'm sure if there's a problem here we can forward-fix or revert.

PR #197, opened 2016-11-17
Retab & compactify the sbt-datatype definitions

dwijnand:


dwijnand:
The diff says +563/−975.

But if you drop the whitespace changes it's actually +137/−549, so I'd recommend reviewing that diff:

https://github.com/sbt/zinc/pull/197/files?w=1

eed3si9n (approved):

PR #198, opened 2016-11-18
Move to new sbt/librarymanagemento

dwijnand:
Currently zincIvyIntegration and zinc tests fail..

Something about not being able to find xsbti classes (which come from util-interface).

Ideas to unblock this - very welcome. Thanks.


eed3si9n:
It seems like you published https://github.com/sbt/util/releases/tag/v0.1.0-M15 but didn't sync it to Maven Central - https://repo1.maven.org/maven2/org/scala-sbt/util-interface/
So you're hitting a legitimate issue. You need to sync it from https://bintray.com/sbt/maven-releases/util/0.1.0-M15#central

dwijnand:
Done. But it still hasn't fixed the issue..

internal/compiler-bridge/src-2.10/main/scala/xsbt/API.scala
```diff
@@ -1,81 +0,0 @@
-/* sbt -- Simple Build Tool
```
eed3si9n, line 1: Why is the compiler bridge being dropped for 2.10?

eed3si9n (changes requested):

internal/compiler-bridge/src-2.10/main/scala/xsbt/API.scala
dwijnand: Because there's no sbt/util for 2.10 https://github.com/sbt/util/pull/44/commits/4e4aa08a1a13bd85c70d60d7129438acd1819406

eed3si9n: We should cross publish util-logging and util-testing for 2.10 if that's what it takes.

PR #202, opened 2016-11-26
Log at info level when compilation completes

cb372:
So this is not the most exciting PR ever, but...

Currently zinc outputs a log before it starts compilation:

```
[info] Compiling 1 Scala source to /Users/chris/tmp/foo/target/scala-2.11/test-classes...
```

but it doesn't log anything (except for timing logs at debug level) when compilation completes. As we all know, compiling Scala code is a potentially long-running operation, so I think it's worth logging on completion.

Up for debate: should this log be skipped if the compiler returned errors? I don't really have a strong opinion either way.

## Background

I was bitten by the lack of this log recently when I tried to run a buggy test containing an infinite loop. sbt printed output like this and then hung:

```
> test
[info] Updating {file:/Users/chris/tmp/sbt-example/}sbt-example...
[info] Resolving jline#jline;2.12.1 ...
[info] Done updating.
[info] Compiling 1 Scala source to /Users/chris/tmp/foo/target/scala-2.11/test-classes...
```

The project was quite implicit-heavy, so my first thought was that the compiler was chasing a diverging implicit or something like that. In actual fact, the compilation had successfully completed and the tests had started running.

### DIFF
--- a/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
@@ -62,6 +62,8 @@ final class MixedAnalyzingCompiler(
       }
     // TODO - Maybe on "Mixed" we should try to compile both Scala + Java.
     if (order == JavaThenScala) { compileJava(); compileScala() } else { compileScala(); compileJava() }
+    if (javaSrcs.size + scalaSrcs.size > 0)
+      log.info("Done compiling.")
   }
 
   private[this] def outputDirectories(output: Output): Seq[File] = output match {


dwijnand (approved):
Sounds good to me.

I think it would be better to put this next to wherever the "Compiling 1 Scala source.." message is.

Also, if @eed3si9n is happy to have this, feel free to backport this to sbt/sbt 0.13 branch.

cb372:
@dwijnand you mean append it to the end of the message that's already been logged?

e.g.

1. Log reads `[info] Compiling 1 Scala source to foo/bar...`
2. Compilation completes
3. The log line now reads `[info] Compiling 1 Scala source to foo/bar... Done.`

I agree that would be ideal, but I couldn't find a way to achieve it using the logger interface that's exposed to zinc.

dwijnand:
That really would be nice. But, no I simply meant in close proximity (code wise) to where that initial log call is made. But now I realise that it's just below (I had expanded the diff up and not found it earlier).

So LGTM! Thanks.

eed3si9n (approved):

PR #206, opened 2016-12-15
Faster zinc

romanowski:
Changes in this PR makes zinc-based compilation comparable intellij-based incremental compilation. 

For single project overhead in compilation (measured as sbt-related phases time in scalc) was reduced from over 10-15% (with bug reported by @wpopielarski even above 30%) to around 5%.

We also reduce significantly overhead in analyzing java compilation (Intellij it is still  faster since it uses faster java compiler).

Overall with some changes to build infra with this changes I get similar performance that Intellij can offer for full build.

More detailed information are placed in commit messages.

Changes are done only to 2.11 sources but can be backported later to 2.10 ones.

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
@@ -4,7 +4,9 @@
 package xsbt
 
 import java.io.File
+import java.util.concurrent.{ TimeUnit, Executors }
 
+import xsbti.AnalysisCallback
 import xsbti.api.DependencyContext
 import DependencyContext._
 
@@ -36,6 +38,26 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with 
   private class DependencyPhase(prev: Phase) extends GlobalPhase(prev) {
     override def description = "Extracts dependency information"
     def name = Dependency.name
+
+    val executor = Executors.newFixedThreadPool(1)
+
+    override def run(): Unit = {
+      super.run()
+      // Wait on all callback calls to finish
+      executor.shutdown()
+      executor.awaitTermination(20L, TimeUnit.MINUTES)
+      ()
+    }
+
+    private def withCallback(op: AnalysisCallback => Unit): Unit = {
+      executor.submit(new Runnable {
+        override def run(): Unit = {
+          op(callback)
+        }
+      })
+      ()
+    }
+
     def apply(unit: CompilationUnit): Unit = {
       if (!unit.isJava) {
         // build dependencies structure
@@ -83,12 +105,11 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with 
         def processDependency(context: DependencyContext)(dep: ClassDependency): Unit = {
           val fromClassName = className(dep.from)
           def binaryDependency(file: File, onBinaryClassName: String) =
-            callback.binaryDependency(file, onBinaryClassName, fromClassName, sourceFile, context)
+            withCallback(_.binaryDependency(file, onBinaryClassName, fromClassName, sourceFile, context))
           val onSource = dep.to.sourceFile
           if (onSource == null) {
             classFile(dep.to) match {
-              case Some((f, binaryClassName, inOutDir)) =>
-                if (inOutDir && dep.to.isJavaDefined) registerTopLevelSym(dep.to)
+              case Some((f, binaryClassName)) =>
                 f match {
                   case ze: ZipArchive#Entry =>
                     for (zip <- ze.underlyingSource; zipFile <- Option(zip.file)) binaryDependency(zipFile, binaryClassName)
@@ -99,7 +120,7 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with 
             }
           } else if (onSource.file != sourceFile) {
             val onClassName = className(dep.to)
-            callback.classDependency(onClassName, fromClassName, context)
+            withCallback(_.classDependency(onClassName, fromClassName, context))
           }
         }
       }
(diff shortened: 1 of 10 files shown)


internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -36,6 +38,26 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
   private class DependencyPhase(prev: Phase) extends GlobalPhase(prev) {
     override def description = "Extracts dependency information"
     def name = Dependency.name
+
+    val executor = Executors.newFixedThreadPool(1)
```
romanowski: This is the simplest possible parallelization - I am open to any more sophisticated (or nicer) ones.
The only problem is that we are in the middle of scala compiler and we've got reduced configuration possibilities.

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -83,12 +105,11 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
         def processDependency(context: DependencyContext)(dep: ClassDependency): Unit = {
           val fromClassName = className(dep.from)
           def binaryDependency(file: File, onBinaryClassName: String) =
-            callback.binaryDependency(file, onBinaryClassName, fromClassName, sourceFile, context)
+            withCallback(_.binaryDependency(file, onBinaryClassName, fromClassName, sourceFile, context))
```
smarter: I would much rather see the thread-handling done inside sbt itself, so scalac synchronously calls `binaryDependency` but the implementation of `binaryDependency` itself will deal with threads

smarter:
> For single project overhead in compilation (measured as sbt-related phases time in scalc) was reduced from over 10-15% (with bug reported by @wpopielarski even above 30%) to around 5%.

Do you have a project somewhere that we can use as a benchmark to test this? Otherwise it's pretty hard to understand the impact of these changes.

romanowski:
@smarter I wish but I cannot share one I tested this on. If you know one I will be more then happy to test.

smarter:
What about compiling http://github.com/scala/scala ?

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
romanowski: It is not that simple since AnalysisCallback is managed by zinc internally. 
@eed3si9n What in your opinion will best way to configure thread-handling there?

romanowski:
Scala still uses sbt 0.13.13 https://github.com/scala/scala/blob/2.12.x/project/build.properties

eed3si9n:
Overall, are these tune-ups focused on making IntelliJ Scala plugin's usage faster?
If so how are the compilation invoked? Are the projects built one by one, or do they run in parallel?
I want to understand some of the tradeoff we are making. For example, if we're caching things, do they increase the memory footprint? If so by how much?

eed3si9n:
As per benchmarking, @cunei was working on building sbt 1 beta from source, and building projects with it. We should definitely try to test how this patch behaves with sbt's task parallelism.

romanowski:
I wasn't clear. I am comparing full compilation using zinc (we've got our own integration) with full compilation in Intellij (using their incremental compiler).

Overhead (mentioned %) was measured as 100% * (time_spent_in_sbt_phases) / (whole time in scalac)

romanowski:
Finally I've got some numbers.

version | API Phase | Dependency Phase | Compilation Time
--- | --- | --- | ---
1.0 | 1.118 | 0.44 | 11.65
this PR | 0.742 | 0.145 | 11.467

Tested on https://github.com/romanowski/shapeless/tree/profiling/problem 

Times exctracted using: https://github.com/romanowski/zinc/commit/afd5c04f74f0f9ade68674feb02f35e37fc1b98d

romanowski:
Rebased and async calls removed. Please let me know if anything more is required.

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
```diff
@@ -47,4 +47,10 @@
      *       Do not depend on it, please.
      */
     boolean nameHashing();
+
+    /** Called at the end of dependency phase. Can be used e.g. to wait on asynchronous tasks. */
```
smarter, line 51: Do you have a concrete usecase for these callbacks?

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
```diff
@@ -27,13 +27,15 @@ private[sbt] object Analyze {
       try { Some(Class.forName(tpe, false, loader)) }
       catch { case e: Throwable => errMsg.foreach(msg => log.warn(msg + " : " + e.toString)); None }
 
-    val productToClassName = new mutable.HashMap[File, String]
+    val productToClassName = mutable.Set.empty[String]
```
eed3si9n: s/productToClassName/classNames/?

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -4,7 +4,9 @@
 package xsbt
 
 import java.io.File
+import java.util.concurrent.{ TimeUnit, Executors }
```
eed3si9n: We no longer need this import correct?

eed3si9n (changes requested):

eed3si9n:
Thanks for benchmarking using OSS project. I'm also happy to see threading removed.

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
romanowski: Yes. I want to implement async calls in my client (since IIRC classDependency and binaryDependency introduce some overhead but I was able to run in parallel to analyzing compilation units).

I fear I will nee to remove it completely or introduce another callback (since I don't see where I can override it in client). @eed3si9n Any pointers?

smarter: Is your client open source?

romanowski: Not yet.

romanowski:
Small change - now I filter out NoSymbol in some situations.

eed3si9n (approved):

Issue #84, opened 2016-03-17
Scripted doesn't support multiprojects


gkossakowski:
This doesn't let us test multi-project support in incremental compiler.

Maybe the solution would be to port such tests to be a plain code that calls into IncHandler directly?

eed3si9n:
In the long run maybe we should run some integration test that runs sbt with freshly compiled zinc:
1. so we can check that zinc didn't break API points for sbt.
2. and be able to run more sophisticated scripted tests

gkossakowski:
This is a pretty good idea on its own but I'd like something more lightweight for testing multi-project support that has a fast turnaround time. Those tests should be really fast to run.

Fast turnaround was my original motivation for introducing specs2-based tests as an alternative to scripted tests.

jvican:
Multi project support was already implemented and nobody closed this ticket. An example is `subproject-dependency`. Here's the proof:
  
```
{
  "projects": [
    {
      "name": "use",
      "dependsOn": ["dep"]
    },
    {
      "name": "dep"
    }
  ]
}
```
  
Unless I'm missing something, this is done. I'm closing it, feel free to reopen if I'm wrong or the motivation of this ticket was different than the multi-project support I refer to.

eed3si9n:
Yes. I fixed in https://github.com/sbt/zinc/pull/208, but didn't close it.

PR #208, opened 2016-12-22
Emulate multi-project in the scripted test

eed3si9n:
Fixes https://github.com/sbt/zinc/issues/84#


internal/zinc-scripted/src/main/contraband/scripted.contra
```diff
@@ -0,0 +1,14 @@
+package sbt.internal.inctest
```
dwijnand, line 1: If this is just to test zinc (properly, with multi-projects), can we please avoid using contraband for this?

dwijnand (changes requested):

internal/zinc-scripted/src/main/contraband/scripted.contra
eed3si9n: I just need some way of describing a build-like structure like https://github.com/sbt/zinc/blob/7185554f6a17839cc2f03d6b853ebca36c99a05b/zinc/src/sbt-test/source-dependencies/binary/build.json

dwijnand: Ok so you need to deserialise some JSON. Don't need contraband or .contra for that.

dwijnand (approved):

Issue #207, opened 2016-12-21
Zinc 1.0.0-X4 regressed on library dependency detection


eed3si9n:
Ref https://github.com/sbt/sbt/pull/2875

### steps

Suppose we have a JAR file named `dep.jar` that defines:

```scala
object A {
  val x = 3
}
```

and a subproject named "uses", which uses `dep.jar`:

```scala
object B {
  val y = A.x
}
```

Next, replace the `dep.jar` so that it no longer defines `object A`, and compile again.

### problem

The compilation succeeds.

### expectation

The compilation fails.

PR #209, opened 2016-12-23
Fix library dependency modification detection

eed3si9n:
This PR is on top of https://github.com/sbt/zinc/pull/208 to reproduce and fix the regression caught by "binary" scripted test.

Fixes sbt/zinc#207

I broke the logic in sbt/zinc#180, which relies on classpath change detection. This reproduces and fixes the library modification detection by invalidating the JAR.
Previously, I was only using the file names to see if the classpath has changed, now it will use the SHA-1 of the content.

### DIFF
--- a/internal/compiler-interface/src/main/contraband/incremental.json
+++ b/internal/compiler-interface/src/main/contraband/incremental.json
@@ -203,6 +203,16 @@
       ]
     },
     {
+      "name": "FileHash",
+      "namespace": "xsbti.compile",
+      "target": "Java",
+      "type": "record",
+      "fields": [
+        { "name": "file", "type": "java.io.File" },
+        { "name": "hash", "type": "int"    }
+      ]
+    },
+    {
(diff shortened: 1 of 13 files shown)


eed3si9n:
Rebased on 1.0

dwijnand:
Just for my own notes, this fixes the regression introduced specifically by https://github.com/sbt/zinc/pull/180/commits/d36c83871fe27f933adea4f122d4d7d1a8a339a7.

dwijnand (approved):
Nice. Sorry for the late review.

dwijnand:
I think this might be the cause of breaking compiler-project/error-in-invalidated.

See https://github.com/sbt/sbt/issues/958 for how this had been fixed.

dwijnand:
Nvm, sorry. Looks like it's simply sbt/sbt#2565 hasn't been forward-ported.

PR #211, opened 2017-01-07
Cross build to Scala 2.12

eed3si9n:


build.sbt
```diff
@@ -8,40 +8,35 @@ import com.typesafe.tools.mima.core._, ProblemFilters._
...
   publishArtifact in Test := false,
   commands += publishBridgesAndTest
 )
 
+def relaxNon212: Seq[Setting[_]] = Seq(
+    scalacOptions := {
+      val old = scalacOptions.value
+      scalaBinaryVersion.value match {
+        case "2.12" => old
+        case _      => old filterNot (Set("-Xfatal-warnings", "-deprecation", "-Ywarn-unused", "-Ywarn-unused-import").apply _)
```
xuwei-k: `.apply _` is unnecessary

build.sbt
```diff
@@ -113,9 +113,11 @@ lazy val zincRoot: Project = (project in file(".")).
         uncommittedChanges.exists(_.nonEmpty)
       },
 
-      // Ignore "git.gitCurrentTags.value.isEmpty"
-      // i.e. allow not-on-tag commits to be considered isSnapshot := false
-      isSnapshot := git.gitUncommittedChanges.value,
```
eed3si9n, line 123: @dwijnand Given that scripted requires publish local, I would like to avoid commit SHA version when it's not tagged.

dwijnand: That's fine. Ideally we should find a way to get scripted to define precise, stable versions instead of using -SNAPSHOT. I changed this to make sbt/sbt scripted tests much faster.

dwijnand (approved):

PR #213, opened 2017-01-27
Add option to disable incremental compilation

romanowski:
Now it is possible to disable incremental compilation metadata creation in IncOptions and directly in AnalysisCallback.

With this flag on e.g. Intellij scala plugin can migrate to new zinc (since it also uses zinc to in it's own incremental compiler but with zinc incremental compilation off)

### DIFF
--- a/internal/compiler-bridge/src-2.10/main/scala/xsbt/CompilerInterface.scala
+++ b/internal/compiler-bridge/src-2.10/main/scala/xsbt/CompilerInterface.scala
@@ -218,8 +218,10 @@ private final class CachedCompiler0(args: Array[String], output: Output, initial
     override lazy val phaseDescriptors =
       {
         phasesSet += sbtAnalyzer
-        phasesSet += sbtDependency
-        phasesSet += apiExtractor
+        if (callback.enabled()) {
+          phasesSet += sbtDependency
+          phasesSet += apiExtractor
+        }
         superComputePhaseDescriptors
       }
     // Required because computePhaseDescriptors is private in 2.8 (changed to protected sometime later).
--- a/build.sbt
+++ b/build.sbt
@@ -311,7 +311,7 @@ lazy val zincScripted = (project in internalPath / "zinc-scripted").
   configure(addSbtUtilScripted)
 
 lazy val publishBridgesAndTest = Command.args("publishBridgesAndTest", "<version>") { (state, args) =>
-  require(args.nonEmpty)
+  require(args.nonEmpty, "Missing arguments to publishBridgesAndTest. Maybe quotes are missing around command?")
   val version = args mkString ""
     s"${compilerInterface.id}/publishLocal" ::
     // using plz here causes: java.lang.OutOfMemoryError: GC overhead limit exceeded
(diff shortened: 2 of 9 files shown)


internal/compiler-bridge/src-2.10/main/scala/xsbt/CompilerInterface.scala
```diff
@@ -218,8 +218,10 @@ private final class CachedCompiler0(args: Array[String], output: Output, initial
     override lazy val phaseDescriptors =
       {
         phasesSet += sbtAnalyzer
```
dwijnand, line 220: You say about `AnalysisCallback#enabled` "If returns false, no sbt phases will be introduced in compiler". Why is `sbtAnalyzer` still added?

romanowski: '''sbtAnalyzer''' only reports generated classes and without it end users will needs to find them manually (e.g. scanning output). 

Main (and only TBH) reason to disable incremental compilation is not to pay overhead (~5% of compilation time after my recent changes) and '''sbtAnalyzer''' almost not affects compilation times.

I will change wording of scala doc.

dwijnand: Thank you. I don't know zinc very well. I wasn't criticising - just learning :)

romanowski: It was good point (the problem was that I created this doc before I realized that I need xsbt-analyzer)

jvican:
@romanowski A small question: why is Intellij using Zinc and disabling incremental compilation? Concrete use case?

romanowski:
Yes. Intellij is using zinc as a scalac wrapper in its own incremental compiler (and then there is no point for using making compilation incremental). 

IIRC Intellij uses custom build of sbt (based on 0.13.5) and has this change included.
https://github.com/JetBrains/intellij-scala/blob/idea163.x/project/dependencies.scala#L44-L48 

@jastice @niktrop @mutcianm can you shade some light here?

niktrop:
Currently we use a patched version of Zinc from sbt 0.13.9. It allows us to use different versions of scala compiler in the same way and provides an API for getting feedback from it.

https://github.com/niktrop/sbt/commits/0.13.9.patched

There are two modes of incremental compilation supported in IDEA. One just uses Zinc, second is adapted from IDEA incremental compiler for java, based on bytecode analysis. So the main reason for use a patched version is to get all information required for IDEA incremental compiler. Other reason is indeed to skip compiler phases introduced to collect data for Zinc when we do not need it.

jvican:
Interesting @niktrop, thanks for elaborating on it. So another question: is the IDEA incremental compiler using bytecode analysis in Scala class files or only Java class files?

niktrop:
It works for any class files. We just have to provide a correspondence between sources and compiled classes. I think we do some additional checks for scala package objects, but thats all.

jvican:
@niktrop AFAIK, class file analysis won't give you information about a lot of Scala dependencies because some of them are not even present in the generated class files, especially those depending on types. Is this a known limitation in your incremental compiler and you intentionally not cover those cases?

niktrop:
We are aware of it. There is a tradeoff between correctness and completeness of dependency analysis and recompilation time. So IDEA incremental compiler for Scala is optimistic one and it works rather well. It was simple to implement using IDEA platform, but we didn't mean to create a full fledged alternative to Zinc.

jvican:
Yeah, I thought so @niktrop. that makes sense -- Just wanted to confirm this is the case. Thanks for elaborating on it.

eed3si9n (approved):

eed3si9n:
The change looks good to me. Can I merge it?

PR #216, opened 2017-01-31
Make incremental compilation artifacts reusable (aka cached compilation)

romanowski:
Goal is achieved by providing custom mappers to TextAnalysisFormat.

Based on client can create CacheProviders that using given mappers can migrate incremental compilation artifacts between workspaces.

This is can be done in two steps:
 * export cache to intermediate format (e.g. map path as a relative to abstract roots)
 * import cache from intermediate format (e.g. replace abstract roots with concrete values).

Zinc is not responsible for maintaining caches (e.g. there is no verification if correct cache provider is used) or searching caches for build, this is client responsibility (zinc only provides utilities).
 
This PR also include two basic cache implementation based on making all paths relative to project root. They are used to test whole mechanism internally and are base for any custom caches.
* project rebased cache (copy artifacts and classes from different project)
* exportable cache (first export cache as pair or zipped classes and incremental compilation metadata) that can be later imported to another workspace

Internally, we are just about to beta tests this with huge scala codebase (way over 100K lines of scala code). Results so far are promising:
 * currently we can observe ~ 4-5 times improvement on full build (when relatively close to cache)
 * most of time is spent on extracting zips with classfiles and with few fixes we should get speedups around x10 times.
Of course workspace after cache import behaves as it was compiled locally.

Our integration is not open source but in future (hopefully close) we will release at least part of it to public domain.

I am working on open source plugins for sbt (both 0.13.x and 1.0.x versions) based on that PR (deep WIP so far): https://github.com/romanowski/hoarder 


.gitignore
```diff
@@ -1 +1,3 @@
 target/
+.idea
```
jvican: Are these additions intended? IMO, these ignores should be local. See how to implement them globally in your computer in https://help.github.com/articles/ignoring-files/.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalNameHashing.scala
```diff
@@ -32,8 +32,10 @@ private final class IncrementalNameHashing(log: sbt.util.Logger, options: IncOpt
       val aNameHashes = a.nameHashes
       val bNameHashes = b.nameHashes
       val modifiedNames = ModifiedNames.compareTwoNameHashes(aNameHashes, bNameHashes)
-      val apiChange = NamesChange(className, modifiedNames)
-      Some(apiChange)
+      if (modifiedNames.regularNames.nonEmpty || modifiedNames.implicitNames.nonEmpty) {
```
jvican: I think I'm missing context here. Why is this check required?

internal/zinc-core/src/test/scala/sbt/inc/TestAnalysisCallback.scala
```diff
@@ -58,40 +58,7 @@ class TestAnalysisCallback(
   def hashFile(f: File): Array[Byte] = Stamp.hash(f).asInstanceOf[Hash].value
 
   def get: TestAnalysis = {
-
```
jvican, line 60: Why is all of this gone? Are we not interested in testing this?

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -162,13 +162,13 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
         case _ => newOne()
       }
     }
-    private def addClassDependency(deps: HashSet[ClassDependency], fromClass: Symbol, dep: Symbol): Unit = {
+    private def addClassDependency(deps: HashSet[ClassDependency], fromClass: Symbol, dep: Symbol): Unit = if (dep != NoSymbol) {
```
jvican, line 163: We should also guard against `null`, just to be safer since scalac can unexpectedly give us a null from time to time. Perhaps it's a good idea to create a utility called `ignoreSymbol` that will guard against `NoSymbol` and `null` at the same time? It's a very recurrent logic in Zinc.

internal/zinc-persist/src/main/scala/sbt/internal/inc/RelationsTextFormat.scala
```diff
@@ -0,0 +1,128 @@
...
+    header: String,
+    selectCorresponding: Relations => Relation[A, B],
+    keyMapper: Mapper[A],
+    valueMapper: Mapper[B]
+  )
+
+  private def stringsDescriptor(header: String, rels: Relations => Relation[String, String]) =
+    Descriptor(header, rels, Mapper.forString, Mapper.forString)
+
+  private val allRelations: List[Descriptor[_, _]] = {
```
jvican, line 29: This should probably be well documented and linked with the original source `Relations` since any change there may affect behaviour here. I would do the same for the rest of the methods in this class. In fact, it's a little bit unfortunate that this is generated manually and not automatically.

jvican (approved):
Hello @romanowski, great job!
  
I've left some minor comments/questions. Hope they are useful. Otherwise, LGTM.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalNameHashing.scala
romanowski: I've encountered situations where for some reason class got different API with no names changed (on 'normal' incremental compilation - not 'cached' ones).

Debugging problems with API hashes is really hard so I gave up looking why class has changed API without any names change and just introduce this check.

eed3si9n:
I'm out in Canada this week for Lightbend meetings, but I'll take closer look at this when I get back.

internal/zinc-core/src/test/scala/sbt/inc/TestAnalysisCallback.scala
romanowski: Tests for that utils are commented out: https://github.com/sbt/zinc/blob/1.0/internal/zinc-core/src/test/scala/sbt/inc/IncrementalTest.scala

internal/zinc-persist/src/main/scala/sbt/internal/inc/RelationsTextFormat.scala
romanowski: Agree on documentation part but IMO first we need to clean up whole logic around relation construction. Code (from below):

```scala
relations match {
      case p :: bin :: lcn :: mri :: mre :: ii :: ie :: lii :: lie :: cn :: un :: bcn :: Nil =>
```

Is IMO one of the most hacky thing in whole zinc codebase and at some point it needs to be cleanup and refactored.
I tried to do so in this PR but after some time I look at diff and I was way to big for single PR (and I needed to fit cached compilation related changes).

.gitignore
romanowski: This is an option but most project I work with has `.idea` added directly in in local gitignore (in this dir intellij stores whole project config). 

I can remove `.idea` if needed.

`zinc/src/test/resources/bin` stores generated jars/classesdirs

dwijnand: `zinc/src/test/resources/bin` is fine and correct to ignore, imo. `.idea` not.

jvican: If `zinc/src/test/resources/bin` is added, then it's because it has generated jars/classesdirs. But I don't see any of those jars in git history. If they are present, they should also be removed. Otherwise, I don't see why we should special case this directory, it hasn't generated anything for me at least.

romanowski: Ok, I will remove `.idea` then

internal/zinc-persist/src/main/scala/sbt/internal/inc/RelationsTextFormat.scala
jvican: I agree, and also it's better to keep PRs short, otherwise it's very difficult to reason about the changes and we should be a little bit conservative and test everything carefully before a merge :smile:.
  
I'll take care of cleaning up the relations logic and maybe then we can document this code.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalNameHashing.scala
jvican: But this check is very weird. You're checking against something that `NamesChange` was already checking. This is how `NamesChange` is currently defined:
```scala
final case class NamesChange(modified0: String, modifiedNames: ModifiedNames) extends APIChange(modified0) {
  assert(
    modifiedNames.regularNames.nonEmpty || modifiedNames.implicitNames.nonEmpty,
    s"Modified names for $modified0 is empty"
  )
}
```
  
What you're doing here is no different than what we were doing before. We never return `None` for the same reason that we never got an `AssertionError` before.

romanowski: Good catch - I recall removing that assertion probably it was lost in rebasing process.

I will remove that new check and if I face another problem with it I will create new PR.

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
romanowski: This is not a new logic - I just noticed that we keep empty name for multiple classes so I just remove it. I never see any nulls here (call sites of `addClassDependency` taking care of it)

jvican: Perfect, even better. But I think we should guard against all of this in the whole Zinc. I'll do that in a PR.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalNameHashing.scala
jvican: Great @romanowski!

jvican:
Have you tested these changes thoroughly @romanowski ? I wonder because the diffs are a little bit scary, and even though you've added several tests I'd like to make sure that real-world behaviour is stable.

romanowski:
Do you know any repository that uses sbt 1.0? I've created mine based on shapless (https://github.com/romanowski/shapeless). 

I tested this a lot with my custom (not open source yet) integration and so far all works fine (including cached compilation based on this PR)

jvican:
No, I do not, but since we'd like to stabilize Zinc for a beta release, we better start thinking about this stuff as soon as possible :smile:. I didn't find the commit that sets up 1.0 in that shapeless fork, maybe you can describe how to try 1.0 out?
  
It's good that this PR has been tested in such an intense codebase as Shapeless. So thanks for that!

romanowski:
I put it wrong: I have shapless repo but I didn't test new zinc there (I tested zinc on non-open source codebase). 
Of course now I will (rather tomorrow).

romanowski:
Here is 1.0 based branch https://github.com/romanowski/shapeless/tree/profiling/problem 
Beware! By default it uses sbt 1.0.0-M4 that has a problem with extracting names (it takes ages for shapless) but it is fixed in 1.0 branch.

romanowski:
I did basic tests with shapeless repo and except #220 that is IMO not related to this PR zinc worked fine.

jvican:
I'm happy to see this merged @romanowski. I agree that #220 is not strictly related to this, though in the same lines. Could you have a look at it? I'll comment on the strategy in the ticket.

romanowski:
I fear I won't be able to look at #220 soon (maybe end of Feb). If I have enough time I would create PR instead of issue :)

romanowski:
Branch is rebased and headers are added for new files.

romanowski:
@eed3si9n is there anything I need to change/explain? Since zinc repo become much more alive now I don't want to rebase this (quite big) PR multiple times :)

eed3si9n:
I'm wondering if we need this change if we implement https://github.com/sbt/zinc/issues/218 instead. The purpose of transformation is to make it machine-independent, correct?

romanowski:
@eed3si9n 

> The purpose of transformation is to make it machine-independent, correct?
This is only one option. Generally with this PR clients can easy modify analysis and make it not only machine independent but also system or even build tool.

#218 can be based on this PR. As commented in #218 I don't think we can make zinc analysis 100% machine independent without full knowledge from build tool. E.g. how you can handle sbt's unmannaged jars or sources? 

Naive implementation of #218 is also part of this PR ( ProjectRebasedCache ).

Is there anyone working on #218? Is there any plan how to implement it?

Moreover general mechanics of zinc generally is not changed (except change Comlilation to compilationTimestamp) and I can remove all optional logic (so this PR will consist only around TextFormat),

eed3si9n:
> Is there anyone working on #218? Is there any plan how to implement it?

I can certainly work on #218. Not 100% sure if it can be done, but my idea was to essentially strip the information until it is machine-independent and have some sort of a tag that can be substituted per machine.

romanowski:
Did you know that even option to scalac can be machine depended? E.g. path to compiler plugin or parameters passed to it. Same goes for almost every entry kind in Analysis so I am interested how do you want to detect and tag all those cases?

I think using my Mappers is a good start for #218. Since you can implement it as specific mapper can merge this PR and stat another one based on my work? Or merge this one and if we can (somehow) implement #218 inside zinc then we can remove all not-used logic?

eed3si9n:
ok. Let's merge this first then.

gabro:
@romanowski thanks for this! Just to understand, will this be available in sbt directly, or should we wait for hoarder to be ready?

romanowski:
@gabro this will be part of sbt 1.0 and AFAIK there is almost no project using it.

Hoarder will use this PR for 1.0 branch and will implement similar solution for 0.13 (I am working on that).

gabro:
Awesome, thanks

stuhood:
In the case of scalac parameters + plugins, etc: those must be part of the
cache key. Therefore, there is no need to strip them from the file. It's
only stuff that has no effect on the outcome of the build that should be
stripped.


On Feb 13, 2017 12:15 PM, "Krzysztof Romanowski" <notifications@github.com>
wrote:

Did you know that even option to scalac can be machine depended? E.g. path
to compiler plugin or parameters passed to it. Same goes for almost every
entry kind in Analysis so I am interested how do you want to detect and tag
all those cases?

I think using my Mappers is a good start for #218
<https://github.com/sbt/zinc/issues/218>. Since you can implement it as
specific mapper can merge this PR and stat another one based on my work? Or
merge this one and if we can (somehow) implement #218
<https://github.com/sbt/zinc/issues/218> inside zinc then we can remove all
not-used logic?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/sbt/zinc/pull/216#issuecomment-279508961>, or mute the
thread

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -160,10 +160,10 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
...
       assert(fromClass.isClass, Feedback.expectedClassSymbol(fromClass))
       val depClass = enclOrModuleClass(dep)
-      if (fromClass.associatedFile != depClass.associatedFile) {
+      if (fromClass.associatedFile != depClass.associatedFile && !depClass.isRefinementClass) {
```
jvican, line 166: @romanowski Why are you doing this? This looks wrong.

romanowski: @jvican Refinement classes do not have byte code representation and it fails later on (we still depends on parents of refinement)

jvican: Right, in that case this is correct. I double checked with SLS.

jvican:
@romanowski Is there something we can do to follow up on @stuhood's remark? This is related https://github.com/pantsbuild/pants/issues/3923.

romanowski:
I mentioned scalac plugins and I don't see how this is related to pant's plugin written in Java (but I don't know pants so I might miss something).

I generally agree with @stuhood's but we've got a case where compiler plugin is a part of same project so it should use relative path in such case. 

Generally as said before this PR open multiple possibilities such as cached compilation or #218

PR #217, opened 2017-02-02
Fix #192: Handle local dependencies for Java classes

### LINKED ISSUE #192: Java Anonymous and local classes are not handled properly
From Class based namehashing (https://github.com/sbt/zinc/pull/86)

jvican:
Aside from fixing #192, the two first commits add more tests and guard against
the occurrence of `NoType` in the traversers.

Zinc incrementally compiles both Java and Scala sources. Most of the time,
Scala sources depend on Java sources and it's important that when Java sources
are modified, Scala sources that depend on those modifications are also
compiled again. Supporting this use case is simple -- Scalac pickles
information from Java class files and turns them into symbols that can be
handled by Zinc.

The situation starts to get trickier when it's turned around. If we want our
Java sources to be recompiled when there are Scala changes, we have to detect
the dependencies at the Java level. Java sources are compiled by javac, so we
have no symbol information whatsoever we can use to detect dependencies. To
solve this problem, Zinc class-loads the generated class files by Javac and
reflectively detects any dependency at the class level. Thanks to the clas file
[constant pool](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.4),
we can also know which classes a given class file refers to -- buying us the
analog of the name hashing algorithm implemented by @gkossakowski in Scala
sources.

Before, local class files were correctly honoured by Zinc and linked to their
original source file (remember that javac creates a new class file for every
class that is not top level). However, local classes were not analyzed by
`ClassAPI` and registered as dependencies.

These commits introduce local dependencies by inheritance from Java classes to
Scala classes. It applies a similar fix to the one provided in https://github.com/gkossakowski/sbt/commit/47416b394e3730d67367da9f011f8ef3de23b576.
In the commit messages, I dive into a more succint explanation of the fixes,
feel free to check them out.

This work is done by the [Scala Center](https://scala.epfl.ch/).

### DIFF
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -16,11 +16,11 @@ object ClassToAPI {
   def apply(c: Seq[Class[_]]): Seq[api.ClassLike] = process(c)._1
 
   // (api, public inherited classes)
-  def process(c: Seq[Class[_]]): (Seq[api.ClassLike], Set[(Class[_], Class[_])]) =
+  def process(classes: Seq[Class[_]]): (Seq[api.ClassLike], Set[(Class[_], Class[_])]) =
     {
-      val pkgs = packages(c).map(p => new api.Package(p))
+      val pkgs = packages(classes).map(p => new api.Package(p))
       val cmap = emptyClassMap
-      val defs = c.flatMap(toDefinitions(cmap))
+      val defs = classes.flatMap(toDefinitions(cmap))
       cmap.lz.foreach(_.get()) // force thunks to ensure all inherited dependencies are recorded
       val classApis = cmap.allNonLocalClasses.toSeq
       val inDeps = cmap.inherited.toSet
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -20,7 +20,7 @@ object ClassToAPI {
     {
       val pkgs = packages(c).map(p => new api.Package(p))
       val cmap = emptyClassMap
-      val defs = c.filter(isTopLevel).flatMap(toDefinitions(cmap))
+      val defs = c.flatMap(toDefinitions(cmap))
       cmap.lz.foreach(_.get()) // force thunks to ensure all inherited dependencies are recorded
       val classApis = cmap.allNonLocalClasses.toSeq
       val inDeps = cmap.inherited.toSet
--- a/internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
+++ b/internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
@@ -69,11 +69,14 @@ private[sbt] object Analyze {
         loadedClasses.partition(_.getCanonicalName != null)
 
       // Map local classes to the sources of their enclosing classes
-      val localClassesToSources = localClassesOrStale.flatMap { cls =>
-        val enclosingLoadedCls = Option(cls.getEnclosingClass)
-        val sourceOfEnclosing = enclosingLoadedCls.map(binaryToSourceName(_))
-        sourceOfEnclosing.flatten.map(src => cls.getName -> src).toList
-      }.toMap
+      val localClassesToSources = {
+        val localToSourcesSeq = for {
+          cls <- localClassesOrStale
+          enclosingCls <- Option(cls.getEnclosingClass)
+          sourceOfEnclosing <- binaryToSourceName(enclosingCls)
+        } yield (cls.getName, sourceOfEnclosing)
+        localToSourcesSeq.toMap
+      }
 
       /* Get the mapped source file from a given class name. */
       def getMappedSource(className: String): Option[String] = {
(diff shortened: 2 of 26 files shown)


jvican:
Is there any reason why CI information is not available? Is the CI running?

lightbend-cla-validator:
Hi @jvican,

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
```diff
@@ -121,21 +119,28 @@ private[sbt] object Analyze {
...
           processDependencies(inheritanceDeps, DependencyByInheritance, className)
       }
 
-      // TODO(jvican): Add local inheritance dependencies
+      // Read API of local classes and process local dependencies by inheritance
+      val localClasses = localClassesOrStale.filter(cls =>
+        localClassesToSources.contains(cls.getName))
+      val localInherited: Map[String, Set[String]] =
+        readInheritanceDependencies(localClasses)
+      nonLocalInherited foreach {
```
Duhemm: Shouldn't it be `localInherited` here?

jvican: Yes, it should, thanks for the catch.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -20,7 +20,7 @@ object ClassToAPI {
     {
       val pkgs = packages(c).map(p => new api.Package(p))
       val cmap = emptyClassMap
-      val defs = c.filter(isTopLevel).flatMap(toDefinitions(cmap))
```
romanowski: can we rename `c` into e.g. `classes`?

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
```diff
@@ -63,22 +63,34 @@ private[sbt] object Analyze {
...
-        readAPI(source, loadedNonLocalClasses).groupBy(_._1).mapValues(_.map(_._2))
+      val loadedClasses = classFiles.map(c => binaryClassNameToLoadedClass(c.className))
+      // Local classes are either local, anonymous or inner Java classes
+      val (nonLocalClasses, localClassesOrStale) =
+        loadedClasses.partition(_.getCanonicalName != null)
+
+      // Map local classes to the sources of their enclosing classes
+      val localClassesToSources = localClassesOrStale.flatMap { cls =>
+        val enclosingLoadedCls = Option(cls.getEnclosingClass)
+        val sourceOfEnclosing = enclosingLoadedCls.map(binaryToSourceName(_))
```
romanowski: Maybe:

```scala
val localClassesToSources: Map[String, String] =  (for {
  cls <- localClassesOrStale
  enclosingLoadedCls <- Option(cls.getEnclosingClass)
  sourceOfEnclosing <- binaryToSourceName(enclosingLoadedCls)
} yield (cls.getName, sourceOfEnclosing))(collection.breakout)
```

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
```diff
@@ -63,22 +63,34 @@ private[sbt] object Analyze {
...
+      }.toMap
+
+      /* Get the mapped source file from a given class name. */
+      def getMappedSource(className: String): Option[String] = {
+        val nonLocalSourceName: Option[String] = for {
+          loadedClass <- binaryClassNameToLoadedClass.get(className)
+          sourceName <- binaryToSourceName(loadedClass)
+        } yield sourceName
+
+        if (nonLocalSourceName.isDefined) nonLocalSourceName
```
romanowski: Maybe: 
```scala
nonLocalSourceName orElse localClassesToSources.get(className)
```

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
```diff
@@ -63,22 +63,34 @@ private[sbt] object Analyze {
...
-          } yield sourceName
-          clsName match {
-            case Some(resolvedClassName) => resolvedClassName
-            // TODO: handle dependencies coming from local classes the same way as it's done for Scala
-            // classes, see: https://github.com/sbt/sbt/issues/1104#issuecomment-169146039
-            case None                    => return
+        val fromClassName: String = {
+          getMappedSource(fromBinaryName) match {
+            case Some(sourceName) => sourceName
+            case None             => return // It could be a stale class file
```
romanowski: TBH I don't like returns in scala code (but I don't want force my way here :) ). Maybe instead:
```scala
def processDependency(onBinaryName: String, context: DependencyContext, fromBinaryName: String): Unit = 
 getMappedSource(fromBinaryName) match {
  case None => 
  case Some(fromClassName) =>
    // Old code here
}
```

romanowski:
LGTM with few small code style suggestions

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
jvican: The return was already there, I didn't introduce it. I didn't want to create major diff change and tried to keep it minimal. I can rewrite it if you like.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: Sure.

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
jvican: 👍

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
jvican: 👍

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
romanowski: It's your call, I understand your point here.

eed3si9n (approved):

eed3si9n:
LGTM
Thanks @jvican and Scala Center for the contribution.

jvican:
I will open another PR (perhaps a batch of other PRs) to backport all my changes to 2.10 sources. I've just realized my changes only affected 2.11 sources.

romanowski:
I think we can wait a while and backport changes in one PR (e.g. my speedup was also targeted to 2.11/2.12 only).

jvican:
Yes, that's what I had in mind -- it's why I said batch of other PRs.

PR #219, opened 2017-02-04
Mark compactify as a passing test

jvican:
Compactify was never marked as a passing test after we got a regression
in https://github.com/sbt/sbt/issues/1553.

For some reason, the scripted setup was not recognizing tasks defined
locally in the `build.sbt`, and so we add this functionality to
`IncHandler` which is the responsible of mapping task keys to their
actual implementation in sbt-scripted.

Aside from moving this functionality around and marking the test as
passing, I took the liberty to modify the name of `output-empty` for
something clearer in spirit: `checkNoClassFiles`.


eed3si9n:
> For some reason, the scripted setup was not recognizing tasks defined locally in the build.sbt

The scripted handler for Zinc does not support sbt, so I've marked a bunch of stuff pending when we started porting scripted tests over.

eed3si9n (approved):

jvican:
Thanks for confirming Eugene. It would be nice if in the future there's a way to still recognize concrete SBT settings like log levels.

PR #221, opened 2017-02-07
Fix #113 and apply several other fixes

### LINKED ISSUE #113: Move SafeLazy to the compiler-interface project?
I can't figure out why `SafeLazy` is in a separate project (`zinc-apiinfo` in sbt 1.0, `api` in sbt 0.13) and needs to be loaded using reflection from the compiler bridge which assumes that it is present on the classpath: https://github.com/sbt/zinc/blob/088a577149b0803b19dcec4f984d4035c5e4fc7b/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala#L139-L144

It seems like it would be much simpler to have `SafeLazy` in `compiler-interface` so that no reflexive call is needed. I don't think that this would affect what classes and classloaders can be GCed, but maybe I'm missing something.

jvican:
This PR fixes [one issue](https://github.com/sbt/zinc/issues/113), cleans up a little bit of Zinc code, adds more helpers
and makes some performance fixes to the code.
  
Please, for full information on the issues fixed, have a look at the commit
messages.


romanowski:
LGTM

Did you run some benchmarks? I am really interested how much we gain from dropping reflection.

smarter:
It probably doesn't make a difference because the JVM should be able to do de-reflection optimizations since the method called via reflection is always the same: http://stackoverflow.com/a/28809546/348497

jvican:
@romanowski Correct @smarter, I explain it here in [this commit message](https://github.com/jvican/zinc/commit/c1f821b1b37704618d54378c25a2464c8ad48a00).

eed3si9n (approved):

PR #228, opened 2017-02-19
Fix #152: Remove all zinc core tests infrastructure

### LINKED ISSUE #152: sbt.inc.IncrementalTest disabled

jvican:
As pointed out by Eugene in https://github.com/sbt/zinc/pull/153,
the incremental test infrastructure that was present in the past was
forgotten in favor of sbt-test tests.

This commit removes all the tests, including the scalacheck generators.
In case we bring scalacheck tests to life down the line, we can always
come back to this commit and revive `TestCaseGenerators`.

P.S. I'm sad to see these tests leaving since the API looks pretty nice.


jvican:
Review by @eed3si9n.

eed3si9n (approved):

dwijnand:
@jvican Sorry, quick question: is there nothing in zinc's build that can be simplified/removed given the loss of this infrastructure?

jvican:
@dwijnand Not that I know of, since this was only touching test sources. I think there's nothing else to remove.

PR #225, opened 2017-02-08
Add JMH benchmarks and improve performance of Zinc phases

jvican:
This pull request applies some minor fixes and improves the performance of the
`Dependency` phase by merging the two-step implementation into one step. For an
explanation of the applied changes and its effect to performance, please check
the commit messages.

This work is done by the [Scala Center](https://scala.epfl.ch).

PR #232, opened 2017-02-21
Backport changes from several previous PRs

jvican:
This commit introduces changes to Scala 2.10 sources from the following
PRs:

* https://github.com/sbt/zinc/pull/225
* https://github.com/sbt/zinc/pull/216
* https://github.com/sbt/zinc/pull/206
* https://github.com/sbt/zinc/pull/221

It also removes a stub for 2.8 compatibility in `DelegatingReporter`.
Support for Scala 2.8 compatibility is not already maintained it.

### DIFF
--- a/internal/compiler-bridge/src-2.10/main/scala/xsbt/API.scala
+++ b/internal/compiler-bridge/src-2.10/main/scala/xsbt/API.scala
@@ -1,6 +1,7 @@
 /* sbt -- Simple Build Tool
  * Copyright 2008, 2009, 2010, 2011 Mark Harrah
  */
+
 package xsbt
 
 import scala.tools.nsc.Phase
@@ -11,7 +12,7 @@ object API {
   val name = "xsbt-api"
 }
 
-final class API(val global: CallbackGlobal) extends Compat {
+final class API(val global: CallbackGlobal) extends Compat with GlobalHelpers {
   import global._
 
   def newPhase(prev: Phase) = new ApiPhase(prev)
@@ -39,12 +40,12 @@ final class API(val global: CallbackGlobal) extends Compat {
       if (global.callback.nameHashing) {
         val extractUsedNames = new ExtractUsedNames[global.type](global)
         val allUsedNames = extractUsedNames.extract(unit)
-        def showUsedNames(className: String, names: Set[String]): String =
+        def showUsedNames(className: String, names: Iterable[String]): String =
           s"$className:\n\t${names.mkString(", ")}"
         debuglog("The " + sourceFile + " contains the following used names:\n" +
           allUsedNames.map((showUsedNames _).tupled).mkString("\n"))
         allUsedNames foreach {
-          case (className: String, names: Set[String]) =>
+          case (className: String, names: Iterable[String]) =>
             names foreach { (name: String) => callback.usedName(className, name) }
         }
       }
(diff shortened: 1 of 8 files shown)


jvican:
I'll be basing my next PR to fix ASF in these changes because they bring Scala 2.10.x and >= 2.11.x up to date. Changes to the type traverser require changes in the compiler bridge test suite, that is shared by both versions.
  
Can we merge this soon and I'll push my changes to fix ASF?

dwijnand:
@jvican Correct me if I'm wrong, but isn't the difference in source code between 2.10 and non-2.10 minimal? Would it not be possible to reduce the difference in source code? Instead of this business of having to "backport" changes in the same repo on the same branch.

Also, why is this even necessary post-merge? Don't we test zinc under 2.10 in CI? Why didn't it fail Travis CI?

jvican:
@dwijnand Indeed, differences are minimal and very focused, they could be abstracted away. But this requires time. I won't be able to do such thing short-term. Also, it's not clear it's possible.
  
> Also, why is this even necessary post-merge? Don't we test zinc under 2.10 in CI? Why didn't it fail Travis CI?
  
My changes to `ExtractUsedNames` and `Dependency` haven't been pushed yet, they are local, so you won't notice they fail. These changes affect correctness, and that's why keeping an old, unmodified version of Zinc in 2.10 will make some tests fail -- the testing suite is shared.

dwijnand:
No, what I mean is take #216 for example. It changes just the non-2.10 version of compiler-bridge (https://github.com/sbt/zinc/pull/216/files#diff-078aa14c94ffc4285ff1c49d727f9a67).

So either:
1. We're not running tests against 2.10 zinc, or
2. There's no test for the changes made in the PR

If it's (2) it isn't great, but if it's (1) it's concerning.

Personally I would prefer if we minimised the changes instead of backporting them. I'll try and find time to work on that, @jvican, I don't mean to burden you.

dwijnand:
#233

jvican:
The reason for which this PR fails demonstrates that we're indeed testing the 2.10 sources. The PRs that you mention don't affect correctness of the compiler, that is, the underlying logic -- they just change the code to make it more performant.
  
> Personally I would prefer if we minimised the changes instead of backporting them.
  
I would also. But since this is a blocker for me to continue my work on Zinc, which will finish in one week, I propose that we merge these changes. We can minimise the differences later down the road.

jvican:
I've just updated this PR to see if now there's more luck with CI. Previous failing test here: https://travis-ci.org/sbt/zinc/jobs/203844727.

dwijnand:
> Refinement classes do not have byte code representation and it fails later on (we still depends on parents of refinement)

(from https://github.com/sbt/zinc/pull/216/files#r99321507)

would imply correctness not performance.

jvican:
I wonder what the root of @romanowski's problem was, because yeah, that change was provided without any test. Hopefully it's just an exception. I think it's safe because we're still traversing the declarations and the parents of the `RefinedType`.

dwijnand:
> But since this is a blocker for me to continue my work on Zinc, which will finish in one week..

I see. I'll work on it tomorrow then.

romanowski:
@jvican I know, I should write a test case for this when I spot a problem. Sorry for that.

I can try to reproduce it if you really like but it may take me a while (and I am quite busy this week).

jvican:
@romanowski Don't worry, I don't think it's important :wink:.

dwijnand:
@jvican I've started making progress, but 2 things: (1) it's not ready yet, and (2) it might involve changing some paths, maybe (I'm not done).

I wouldn't want to further delay your work _as well as_ mess you up by moving files around.

So let's merge.

dwijnand (approved):

jvican:
Thanks, if you're still stuck on this after one week, I can help you out. The changes are not trivial and they may affect performance. You'll probably need to do some tricks with implicit resolution for it to work correctly.

PR #236, opened 2017-02-24
Improve binary API and documentation

jvican:
This PR focuses primarily on improving API documentation.
It also does some modifications to code that deals with intricate logic to make 
it easier to read for both users of the API and maintainers.

As always, I suggest having a look at the commit messages to get some hints on
the concrete tasks carried out in every commit. My plan is to open a new PR
soon targeting structural changes to the API, to make it usable from both Java
and Scala.

In short, this PR:
* Documents better the compiler interface, the compile core project and the
  zinc project.
* Starts to define the interface of `IncrementalCompiler`.
* Removes unused imports, and makes hard-to-grasp code more readable.

Review by @dwijnand, @romanowski, @eed3si9n.


internal/compiler-interface/src/main/java/xsbti/compile/IncrementalCompiler.java
```diff
@@ -37,32 +38,69 @@
...
+                          Output output,
+                          GlobalsCache globalsCache,
+                          String[] scalacOptions,
+                          String[] javacOptions,
+                          Maybe<CompileAnalysis> previousAnalysis,
+                          Maybe<MiniSetup> previousSetup,
+                          PerClasspathEntryLookup perClasspathEntryLookup,
+                          Reporter reporter,
+                          CompileOrder compileOrder,
+                          // Has to be boxed to override in Scala,
```
jvican, line 99: I have not yet been able to reproduce this and file a bug report.

internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala
```diff
@@ -10,121 +10,264 @@ package internal
...
+  /**
+   * Distinguish Dotty and Scala version given the version number.
+   * FIXME: This implementation assumes that dotty will be `0.x` for some time.
+   */
+  def isDotty(version: String): Boolean = version.startsWith("0.")
 
-  def apply(org: String, version: String, launcher: xsbti.Launcher): ScalaInstance =
-    // Due to incompatibility with previous launchers if scalaOrg has default value revert to an existing method
+  /** Create a [[ScalaInstance]] from a given org, version and launcher. */
+  def apply(
```
jvican: Removing this in next commit, unnecessary.

dwijnand:
> +2,421 −798

Thankfully lots of it is just text, but, still, that's a big PR :)

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
```diff
@@ -11,54 +11,171 @@
...
+     * @param context The kind of dependency established between
+     *                <code>onClassName</code> and <code>sourceClassName</code>.
+     *
+     * @see xsbti.api.DependencyContext
+     */
+    void classDependency(String onClassName,
+                         String sourceClassName,
+                         DependencyContext context);
+
+    /**
```
eed3si9n, line 40: Maybe it's slightly offtopic for this PR, but we should avoid the term "binary" in Zinc because it's used in several different meanings. I think external JARs should be called a "library", and therefore this method should be renamed to `libraryDependency`. I've already started the transition here - https://github.com/sbt/zinc/commit/38ad637d2b91567877d319cace2c396869bb194d

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
```diff
@@ -11,54 +11,171 @@
...
+     * Indicate that the class <code>fromClassName</code> depends on a class
+     * named <code>onBinaryClassName</code> coming from class or jar file
+     * <code>binary</code>.
+     *
+     * @param onBinary Binary name being depended on.
```
eed3si9n: Isn't onBinary just a library file here?

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
```diff
@@ -11,54 +11,171 @@
...
-     *       Do not depend on it, please.
+     * @see xsbti.api.DependencyContext for more information on the sort of
+     * dependency that can be registered.
+     *
+     * The implementation of this method is meant to be free of side-effects.
+     * It's added to {@link AnalysisCallback} because it indicates how other
+     * callback calls should be interpreted by its implementations and clients.
+     *
+     * This method is an implementation detail and can be removed at
+     * any point without any deprecation. Do not depend on it.
```
eed3si9n: (Also potentially off topic on this PR) Perhaps we should just remove it now. This is a boolean flag indicating it uses source-based name hashing as opposed to the older incremental compiler. Now we no longer have either of the algorithms, and we have new class-based name hashing.

internal/compiler-interface/src/main/java/xsbti/compile/IncrementalCompiler.java
```diff
@@ -37,32 +38,73 @@
...
+     */
+    CompileResult compile(ScalaCompiler scalaCompiler,
+                          JavaCompiler javaCompiler,
+                          File[] sources,
+                          File[] classpath,
+                          Output output,
+                          GlobalsCache globalsCache,
+                          String[] scalacOptions,
+                          String[] javacOptions,
+                          Maybe<CompileAnalysis> previousAnalysis,
```
eed3si9n, line 94: We should phase Maybes out and use java.util.Optional.

eed3si9n (changes requested):

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
jvican: You're right, let's do it.

internal/compiler-interface/src/main/java/xsbti/compile/IncrementalCompiler.java
jvican: I think this is a bad idea, because `java.util.Optional` was introduced in JDK8. If we use it, we force all our users to run the compiiler bridge in JDK8. I also think it's better to be consistent with the sbt interface and keep it.

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
jvican: I think that `library` gives a completely different meaning. Also, a binary name doesn't come from a library necessarily, it can come from a user project that has already been compiled. I really think `binary` is the right term to use here, at least from Zinc's perspective, and we should keep it. Users should understand Zinc notation to use the API, and for that we can better document what binary means. But, since we're in the JVM, binary is the right choice because it also tells the user that we're conforming to JLS, which has the nicety that the user knows what's the supposed format of this name.

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
jvican: You're right, this is incorrect. Will fix real quick.

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
eed3si9n: In Zinc code base "binary" could mean `*.jar` file, `*.class` file, `*.class` representation or output of `*.scala` construct, zeros-and-ones serialization form, and finally it could also mean a pair of class name and companion object class. Which meaning is it used here?

jvican: In this case, `binary` represents a classpath entry (i.e. from where the Scala compiler has unpickled the info of `onBinaryClassName). That classpath entry is either a JAR or the exact class file where it comes from (so, in this case, it will be a file inside a passed classpath entry since you cannot pass files directly in the class file).

jvican: I can document it better if you want, but since we talk about binary dependencies, we should perhaps rename `binary` to `binaryEntry` with the previous description I've given.
  
This should not be difficult to understand for Zinc users since this API only deals with dependencies and compiler-related information, so it should not be confused with other contexts.

jvican:
This is ready for review again. Comments addressed.

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
eed3si9n: If it represents a classpath entry, why not call it `classpathEntry`?
There's https://github.com/sbt/zinc/blob/1.0/internal/compiler-interface/src/main/java/xsbti/compile/PerClasspathEntryLookup.java

jvican: Because it's not a classpath entry per se. A classpath entry cannot be a single class file, it has to be a either a directory or a jar. For people knowing this, it can be confusing because they would expect a directory to be passed, not the class file where the binary name actually comes from (which is what you shall pass).

internal/compiler-interface/src/main/java/xsbti/compile/IncrementalCompiler.java
eed3si9n: Zinc (and sbt 1.0) is based on Scala 2.12, so we've already made the commitment to JDK 8 and above. The latest `Position` already uses java.util.Optional for example - https://github.com/sbt/util/blob/a12045ed4282b26179584593bf553a06b42ad1cc/internal/util-interface/src/main/java/xsbti/Position.java

jvican: @eed3si9n Are you sure you want to make that commitment? We also compile to 2.11, so the folks that want to use that won't be able to run it on JDK7. Honestly, I think that a lot of people are stuck in it yet.
  
What I propose is that we keep this out of this PR and discuss it separately in another issue.

eed3si9n: ok. Let's take it somewhere else.

eed3si9n (approved):

internal/compiler-interface/src/main/contraband/incremental.json
```diff
@@ -124,29 +117,6 @@
           ]
         },
         {
-          "name": "antStyle",
```
dwijnand, line 127: The scala build currently "relies" on antStyle, for some definition of "relies":

* https://github.com/scala/scala/blob/5f1a638cf5f3b81fd3d8b4fa92a5a52ac5733517/build.sbt#L884
* https://github.com/scala/scala/blob/5f1a638cf5f3b81fd3d8b4fa92a5a52ac5733517/README.md#incremental-compilation

Are we just going to ignore them? Or has this been cleared with them on some back-channel?

internal/compiler-interface/src/main/contraband/incremental.json
```diff
@@ -103,13 +103,6 @@
           ]
         },
         {
-          "name": "nameHashing",
```
dwijnand, line 106: **Please** can we have a deprecation cycle when we remove things??

Upgrading sbt/sbt (and I assume every other client of sbt/zinc) is such a PITA if we change things like this.

internal/compiler-interface/src/main/java/xsbti/compile/DefinesClass.java
```diff
@@ -8,14 +8,19 @@
...
+ * Determine whether a classpath entry contains a class.
  *
- * The corresponding classpath entry is not exposed by this interface. It's tied
- * to it by a specific class that implements this interface.
+ * The corresponding classpath entry is not exposed by this interface.
  */
 public interface DefinesClass {
     /**
-     * Returns true if the classpath entry contains the requested class.
+     * Return true if the classpath entry contains the requested class.
```
dwijnand, line 17: Docs should be in the 3rd person.

So "DefinesClass determines whether..." and "DefinesClass.apply returns true if...". Not "determine" and "return".

internal/compiler-interface/src/main/java/xsbti/compile/IncOptionsUtil.java
```diff
@@ -7,16 +7,17 @@
...
+import xsbti.Maybe;
+
 import java.io.File;
 import java.util.HashMap;
 import java.util.Map;
-import xsbti.Maybe;
 
 /**
- * Helper object for xsbti.compile.IncOptions
+ * Define a helper class to instantiate {@link IncOptions}.
```
dwijnand, line 17: In what way does IncOptionsUtil "define a helper class"?

It _is_ a helper class/object.

zinc/src/main/scala/sbt/inc/IncrementalCompilerUtil.scala
```diff
@@ -1,17 +0,0 @@
...
- * This software is released under the terms written in LICENSE.
- */
-
-package sbt
-package inc
-
-import xsbti.compile.IncrementalCompiler
-import sbt.internal.inc.IncrementalCompilerImpl
-
-object IncrementalCompilerUtil {
```
dwijnand, line 14: Why did we move this to `xsbti.compile`?

Even if we should move it (which is debatable) we should move it with a deprecation cycle. Unless it's strictly not possible.

internal/compiler-interface/src/main/java/xsbti/compile/IncrementalCompilerUtil.java
```diff
@@ -0,0 +1,63 @@
...
+     * the compiler interface to be used and {@link ClasspathOptions}.
+     *
+     * @param scalaInstance The Scala instance to be used.
+     * @param compilerBridgeJar The jar file of the compiler bridge.
+     * @param classpathOptions The options of all the classpath that the
+     *                         compiler takes in.
+     *
+     * @return A Scala compiler with the given configuration.
+     */
+    ScalaCompiler scalaCompiler(ScalaInstance scalaInstance,
```
dwijnand, line 24: Just so it's clear and we agree to what we're doing:

this is the API that @eed3si9n had intentionally removed from public API in #30, now brought back.

dwijnand (commented):
Some food-for-thought, post-merge.

internal/compiler-interface/src/main/java/xsbti/compile/DefinesClass.java
jvican: My understanding is that docs should be in imperative. I'll double-check.

internal/compiler-interface/src/main/contraband/incremental.json
jvican: It is my understanding that deprecation cycles are useful when evolving APIs, in minor versions. Zinc 1.0 brings to the table a major version with a completely reworked API. IMO, we should not use them here.

internal/compiler-interface/src/main/contraband/incremental.json
jvican: Ant style was already deprecated and removed from the sources. This commit only removes the rest that was not removed back then. I don't know who did it, but there was no implementation making use of this option, IIRC. In any case, I think that now it's the time to remove it because this option was never used and we kept it in the 0.13.13 cycle for reproducibility.

internal/compiler-interface/src/main/java/xsbti/compile/IncOptionsUtil.java
jvican: True, thanks for spotting!

zinc/src/main/scala/sbt/inc/IncrementalCompilerUtil.scala
jvican: Same comment on deprecations as I did some comments ago. I don't think we should use deprecations here. We're reworking the whole API from scratch.

internal/compiler-interface/src/main/java/xsbti/compile/IncrementalCompilerUtil.java
jvican: I was not aware of this, I'll have a look. It's not clear to me why this was removed.

internal/compiler-interface/src/main/contraband/incremental.json
dwijnand: Not using them is making your project's pre-release clients' life hell.

I don't follow all changes that happen in Zinc. But a few times now I've dealt with breaking changes with absolute no guidance.

Maybe we shouldn't update sbt/sbt any more until Zinc can start being more cooperative.

internal/compiler-interface/src/main/contraband/incremental.json
dwijnand: I see, my mistake. I hadn't realised the damage had already been done. My apologies.

internal/compiler-interface/src/main/contraband/incremental.json
jvican: I think that to make migration easier we should describe all the breaking changes in a document. But I don't think we have to use deprecations, we're doing a face lift to Zinc so this is the right moment to remove deprecations and do breaking changes. Otherwise we're doomed to maintain the previous, unstable API.

zinc/src/main/scala/sbt/inc/IncrementalCompilerUtil.scala
eed3si9n: I didn't catch this move. Maybe there's a bit of confusion on what APIs are from who's point of view.

jvican: This was moved because we need an api of utils from which people can get default implementations without exposing the Scala implementations. This move is the first step towards this direction.

internal/compiler-interface/src/main/java/xsbti/compile/DefinesClass.java
dwijnand: If you're happy to accept a Java style guide, there's this:

> + Use 3rd person (descriptive) not 2nd person (prescriptive). 
> The description is in 3rd person declarative rather than 2nd person imperative.
> Gets the label. (preferred)
> Get the label. (avoid)

http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#styleguide

internal/compiler-interface/src/main/contraband/incremental.json
dwijnand: > I think that to make migration easier we should describe all the breaking changes in a document.

Migration from what to what?

In https://github.com/sbt/sbt/pull/2969 I'm trying to deal with breaking changes between zinc 1.0.0-X8 and zinc 1.0.0-X9 and suffering the consequences. Are you going to write me a migration guide for that?

> But I don't think we have to use deprecations, we're doing a face lift to Zinc so this is the right moment to remove deprecations and do breaking changes. Otherwise we're doomed to maintain the previous, unstable API.

You're free to define your deprecation cycles to have the version or time span you see fit. Just tell me what they are. If it's as quick as "deprecated in X9, removed in X10, so you might have to upgrade to X9 before upgrading to X10" that's fine. At least I don't have to reverse engineer Zinc pull requests to get sbt to compile.

jvican: > You're free to define your deprecation cycles to have the version or time span you see fit. Just tell me what they are. If it's as quick as "deprecated in X9, removed in X10, so you might have to upgrade to X9 before upgrading to X10" that's fine. At least I don't have to reverse engineer Zinc pull requests to get sbt to compile.
  
This causes an overhead that slows us down. Zinc at its current state is not stable enough to deserve these procedures, IMO.
  
> In sbt/sbt#2969 I'm trying to deal with breaking changes between zinc 1.0.0-X8 and zinc 1.0.0-X9 and suffering the consequences. Are you going to write me a migration guide for that?
  
Yes, I will. IMO, this can be very good for sbt, pants and cbt maintainers. We can have a new policy in which every breaking change has to be documented in the release notes.

internal/compiler-interface/src/main/java/xsbti/compile/DefinesClass.java
jvican: Nice, I'll do it this way.

internal/compiler-interface/src/main/contraband/incremental.json
dwijnand: I'm happy with that. Let's no forget this new policy.

PR #237, opened 2017-02-24
Make ExtractAPI#annotations phase travel to entering typer, again

dwijnand:
This is a revert of the change that was made in:

https://github.com/sbt/zinc/pull/86/commits/b0a8a91aaa5b6f55714986f7bf389ec8aa091ec0#diff-babf52667b7dd64d625e0bcddfd8da48R211

Here's my analysis of what happened.

Prior to creating this commit Greg was working on class-based
name-hashing in a branch in sbt/sbt, targeting sbt 0.13.  With this
commit Greg was re-syncing his changes with the changes that had
happened in the sbt/zinc repo, to be able to PR sbt/zinc.

However, there was an important difference in sources between sbt/sbt
and sbt/zinc. The Scala compiler/reflect APIs change over time, for
instance in scala 2.11 the "atPhase" method was deprecated in favour of
the brand-new, more precise, but otherwise identical "enteringPhase"
method.  But sbt/sbt didn't care about deprecations and just used
"atPhase" for all versions of Scala it supported (2.8/2.9/etc).  In
sbt/zinc, instead, the sources had been split into 2.10 and 2.11+
sources, and the 2.11+ sources had switched to the newer, non-deprecated
APIs. So "atPhase" had been replaced with "enteringPhase".

So when Greg was re-syncing the changes in his branch with the changes
in sbt/zinc he had to deal with (a) the sources being split and (b) the
compiler API differences. At this point I think 1 of the 2 following
plausible scenarios happened. Either:

1. For one reason or another Greg accidentally, and incorrectly, thought
it was the correct implementation, perhaps he forgot that atPhase and
enteringPhase are synonymous. So he made the 2.11+ code invoke
enteringPhase(currentRun.typerPhase.next) instead of
enteringPhase(currentRun.typerPhase), thinking it was the correct
equivalent to atPhase(currentRun.typerPhase) in the 2.10 code.

2. Greg intentionally, purposely, made, all of a sudden, a judgement
call, while syncing, dealing with merge conflicts and file path changes,
that ExtractAPI#annotations MUST phase travel to exiting the typer phase
(via enteringPhase(typer.next)). And then he (a) changed _only_ the
2.11+ code, leaving the the 2.10 code travel to the wrong place _and_
(b) didn't mention it in his "highlight of less obvious changes" in the
commit message.

I find 1 to be much, much more plausible, so let's revert to what has
worked since its inception in 2010:

https://github.com/sbt/sbt/commit/af4f41e052e6e077c2d590ba8ddbfda19769f280#diff-d31107a2fcd3c8482cbac4ee198375b6R311

and today in the latest stable release, 0.13.13:

https://github.com/sbt/sbt/blob/v0.13.13/compile/interface/src/main/scala/xsbt/ExtractAPI.scala#L189

---

/cc @gkossakowski @retronym

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
@@ -209,7 +209,7 @@ class ExtractAPI[GlobalType <: Global](
     }
 
   private def annotations(in: Symbol, s: Symbol): Array[xsbti.api.Annotation] =
-    enteringPhase(currentRun.typerPhase.next) {
+    enteringPhase(currentRun.typerPhase) {
       val base = if (s.hasFlag(Flags.ACCESSOR)) s.accessed else NoSymbol
       val b = if (base == NoSymbol) s else base
       // annotations from bean methods are not handled because:


jvican:
We've been discussing this all day [here](https://gitter.im/sbt/zinc-contrib).
  
My take: this change may or may have not affected the behaviour of Zinc. It's not clear yet. The fact that our current Zinc for 2.11 runs *after* typer means that we could be affected by the desugarings produced in typer, or weird synthesized members or transformations we're not aware of. As far as I know, typer is dealing with a lot of Scala features, not only typing (e.g. default arguments, named arguments, implicit desugarings). So perhaps this could have an effect on the way Zinc tracks dependencies (this is unlikely, but it could be possible).
  
Reverting the change sounds good to me, though. I think this was an accident and the fact that Zinc has been using `atPhase(currentRun.typer)` for more than 4 years makes me think this is the way it's supposed to be. Otherwise a good  explanation would exist in commit messages (which doesn't).
  
This is the change that introduced `atPhase` for the first time in `ExtractAPI`: https://github.com/gkossakowski/sbt/commit/af4f41e052e6e077c2d590ba8ddbfda19769f280.
  
@romanowski suggested to run this change on several well-known Scala projects and see what happens. I think that's a good idea, let's see what the results are.

dwijnand:
I forgot to mention: this change removes a difference between the 2.10 and 2.11+ implementations, see this diff between the two:

```scala
--- internal/compiler-bridge/src-2.10/main/scala/xsbt/ExtractAPI.scala	2017-02-24 16:58:41.000000000 +0000
+++ internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala	2017-02-24 
@@ -204,3 +211,3 @@
   private def annotations(in: Symbol, s: Symbol): Array[xsbti.api.Annotation] =
-    atPhase(currentRun.typerPhase) {
+    enteringPhase(currentRun.typerPhase.next) {
       val base = if (s.hasFlag(Flags.ACCESSOR)) s.accessed else NoSymbol
```

**NOTE:** Don't be confused,`atPhase` and `enteringPhase` are strictly equivalent:

https://github.com/scala/scala/blob/v2.12.1/src/reflect/scala/reflect/internal/SymbolTable.scala#L420

---

So:

> this change may or may have not affected the behaviour of Zinc. It's not clear yet.

Given that none of the tests aren't failing, particularly the 2.10 tests, if this does affect the behaviour of Zinc then it must _only_ affect Zinc when using Scala 2.11+.

Or there's simply is no test case that warrant this to be changed originally.

@jvican Would you prefer if I changed the 2.10 code to `enteringPhase(currentRun.typerPhase.next)`?

jvican:
I think it's good as-is. Rgarding the change of behaviour, sorry for not being explicit, I meant change of behaviour for 2.11+, when the change was introduced.

romanowski:
I love PRs that have 100 times more description and discussion then actual code changes :)

Mentioned builds succeeded but this was 'smoke tests' (big codebase was compiled without any exception/errors, nothing more).

I think we should merge it.

jvican:
Do we have tests to check the correct behaviour of annotations in Scala? Otherwise, it would be a good addition to the PR.

dwijnand:
_I_ don't.

dwijnand:
@jvican 

> I like your emphasis on the two scenarios
It's obvious you don't know what's the real reason

I've been studying what phase travelling actually means, and I've found some informative commit messages from Paul:

+ https://github.com/scala/scala/commit/4573e06a278e9505b7cf1b6f1390e2833fb70fec
+ https://github.com/scala/scala/commit/da12146b8f3a5215e9685cd978d898131f67f1eb
+ https://github.com/scala/scala/commit/2e69ae41d9b4af6dd87e14e5f6f980c51e164e4c

So it would seem that since its inception 7 years ago and still today in 0.13.13 ExtractAPI has travelled to the beginning of the typer phase when doing whatever it does in "def annotations". Except for as of 1 year ago where the 2.11+ sources here (only pre-released) were switch so it runs in at the beginning of patmat (or whatever happens to be the phase after typer).

Edit: I think that for Scala 2.10-2.12, even when running under Zinc, the phase after typer is always patmat, unless you specific differently with the sysprop sbt.api.phase.

jvican:
Okay, I'm weighing in again.
  
After giving it some thought, I think this PR fixes a bug in 2.11 sources of Zinc. Running after typer is bad -- I think that some annotations are removed by the typer because it applies its transformations. For instance, `varargs`. If we have a method annotated with `varargs` and we remove it, it changes the generated bytecode (on which a potential user could be relying on). However, before this PR, Zinc would not invalidate it because it wouldn't detect a change in its API. The same could happen for other annotations that affect the bytecode generation of the API. I don't know how may of these we have *now*, but since it's possible for them to exist, we need to guard against them and run before `typer`.
  
So, I would propose that we merge this change. I'm pretty sure it's the right fix.

dwijnand:
Thanks Jorge, that sounds right to me too.

jvican:
Okay, I've just double-checked. This does not only affect `varargs`, it also affects annotations like `field` and `getter` since typer is the responsible of synthesizing getters and setters.

dwijnand:
In a weird way, thank god that the sources were duplicated and only one version was changed. It allowed for an easy way to discover this bug. It's the risk you pay when you merge +11,154 −3,866 PRs.

romanowski:
Just to be sure I played with different annotations and compare output with and without ```.next``` and results was always the same. 

Now am almost 100% sure that merge was good decision.

Issue #230, opened 2017-02-20
Zinc has inconsistent names: ClassfileManager vs ClassFileManager


jvican:
It seems that https://github.com/sbt/zinc/commit/47db3df0c6914d4a479917f38f07eee1e48e61ec introduced a name change in `ClassfileManager`, turning it into `ClassFileManager`. This change is inconsistent with the object living in `sbt.internal.inc` that is named `ClassfileManager`.
  
For consistency reasons, we should stick with one style. It seems the whole new Java API is written in the capital letters style, so I slightly lean towards renaming the object `ClassfileManager` to `ClassFileManager`. However, I wonder if this change is worth it because IDEs and key Scala projects like Scala.js need to keep track of these changes. Yes, we've already changed a lot the API, another change doesn't matter. But yet, I'd like to start a discussion on this because it's not clear to me where we should care about users and where we shouldn't. This affects current build tools like Pants (/cc @stuhood @peiyuwang), Bazel, sbt (@dwijnand, @eed3si9n), and the aforementioned Scalajs (@sjrd, perhaps also Scala Native? /cc @densh).

jvican:
Related patch https://github.com/sbt/zinc/commit/47db3df0c6914d4a479917f38f07eee1e48e61ec aside from https://github.com/sbt/zinc/commit/47db3df0c6914d4a479917f38f07eee1e48e61ec.

romanowski:
+1

It used to be `ClassfileManager` but was recently changed (in some places) to `ClassFileManager`.

peiyuwang:
Also +1 for unifying naming. 

fwiw, inconsistency already exists before 47db3df0c6914d4a479917f38f07eee1e48e61ec. So when `ExternalHooks.ClassFileManager` is merged with trait `ClassfileManager` I sticked to the Java naming. I probably should have called out the in the PR, my appologies.

FYI, in case API changes are unavoidable, Pants has a [deprecation process](http://www.pantsbuild.org/deprecation_policy.htm) that works for us so far.

jvican:
@peiyuwang Indeed, that change was introduced before, when the Java API was designed with contraband. Sorry for the misunderstanding.
  
I don't know how malleable is the new Zinc API in terms of binary compatibility. My understanding is that we msut have a binary compatible API that we can maintain over two years (at the very least).
  
[...]

PR #252, opened 2017-03-03
Rename ClassfileManager to ClassFileManager for consistency

knirski:
Fix naming inconsistency described in #230

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,3 +1,7 @@
 - Support for Scala 2.7, 2.8 and 2.9 has been dropped. [#86][86]
 
   [86]: https://github.com/sbt/zinc/pull/86
+  
+- sbt.internal.inc.ClassfileManager has been renamed to sbt.internal.inc.ClassFileManager for consistency. [#87][87]
+
+  [87]: https://github.com/sbt/zinc/pull/86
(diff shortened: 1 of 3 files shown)


MIGRATION.md
```diff
@@ -1,3 +1,7 @@
 - Support for Scala 2.7, 2.8 and 2.9 has been dropped. [#86][86]
 
   [86]: https://github.com/sbt/zinc/pull/86
+  
+- sbt.internal.inc.ClassfileManager has been renamed to sbt.internal.inc.ClassFileManager for consistency. [#87][87]
```
eed3si9n, line 5: I don't think internal class name change needs to be included to migration guide.

eed3si9n (changes requested):

MIGRATION.md
jvican: I would suggest to include these changes in the MIGRATION guide because sbt and pants maintainers will need to migrate it, and it's better that we let them know the changes straight away. We can remove these internal changes for the final MIGRATION guide from 0.13.x to 1.0.x.

jvican (approved):
Thanks for investing your time into this @sepulkarz! You've done a valuable work.
  
Can @eed3si9n sign off on this PR and we merge it right away?

eed3si9n (approved):

jvican:
Thanks for merging @eed3si9n!

PR #253, opened 2017-03-03
Replace use of Scala collections by Java's

wlk:
issue #245 

Tested on `Intel(R) Core(TM) i7-3610QM CPU @ 2.30GHz`

shapeless benchmark on `1.0`:
```
[info] Benchmark                                                                   (_tempDir)    Mode  Cnt           Score            Error   Units
[info] HotShapelessBenchmark.compile                                        /tmp/sbt_50a6156d  sample   18       18956.390 ±        470.882   ms/op
[info] HotShapelessBenchmark.compile:compile·p0.00                          /tmp/sbt_50a6156d  sample            18018.730                    ms/op
[info] HotShapelessBenchmark.compile:compile·p0.50                          /tmp/sbt_50a6156d  sample            19142.803                    ms/op
[info] HotShapelessBenchmark.compile:compile·p0.90                          /tmp/sbt_50a6156d  sample            19555.523                    ms/op
[info] HotShapelessBenchmark.compile:compile·p0.95                          /tmp/sbt_50a6156d  sample            19797.115                    ms/op
[info] HotShapelessBenchmark.compile:compile·p0.99                          /tmp/sbt_50a6156d  sample            19797.115                    ms/op
[info] HotShapelessBenchmark.compile:compile·p0.999                         /tmp/sbt_50a6156d  sample            19797.115                    ms/op
[info] HotShapelessBenchmark.compile:compile·p0.9999                        /tmp/sbt_50a6156d  sample            19797.115                    ms/op
[info] HotShapelessBenchmark.compile:compile·p1.00                          /tmp/sbt_50a6156d  sample            19797.115                    ms/op
[info] HotShapelessBenchmark.compile:·gc.alloc.rate                         /tmp/sbt_50a6156d  sample   18         176.279 ±          4.358  MB/sec
[info] HotShapelessBenchmark.compile:·gc.alloc.rate.norm                    /tmp/sbt_50a6156d  sample   18  3598116903.111 ±    8115563.410    B/op
[info] HotShapelessBenchmark.compile:·gc.churn.Compressed_Class_Space       /tmp/sbt_50a6156d  sample   18           0.005 ±          0.006  MB/sec
[info] HotShapelessBenchmark.compile:·gc.churn.Compressed_Class_Space.norm  /tmp/sbt_50a6156d  sample   18      110556.889 ±     121365.225    B/op
[...]

PR #257, opened 2017-03-05
Replace use of Scala collections by Java's - alternative version

wlk:
Alternative to: https://github.com/sbt/zinc/pull/253

In this PR I'm using `allUsedNames.asScala.foreach{....}`


jvican:
Hey @wlk, thanks for doing this duplicate! I think that using `asScala` is not ideal and since we're using Java collections we may want to go all the way down :wink:. Closing, and we'll focus on the other PR!

PR #258, opened 2017-03-05
Fix compiler warnings

wlk:
I noticed that there are a lot of compiler warnings during build, this PR attempts to fix (a large) portion of them:
* unused imports
* unused private methods
* unused vals

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
@@ -162,6 +162,7 @@ class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) ext
         def usedNameInImportSelector(name: Name): Unit = {
           if (!isEmptyName(name) && (name != nme.WILDCARD) && !names.contains(name)) {
             names += name
+            ()
           }
         }
         selectors foreach { selector =>
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -24,9 +24,8 @@ object ClassToAPI {
   // (api, public inherited classes)
   def process(classes: Seq[Class[_]]): (Seq[api.ClassLike], Set[(Class[_], Class[_])]) =
     {
-      val pkgs = packages(classes).map(p => new api.Package(p))
       val cmap = emptyClassMap
-      val defs = classes.flatMap(toDefinitions(cmap))
+      classes.foreach(toDefinitions(cmap)) // force recording of class definitions
       cmap.lz.foreach(_.get()) // force thunks to ensure all inherited dependencies are recorded
       val classApis = cmap.allNonLocalClasses.toSeq
       val inDeps = cmap.inherited.toSet
@@ -72,6 +71,7 @@ object ClassToAPI {
 
   def toDefinitions(cmap: ClassMap)(c: Class[_]): Seq[api.ClassLikeDef] =
     cmap.memo.getOrElseUpdate(classCanonicalName(c), toDefinitions0(c, cmap))
+
   def toDefinitions0(c: Class[_], cmap: ClassMap): Seq[api.ClassLikeDef] =
     {
       import api.DefinitionType.{ ClassDef, Module, Trait }
@@ -109,7 +109,7 @@ object ClassToAPI {
     if (!Modifier.isPrivate(c.getModifiers))
       cmap.inherited ++= parentJavaTypes.collect { case parent: Class[_] => c -> parent }
     val parentTypes = types(parentJavaTypes)
-    val instanceStructure = new api.Structure(lzyS(parentTypes.toArray), lzyS(all.declared.toArray), lzyS(all.inherited.toArray))
+    val instanceStructure = new api.Structure(lzyS(parentTypes), lzyS(all.declared.toArray), lzyS(all.inherited.toArray))
     val staticStructure = new api.Structure(lzyEmptyTpeArray, lzyS(all.staticDeclared.toArray), lzyS(all.staticInherited.toArray))
     (staticStructure, instanceStructure)
   }
@@ -133,7 +133,6 @@ object ClassToAPI {
   private val emptySimpleTypeArray = new Array[xsbti.api.SimpleType](0)
   private val lzyEmptyTpeArray = lzyS(emptyTypeArray)
   private val lzyEmptyDefArray = lzyS(new Array[xsbti.api.ClassDefinition](0))
-  private val lzyEmptyStructure = lzyS(new xsbti.api.Structure(lzyEmptyTpeArray, lzyEmptyDefArray, lzyEmptyDefArray))
 
   private def allSuperTypes(t: Type): Seq[Type] =
     {
@@ -166,7 +165,7 @@ object ClassToAPI {
 
   @deprecated("Use fieldToDef[4] instead", "0.13.9")
   def fieldToDef(enclPkg: Option[String])(f: Field): api.FieldLike = {
-    val c = f.getDeclaringClass()
+    val c = f.getDeclaringClass
     fieldToDef(c, classFileForClass(c), enclPkg)(f)
   }
 
@@ -249,7 +248,7 @@ object ClassToAPI {
     def ++(o: Defs) = Defs(declared ++ o.declared, inherited ++ o.inherited, staticDeclared ++ o.staticDeclared, staticInherited ++ o.staticInherited)
   }
   def mergeMap[T <: Member](of: Class[_], self: Seq[T], public: Seq[T], f: T => api.ClassDefinition): Defs =
-    merge[T](of, self, public, x => f(x) :: Nil, splitStatic _, _.getDeclaringClass != of)
+    merge[T](of, self, public, x => f(x) :: Nil, splitStatic, _.getDeclaringClass != of)
 
   def merge[T](of: Class[_], self: Seq[T], public: Seq[T], f: T => Seq[api.ClassDefinition], splitStatic: Seq[T] => (Seq[T], Seq[T]), isInherited: T => Boolean): Defs =
     {
@@ -339,11 +338,11 @@ object ClassToAPI {
       val targs = t.getActualTypeArguments
       val args = if (targs.isEmpty) emptyTypeArray else arrayMap(targs)(t => reference(t): api.Type)
       val base = reference(t.getRawType)
-      new api.Parameterized(base, args.toArray[api.Type])
+      new api.Parameterized(base, args)
     }
   def reference(t: Type): api.SimpleType =
     t match {
-      case w: WildcardType       => reference("_")
+      case _: WildcardType       => reference("_")
       case tv: TypeVariable[_]   => new api.ParameterRef(typeVariable(tv))
       case pt: ParameterizedType => referenceP(pt)
       case gat: GenericArrayType => array(reference(gat.getGenericComponentType))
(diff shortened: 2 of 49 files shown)


jvican:
This is so great @wlk :smile:!
  
Let us know when this is ready, you probably want to squash all the commits since the previous ones were failing. That way, we make sure that commits pass build 👍

.gitignore
```diff
@@ -1,2 +1,3 @@
 target/
 zinc/src/test/resources/bin
+.idea
```
dwijnand: If you are an IntelliJ IDEA user I would suggest you configure your global gitignore file, instead of expecting every project to include these ignores or adding them while submitting a pull request.

GitHub has a nice little help article about it: https://help.github.com/articles/ignoring-files/.

The general principle is that .gitignore should only include paths for files that are generated by running the build, and not include a random assortment of paths generated by other tools.

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
```diff
@@ -162,6 +162,7 @@ class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) ext
         def usedNameInImportSelector(name: Name): Unit = {
           if (!isEmptyName(name) && (name != nme.WILDCARD) && !names.contains(name)) {
             names += name
+            ()
```
dwijnand, line 165: Just out of my own curiosity, could you explain this change?

internal/zinc-apiinfo/src/main/scala/xsbt/api/SameAPI.scala
```diff
@@ -15,32 +15,20 @@ import scala.collection.{ immutable, mutable }
...
-		println(ShowAPI.show(b))*/
-
-      val result = (new SameAPI(false, true)).check(a, b)
-      val end = System.currentTimeMillis
-      //println(" API comparison took: " + (end - start) / 1000.0 + " s")
-      result
-    }
+  def apply(a: ClassLike, b: ClassLike): Boolean = {
+    new SameAPI(false, true).check(a, b)
+  }
```
dwijnand: Could you drop the brackets and one-line this please? Looks short enough.

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
```diff
@@ -45,16 +42,16 @@ private[sbt] object Analyze {
...
       classFile = Parser(newClass);
-      sourceFile <- classFile.sourceFile orElse guessSourceName(newClass.getName);
+      _ <- classFile.sourceFile orElse guessSourceName(newClass.getName);
       source <- guessSourcePath(sourceMap, classFile, log);
       binaryClassName = classFile.className;
       loadedClass <- load(binaryClassName, Some("Error reading API from class file"))
     ) {
       binaryClassNameToLoadedClass.update(binaryClassName, loadedClass)
-      val srcClassName = binaryToSourceName(loadedClass)
+      val maybeSrcClassName = binaryToSourceName(loadedClass)
```
dwijnand: I don't see the need for the Hungarian notation. I prefer the former.

dwijnand (changes requested):
Thank you for spending time on this @wlk!

I have a few questions / requested changes, but other than that this looks good.

Welcome to Zinc.

.gitignore
wlk: no problem

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
wlk: There is a warning `discarded non-Unit value` I followed the suggestion from here: http://stackoverflow.com/questions/13415307/suppress-discarded-non-unit-value-warning (also saw this pattern in few other places in code)

dwijnand: Oh discarded _inside_ the if block, of course. Thanks.

internal/zinc-apiinfo/src/main/scala/xsbt/api/SameAPI.scala
wlk: fixed

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
wlk: No problem, I did this because few lines below we are reusing name `srcClassName` inside `match` -> I renamed the value inside the match to avoid confustion

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
```diff
@@ -55,9 +52,9 @@ private[sbt] object Analyze {
         .orElse(Option(loadedClass.getEnclosingClass).flatMap(binaryToSourceName))
 
       srcClassName match {
-        case Some(srcClassName) =>
-          analysis.generatedNonLocalClass(source, newClass, binaryClassName, srcClassName)
-          classNames += srcClassName
+        case Some(className) =>
+          analysis.generatedNonLocalClass(source, newClass, binaryClassName, className)
+          classNames += className
```
dwijnand, line 57: One of the wonderful things about pattern matching (in my opinion) is the ability to shadow.

Was this change triggered by a warning? If so, what was the warning about? And who emitted it?

wlk: Good observation, this was warning generated by intellij "Suspicious shadowing by a Variable Pattern". I don't have strong opinion how to approach this but usually I try to avoid shadowing by using different names

dwijnand: Ah right, yeah I have that inspection turned off. This is its motivation:

> Detects a Variable Pattern that shadows a stable identifier defined in the enclosing scope. To perform an equality test against that value, use backticks.
>
> Before:
> ```scala
> val foo = 0
> 0 match {
>     case foo =>
> }
> ```
>
> After:
> ```scala
> val foo = 0
> 0 match {
>     case `foo` =>
> }
> ```

Food for thought :-)

dwijnand (approved):

dwijnand:
Approved, pending Travis CI building it successfully obviously.

jvican (approved):
LGTM!
  
One minor suggestion: squash all the commits :smile:. Let's merge when CI passes.

dwijnand:
You're getting

> [info] ! Parser.able to parse all relevant classes: Exception raised on property evaluation.
[info] > ARG_0: class sbt.internal.inc.classfile.ParserSpecification$
[info] > Exception: java.lang.RuntimeException: Unknown constant: 97

in the 2.11 build. And a whole lot of errors in the 2.12 build.

wlk:
@dwijnand I'll have more time to look at this tomorrow, most likely I have removed import that is reported as unused, but is in fact used (I have already stumbled on few cases like this here)

jvican:
> I'll have more time to look at this tomorrow, most likely I have removed import that is reported as unused, but is in fact used (I have already stumbled on few cases like this here)
  
Indeed, this happens because Intellij knows that you're using a concrete Scala version. The issue is that we use implicits for compatibility reasons, so that if a method is not defined in a class the compiler automatically tries to get an implicit class with that method. We use this to keep track of changes in the Scala compiler API.
  
That, mixed with some false positives from Intellij sides, may be the culprit of this. Double check that the removed imports are indeed used and you're good to go. Looking forward to merge this and remove all this clutter from my screen 👍

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
romanowski: TBH in most PRs where `()` is added sombody is asking about it.

@eed3si9n @jvican @dwijnand Maybe we should remove that compiler setting?

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -24,9 +24,9 @@ object ClassToAPI {
   // (api, public inherited classes)
   def process(classes: Seq[Class[_]]): (Seq[api.ClassLike], Set[(Class[_], Class[_])]) =
     {
-      val pkgs = packages(classes).map(p => new api.Package(p))
+      packages(classes).map(p => new api.Package(p))
```
romanowski: If we really don't use return values we should replace `map` and `flatMap` with `foreach` (and remove overhead for creating unused collections).

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/LocalJava.scala
```diff
@@ -73,14 +73,14 @@ final class LocalJavadoc() extends XJavadoc {
   override def run(sources: Array[File], options: Array[String], incToolOptions: IncToolOptions,
     reporter: Reporter, log: XLogger): Boolean = {
     val cwd = new File(new File(".").getAbsolutePath).getCanonicalFile
-    val (jArgs, nonJArgs) = options.partition(_.startsWith("-J"))
+    val (_, nonJArgs) = options.partition(_.startsWith("-J"))
```
romanowski: If we care only about 'false' part of `partition` we should use `filterNot`

romanowski (approved):
Good stuff! Finally someone was brave enough to clean up the code :)

Except to minor rewrites (that are consequence of cleanup) LGTM

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
dwijnand: No, value discarding is a real source of error, particularly with side-effecting (such as mutating) code.

One way to deal though with this is to explicitly discard the unused value. Either as a method or as an extension method (e.g .toUnit), and then the method could document its reason for existence.

The real solution is to change the compiler so -Ywarn-value-discard doesn't warn about this.type. Then you'll fix it for all codebases :-)

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: Agreed 😄

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/LocalJava.scala
jvican: Agreed 👍

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
jvican: I vote in favor of removing that warning. IMO it's unnecessary. I generally think that this codebases abuses warnings. The same goes for unused imports.
  
I don't see why this is a real source of error. I have never experienced it.

romanowski: I disagree with unused imports since later on we are left with tons on unused imports. However we don't need to apply it to all compilations - IMO we can apply it only for CIs builds (since hacking with fatal unused imports is not pleasant).

romanowski (approved):

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
wlk: I think the compiler warnings do help with keeping the code more maintainable (assuming you change code to prevent warnings), so IMO both warnings could be left in as they are, maybe `-Ywarn-value-discard` is a little to verbose but that's not a big deal.

wlk:
Commits squashed and travis should report green status in a moment. 
There are still outstanding warnings, but I think this PR big enough for now.

The warnings I didn't change are related to scaladoc and few deprecation warnings

PR #259, opened 2017-03-05
Attempt to speed up tests issue #229

liosedhel:
First approach, split test using strongly connected component parameter (computed for each node).
Not sure if this is the best approach though, as what we really care is leveraging testing time (not grouping dependencies).
zinc_dep_graph

PR #261, opened 2017-03-06
Test 2.10 bridges in Travis

jvican:
Issue #259 introduced a side effect that caused our build not to test 2.10
bridges. This PR makes sure that these test for 2.10 are run in independent
instances of the matrix.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,7 +24,7 @@ def commonSettings: Seq[Setting[_]] = Seq(
   crossScalaVersions := Seq(scala211, scala212),
   mimaPreviousArtifacts := Set(), // Some(organization.value %% moduleName.value % "1.0.0"),
   publishArtifact in Test := false,
-  commands ++= Seq(publishBridgesAndTest, publishBridges)
+  commands ++= Seq(publishBridgesAndTest, publishBridges, crossTestBridges)
 )
 
 def relaxNon212: Seq[Setting[_]] = Seq(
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -327,6 +327,16 @@ lazy val zincScripted = (project in internalPath / "zinc-scripted").
   ).
   configure(addSbtUtilScripted)
 
+lazy val crossTestBridges = {
+  Command.args("crossTestBridges", "<version>") { (state, args) =>
+    require(args.nonEmpty, "Missing Scala version argument.")
+    compilerBridgeScalaVersions.flatMap { (bridgeVersion: String) =>
+      s"wow $bridgeVersion" ::
+        s"${compilerBridge.id}/test" :: Nil
```
dwijnand: If you're going to wow once (and you're also not wow-ing back), why don't you just plz?

`plz $bridgeVersion ${compilerBridge.id}/test`

dwijnand (approved):
Good catch.

build.sbt
jvican: The reason why we don't is because `plz` causes `OutOfMemoryError`. I've actually experienced this myself locally and there's a comment in `publishBridgesAndTest` that makes reference to it.

dwijnand: It went from 9 plzes to 3 wows https://github.com/sbt/zinc/pull/211/commits/d5d59b6f4c563a5f9aa5d095295b93b912b90cf3

Is it something intrinsic about plz? At the time I just thought it was just the number of them.

Anyways, not a blocker.

jvican: Oh, didn't know that. I actually have no idea about this, I thought that `plz` had some specific implementation that caused the out of memory error.  I'll change this.

dwijnand: Sadly no sbt-doge issue spawn from the original cause.

jvican: What do you mean?

dwijnand: No issue was opened against sbt/sbt-doge about the issue that triggered the change in #211.

So there's no clear information about whether there is an underlying issue in plz.

dwijnand:
@jvican Could you have a look at this, please?

jvican:
@dwijnand Yes.

jvican:
I have just updated this @dwijnand. Merge when it's ready please, won't be online tomorrow.

dwijnand:
Thanks @jvican.

dwijnand (approved):

PR #266, opened 2017-03-15
Add a scripted test for package private class

gkossakowski:
Add a test that shows package private classes are handled properly in
zinc now. This test doesn't pass in sbt 0.13.

@johnynek ran into this problem while working on Twitter's summingbird. I'm adding a test case to note that buggy dependency tracking of package private classes has been fixed in zinc 1.0.


jvican (approved):
LGTM. Thanks for doing this @gkossakowski!

eed3si9n (approved):
Welcome back.

gkossakowski:
👍 
Thanks for snappy response.

dwijnand:
Hey! o/

Issue #227, opened 2017-02-19
Zinc doesn't handle constants (final val) correctly


jvican:
@gkossakowski explains why [here](https://github.com/sbt/sbt/issues/1543). This is a port from an old Zinc issue. The issue is that scalac inlines constants too early, and the incremental compiler doesn't see the dependency on the constants because no reference is left.
  
In essence, the fix for this one has to happen in scalac land. The germane JIRA ticket is [SI-7173](https://issues.scala-lang.org/browse/SI-7173). Adriaan already tried to fix it in [this PR](https://github.com/scala/scala/pull/4880) but it was closed for timing issues (I guess). Future solutions should build on his approach, which is the most likely to be fixed in Scalac. Jason has also done some work along these lines, see Adriaan's PR.

jvican:
Okay, I propose a partial solution to this problem (which I have implemented and will push soon). Since typer is inlining constants trees in current Scala and this is unlikely to change in the short term, I suggest that we keep track of constant types by extracting their `stringValue` and registering them.
  
Here's how I've done it. I added the following case to the `traverse` method in`ExtractUsedNamesTraverser`:
```scala
      case valDef: ValDef =>
        // Recognise final constants at type level
        if (valDef.symbol.hasFlag(Flags.FINAL)) {
          valDef.tpt.tpe match {
            case ConstantType(value) =>
              /* HACK: Register known constants at the type level since current
               * Scala typer inlines constants at the tree level. This does not
               * cover all the cases, see #227 for more information. */
              val names = getNamesOfEnclosingScope
              val constantName = NoSymbol.name.newName(value.stringValue)
              if (!isEmptyName(constantName) && !names.contains(constantName))
                names += constantName
            case _ => ()
          }
        } else ()
```
  
As you see, we're only interested in constants defined by `ValDef`.
  
This solution is partial because it does not handle the case where typer widens types on its own. This happens when we have a constant type on the right-hand side of a `ValTree` and a "super", more general type on the left-hand side, in which case typer unifies them. The following code snippet illustrates the cases where it does and doesn't work:
  
```scala
object Foo {
  final val foo = 12 // works
  final val foo: Int = 12 // doesn't work, typer widens type for declaration to `Int`
}
```
  
Note that if the definition of `foo` is inherited from a super class and is abstract, widening will also occur.
  
[...]

Issue #226, opened 2017-02-19
Zinc doesn't support wildcard-imported 'name' and 'implicit' shadowing


jvican:
This issue is ported from sbt/sbt https://github.com/sbt/sbt/issues/1751.
  
## Problem  
In essence, Zinc is unable to register dependencies of shadowed names coming from two wildcard imports. This means that this situation is not handled correctly:
  
```scala
// A.scala
import p1._
import p2._
class A {
  val b = new B()
}

// p1.scala
package p1 {
  class B
}

// p2.scala
package p2 {
  // Uncomment in next iteration -> Zinc doesn't recompile A
  // class B
} 
```
  
The problem is that `p2` definitions shadow `p1`'s and it's not possible for Zinc to know that beforehand. Zinc only know what the compiler knows, and since we don't see any reference to `p2.B`, the dependency on that `B` will **never** be registered.

## Possible solutions
What can we do about this? I'm afraid there's nothing to do.
  
The name hashing algorithm buys us a very good tradeoff to detect dependencies, but the only constraint is that dependencies are explicit throughout the trees of a given compilation unit (which is the case for every used definition).
  
There is no way to detect this new dependency reasonably because we don't know if it will ever happen, and therefore we need to guard us against any instance that could introduce it.
  
To support this use case, we would need to register phantom dependencies on all used names coming from imports used in all the enclosing scopes of a given wildcard import, for every name that is not defined in the imported symbol. In our example, we would get all the members of `p1` that are not defined in `p2` and create a "phantom" dependency on that name as if it were defined in`p2` (and bail out if it's defined).
  
For the sake of correctness, we should do that for every member defined in `p1` and any wildcard import that is present in an enclosing scope. Then, the easy part: our algorithm would identify those "phantom" dependencies when changes in API happen. If the API introduces a member with the target name of the "phantom" dependency, the sources that introduced the dependency should be recompiled.
    
But hey, we're not done yet. We also need to register phantom dependencies on names coming from inherited classes and defined present in enclosing scopes (e.g. nested objects have access to all the names in enclosing classes, so if we use a wildcard import inside that object, we need to register phantom dependencies on the `symbol` of the `Import` node for every accessible name). How fun!
  
At this point, we realise that supporting this use case is totally unfeasible. The complexity of the algorithm and the memory required to host those phantom dependencies would be very big, especially given:
* The frequency in which wildcard imports happen in Scala code.
* The large number of definitions present in any scope.
* The complex scoping rules that Scala has.
  
[...]

PR #267, opened 2017-03-15
Implement used scopes for used names and implement PatMat scope

romanowski:
## Motivation

In our large codebase we use sealed classes/traits a lot and we quite often adds new values to them (e.g. we represent algorithms this way). Children of such sealed classes/traits change quite often (and this is usually done by changing mixed traits) causing **all usages** of sealed classes to be recompiled.

### How we can improve zinc?

I am almost sure that (excluding macros and other non-standard) the only place where base trait is used that needs to be recompiled on change in child's type is pattern matching (to emit warning). To be more we need to invalidate all places where we match on type that contains mentioned class. Example below:

```scala
sealed trait A
object FooA
```

If we change ```FooA``` to ```FooA with Serializable``` we need to recompile all direct usages of ```FooA``` (this is not changed) but we don't need to recompile e.g. this: ```def bar(a: A) = println(a)```.

However every usage like:
```scala
def foo(a: A) = a match { ... }
def foor(b: Option[A]) = b match { ... }
```
Needs to be recompiled.

If we are using A inside ```case``` statements we don't need to recompile that usages (since warning won't be emitted in such case. Similarly if we are using unnaply to create value e.g. ```List(A, name) = myList()``` no warning will be emitted so we don't need to recompile this.

@retronym Please correct me if I am wrong since I am not compiler expert :)

## Implemetation detalis

In order to support this in zinc I created UseScope that tells about scope where name is used. So far we've got 3 scopes: ```Default```, ```Implicit``` and ```PatMat```. As you may guess current logic is implemented as ```Default```. ```Implicits``` are used so far only to find out if given class does not contains any implicit change (so is only used in API not to track implicit dependencies, but we may change this in future). Hashes are produced for each applicable scope so
 * plain ```def``` will declare hashes only in ```Default``` scopes
 * implicit class/def/val in both ```Default``` and ```Implicit```
 * sealed trait in ```Default``` and ```Implicit```

Hashes for ```Implicit``` and ```Default``` are equal (but we can change that) and for ```PatMat``` we include types of children.

If for given class only ```PatMat``` scope is changed we don't invalidate all places where only ```Default``` is used.

In future this mechanism can be used in #227 #226 


romanowski:
I will include benchmark results soon.

internal/zinc-persist/src/main/scala/sbt/internal/inc/EnumSetSerializer.scala
```diff
@@ -0,0 +1,31 @@
+package sbt.internal.inc
+
+import java.util
+
+import scala.reflect.ClassTag
+
+case class EnumSetSerializer[E <: Enum[E]: ClassTag](allValues: Array[E]) {
```
romanowski, line 14: This is quite naive approach. I am open for suggestion about possible alternatives.

jvican: This is scary...

romanowski: But works :) As said I am open for suggestions.

internal/zinc-core/src/main/scala/sbt/internal/inc/ClassFileManagers.scala
```diff
@@ -16,7 +16,7 @@ import collection.mutable
 import xsbti.compile.{ DeleteImmediatelyManagerType, IncOptions, TransactionalManagerType }
 import xsbti.compile.ClassFileManager
 
-object ClassFileManager {
+object ClassFileManagers {
```
jvican: Why this change to the API? I suggest we leave it as it was before since it's a "companion object" of `ClassFileManager`. I'm open to changing the name in other PR, but I don't see the motivation for this name change here.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalNameHashing.scala
```diff
@@ -36,15 +36,15 @@ private final class IncrementalNameHashing(log: sbt.util.Logger, options: IncOpt
     if (SameAPI(a, b))
       None
     else {
-      val aNameHashes = a.nameHashes
-      val bNameHashes = b.nameHashes
+      val aNameHashes = a.nameHashes()
```
jvican: `nameHashes` returns something, no? I think it's safe to leave the parens as they were :smile:.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalNameHashing.scala
```diff
@@ -94,20 +94,35 @@ private final class IncrementalNameHashing(log: sbt.util.Logger, options: IncOpt
...
+      else {
+        val byTransitiveInheritance =
+          if (transitiveInheritance.nonEmpty) s"by transitiveInheritance: $transitiveInheritance" else ""
+        val byLocalInheritance =
+          if (localInheritance.nonEmpty) s"by localInheritance: $localInheritance" else ""
+        val byMemberRef =
+          if (memberRef.nonEmpty) s"by transitiveInheritance: $memberRef" else ""
+
+        s"""$change invalidates ${all.size} classes due to ${memberRefInvalidator.invalidationReason(change)}
+           |\t$byTransitiveInheritance $byLocalInheritance $byMemberRef
```
jvican: New lines? 💃

internal/zinc-apiinfo/src/main/scala/xsbt/api/HashAPI.scala
```diff
@@ -136,11 +139,10 @@ final class HashAPI(includePrivate: Boolean, includeParamNames: Boolean, include
 
   def hashModifiers(m: Modifiers) = extend(m.raw.toInt)
 
-  def hashAPI(c: ClassLike): Hash =
+  def hashAPI(c: ClassLike): Unit =
     {
       hash = 1
       hashClass(c)
-      finalizeHash
```
jvican, line 143: Why don't we finalize hash here?

jvican (changes requested):
Hey @romanowski, I didn't want to yak shave so I've just pushed a commit that fixes some questions/concerns I had. More info in the commit message.
  
These changes look great and I'm happy to merge them! I have some questions first, though:
  
1. The provided tests are great, but we should have several scripted tests to check that this actually has the right behaviour. Can you please do that?
2. It seems you've glossed over extractors in pattern matching targets. It would be great if you include them in the tests and make sure they work (make sure to check that nested extractors work too).
3. The analysis that you provide in this PR for pattern matching is not as smart as it could be, if I understand these changes correctly. For instance:
```scala
def foo(a: A) = a match { ... }
```
  
This has to be recompiled iff the pattern match is exhaustively checking every possible child. If you have the following body `a match { case a2: SubTypeOfA => ???; case a: A => ??? }`, you don't need to recompile it since there's a default case that covers all the subtypes of A. You can generalize this as: "whenever there's a default case that covers the sealed super type, it's safe to ignore recompilation".
  
Please, confirm if this is an issue with your PR. The way I see this right now, it looks like the usability of these changes is not that great because if you `a` in the previous pattern match is `A` and not `Any`, you will always recompile. Perhaps I'm wrong and I didn't understand these changes well enough, though.
  
You don't have to do this in this PR, this is already an improvement over what we had before. We can do it as a follow up. I'm happy to implement it.
  
That aside, serialisation is far than ideal, but there's no clear escape, so I'm fine with that :wink:.

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
```diff
@@ -78,63 +76,71 @@ class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) ext
...
-                  ()
-              }
+          val scopedNamesInFirstClass = namesInFirstClass.scopedNames
+
+          namesInFirstClass.defaultNames.addAll(defaultNamesTopLevel)
+          scopedNamesTopLevel.foreach { (topLevelName, topLevelScopes) =>
+            val existingScopes = scopedNamesInFirstClass.get(topLevelName)
+            if (existingScopes == null)
+              scopedNamesInFirstClass.put(topLevelName, topLevelScopes)
+            else existingScopes.addAll(topLevelScopes)
```
romanowski, line 109: This emits warning. That is why I added ```()``` in the end.

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
```diff
@@ -78,63 +76,71 @@ class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) ext
     }
   }
 
+  private val DefaultScopes = EnumSet.of(UseScope.Default)
```
romanowski: EnumSets are mutable so both this vals should be defs

romanowski (commented):
@jvican Thanks for that.

I see that you don't like pattern matching on null but I am fine with that.

The only thing we should change is manetioned vals (to defs).

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
romanowski: BTW this is quite strange to do a review in your own PR :)

jvican: Yeah, I've realised that.

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
jvican: Thanks for the tip, this was not clear to me!

jvican:
I'm trying to see why my commit fails in CI. The compiler error makes no sense since I haven't changed anything in compat infrastructure:
  
```
[error] /home/travis/build/sbt/zinc/internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala:276: value hasSymbolField is not a member of ExtractUsedNames.this.global.Tree

[error]       case t if t.hasSymbolField =>

[error]                   ^
```

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
jvican: Oh, this is by no means my PR, it's yours @romanowski.

jvican:
Oh, I almost forgot @romanowski. Post the benchmarks when you have them, I'm curious to see the results. Here you go with some instructions on how to run them: https://github.com/sbt/zinc/blob/1.0/CONTRIBUTING.md.
  
I would use the Scala standard library, you can cherry-pick its benchmarks [from this commit](https://github.com/scalacenter/zinc/commit/3da86c61c6f90dd8cb6aee1b742d49ffdd87fd6a).

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
romanowski: Don't worry, I don't care about that much :)

jvican: I prefer to push changes and then discuss them, than discussing them and forcing you to spend time on it. It's just more efficient :wink:.

romanowski:
[Benchmarks](https://github.com/sbt/zinc/files/852478/benchmarks.txt)  as promised. I've also have full outputs if needed.

internal/zinc-apiinfo/src/main/scala/xsbt/api/HashAPI.scala
romanowski: To keep API unified and force users to use HashApi.apply.

romanowski:
Regarding @jvican review:

1. My tests are more or less scripted tests done in Scala code but I also added scripted one.

2. I don't know if I get this correctly. Do you want me to include a tests for cases like one below?
```scala
val List(Option(SealedChildren1(a)), None, Some(b: Sealed), Some(SealedChildren2(c)) = ???
```
3. You are right, this is quite simple solution. So it should say that  child of sealed trait/class changed zinc recompiles **every place where warning can be reported**.  So there is a lot of space to improve that (the only thing we need to do is to improve mechanism to decide if we depend on given name in PatMat Scope). Good and quite cheap and easy to implement option will be check if default case is provided.

jvican:
> I don't know if I get this correctly. Do you want me to include a tests for cases like one below?
  
Yes, that's right, and the same for `@`:
  
```scala
any match {
  case s @ Some(value) =>
  case None =>
```

jvican:
I will have a look at the benchmarks soon, I'm now unavailable. Thanks for responding to all the feedback @romanowski.

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
```diff
@@ -175,6 +248,10 @@ class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType) ext
     }
 
     private def handleClassicTreeNode(tree: Tree): Unit = tree match {
+      // Register names from pattern match target type in PatMatTarget scope
+      case ValDef(mods, _, tpt, _) if mods.isCase && mods.isSynthetic =>
+        updateCurrentOwner()
+        PatMatDependencyTraverser.traverse(tpt.tpe)
```
retronym, line 254: Looking at the types of synthetic case vals seems like a fairly robust way to find usages of sealed-ness. As you mentioned, this doesn't cover hidden dependencies from macros (e.g.. shapeless `Generic`). As long as we provide configuration to let people enable/disable this optimization, that's okay in my view.

jvican:
Benchmark results on this one look great, there's a small decrease of 300s in p95, but we can live with it. Thanks for making them.
  
Last thing on this PR before proposing to merge it: can we add an option to `IncOptions` that enables/disables this mode as Jason suggests?
  
In the long term, when the support for sealed classes detection improves, we can make it default. For now, though, this PR may undercompile under some scenarios (macro dependencies) while before we were overcompiling. We know macro dependencies are not common, but we should try to protect users from any case in which Zinc undercompiles.

romanowski:
Option added and optimization is off by default.

jvican:
Perfect, this is good to go @romanowski. I'll push the approve button in a few hours, I'm in mobile now. Can someone else have a look at this patch? @eed3si9n @dwijnand

jvican (approved):

internal/compiler-interface/src/main/java/xsbti/UseScope.java
```diff
@@ -0,0 +1,24 @@
...
+ * The incremental compiler uses [[UseScope]] to determine some Scala semantics
+ * assumed in the presence of a name in a concrete position. For instance,
+ * [[PatMatTarget]] is used for names that appear as the target types of a
+ * pattern match.
+ *
+ * The order of declaration of these is crucial. Don't change it.
+ */
+public enum UseScope {
+    // Don't add more than 6 scopes. Otherwise, change `Mapper` implementation.
+    Default, Implicit, PatMatTarget
```
eed3si9n: Could we add documentation on what these scopes represent?

eed3si9n (dismissed):

eed3si9n:
Added a minor comment, but overall looks good. Nice work!

romanowski:
Added some documentation to `UseScope`.

Also backported some small changes from our integration (`ClassloaderCache` can save us even 10% of compilation time, this is amazing!).

dwijnand (approved):

jvican:
> In future this mechanism can be used in #227 #226

Please, let's avoid this kind of statements without further double checking with maintainers. I explained in the Gitter channel that this is not correct, and this can lead to confusion as shown in #226 and #227.

PR #270, opened 2017-03-17
Adding pending test for #269

romanowski:
Zinc is undercompiling in this case. Reported as #269 

For certain reasons this time I have to paste the below disclaimer:
THIS PROGRAM IS SUBJECT TO THE TERMS OF THE BSD 3-CLAUSE LICENSE.

THE FOLLOWING DISCLAIMER APPLIES TO ALL SOFTWARE CODE AND OTHER MATERIALS CONTRIBUTED IN CONNECTION WITH THIS SOFTWARE:
THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS “AS IS” AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR
ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.
ONLY THE SOFTWARE CODE AND OTHER MATERIALS CONTRIBUTED IN CONNECTION WITH THIS SOFTWARE, IF ANY, THAT ARE ATTACHED TO (OR OTHERWISE ACCOMPANY) THIS SUBMISSION (AND ORDINARY
[...]

PR #272, opened 2017-03-18
Fix #269: Browse original type trees in Dependency

### LINKED ISSUE #269: Zinc is undercompiling in case of type projection where Scalac expands to another type
In case included in #270 zinc is undercompiling.

I don't get to the bottom of problem (yet) but I think this is caused by the way how we handle type fields.

jvican:
From the commit message:

```
Fix #269: Traverse original trees in `Dependency`

Original type trees have to be traversed if present to correctly handle
dependencies specified in explicit, user-defined types that are expanded
by the compiler, in cases such as type projections.

The example `fuzzy-types` is undercompiling because the prefix
`FactoryProvider` is totally lost at `Usage` after typer. This is one of
these cases where Scalac is expanding types and not leaving any trace to
keep track of the original dependency.

`FactoryProvider.type#MyFactory#Product` becomes `foo.B with
foo.Nil#Product`, and therefore the type dependency traverser doesn't
see `FactoryProvider`. Traversing original type trees fixes the issue.
```


jvican:
The first two commits should go away as soon as #270 is merged.

jvican:
This has been rebased, it's awaiting for review.

eed3si9n (approved):
LGTM

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -376,8 +376,16 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
...
-      // In some cases (eg. macro annotations), `typeTree.tpe` may be null. See sbt/sbt#1593 and sbt/sbt#1655.
+      /* Original type trees have to be traversed because typer is very
+       * aggressive when expanding explicit user-defined types. For instance,
+       * `Foo#B` will be expanded to `C` and the dependency on `Foo` will be
+       * lost. This makes sure that we traverse all the original prefixes. */
       case typeTree: TypeTree if !ignoredType(typeTree.tpe) =>
+        val original = typeTree.original
+        if (original != null && !inspectedOriginalTrees.contains(original)) {
+          traverse(original)
+          inspectedOriginalTrees.add(original)
```
Duhemm, line 387: Can we get cycles here?

`inspectedOriginalTrees` was introduced because there were some macros that had themselves as original trees.

jvican: *Theoretically* yes.
  
In addition to that, it serves us as a caching layer, which was my primary goal when implementing it.

jvican:
Can you have a look at this @dwijnand?
  
Also, FYI @eed3si9n & @dwijnand, I was thinking it would be a good idea to cut a release of Zinc soon that includes:
  
1. This fix.
2. Handling of private members (working on it).
3. Handling of constants (working on it).
  
WDYT guys?

romanowski:
@jvican Did you run benchmarks? I wonder how much overhead is added here.

I think release is good idea and as you may expect I vote for my pr #267 be included :)

jvican:
I have not run benchmarks on this one. Why? Because:
  
1. It's a correctness issue -- it fixes a bug.
2. We're already doing the original tree traversal in `ExtractUsedNames`, this PR only does the equivalent in `Dependency`. E.g. we're not introducing new behaviour.
  
We haven't discussed this publicly, but I think we should not run benchmarks for bugfixes, only for new features and PRs that change the structure of the algorithm (like #267, for instance, that happens to do both).
  
Knowing the benchmark results for every bugfix will take the author of the PR an insanely amount of time and will not give anything to the maintainers: it's a bugfix, they have to merge it in any case.
  
I would be for checking performance results on bugfixes that touch a lot of code and sensible parts of the Scalac phases. But as this PR only propagates a behaviour we are doing in `ExtractUsedNames`, I suggest not to run them.

eed3si9n:
@jvican I created GitHub milestone named 1.0.0-X11 to track the next release - https://github.com/sbt/zinc/milestone/2

jvican:
Thanks @eed3si9n!

dwijnand (approved):

PR #274, opened 2017-03-20
Add pending tests for changing inherited type in sealed trait children.

romanowski:
IncHandler now understands scalac options.

Add pending tests for changing inherited type in sealed trait children.

This was found during #267 and is a problem also there.


jvican (approved):
LGTM!
  
Glad to see this as a pending test as pointed out in #267.
  
Regarding your scalac options fix, can you be more explicit in the description? I had to look into it too deeply, and maybe we can avoid that for the rest of maintainers :smile:.

romanowski:
```Property `scalac.options` in incOptions.properties is now passed as options to scalac (value is splitted using ` +` regexp).``` added to commit message. Is anything more required?

jvican:
Thanks!
  
No, don't think so. Let's wait for another signoff.

jvican:
Ping @eed3si9n | @dwijnand.

dwijnand (approved):

PR #280, opened 2017-04-04
Fix #114: Remove unnecessary use of lazy in classes defined by compiler-interface

### LINKED ISSUE #114: Remove unnecessary use of lazy in classes defined by compiler-interface
In https://github.com/sbt/zinc/blob/1.0/internal/compiler-interface/src/main/contraband/type.json and https://github.com/sbt/zinc/blob/1.0/internal/compiler-interface/src/main/contraband/definition.json many types are declared lazy, but it seems that only `Structure#inherited` really needs to be lazy (because a class might contain itself as an inherited member: `class A { class B extends A }`. I think that `ClassLike#selfType`, `ClassLike#structure`, and `Structure#declared` can be made strict (or at least, I was able to compute them strictly in the dotty compiler bridge and not blow up anything).

ttreyer:
As explained in issue #114, most of the types declared in `type.json` and `definition.json` do not need to be `lazy`.
They have been made strict and a test had to be modified to handle `Seq` results in varying order.

### DIFF
--- a/internal/compiler-interface/src/main/contraband/definition.json
+++ b/internal/compiler-interface/src/main/contraband/definition.json
@@ -101,11 +101,11 @@
           "type": "record",
           "fields": [
             { "name": "definitionType",        "type": "DefinitionType" },
-            { "name": "selfType",              "type": "lazy Type"      },
-            { "name": "structure",             "type": "lazy Structure" },
+            { "name": "selfType",              "type": "Type"           },
+            { "name": "structure",             "type": "Structure"      },
             { "name": "savedAnnotations",      "type": "String*"        },
             { "name": "childrenOfSealedClass", "type": "Type*"          },
-            { "name": "topLevel",              "type": "Boolean"        },
+            { "name": "topLevel",              "type": "boolean"        },
             { "name": "typeParameters",        "type": "TypeParameter*" }
           ]
         }
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -83,11 +83,11 @@ object ClassToAPI {
       val topLevel = c.getEnclosingClass == null
       val name = classCanonicalName(c)
       val tpe = if (Modifier.isInterface(c.getModifiers)) Trait else ClassDef
-      lazy val (static, instance) = structure(c, enclPkg, cmap)
-      val cls = new api.ClassLike(name, acc, mods, annots, tpe, lzyS(Empty), lzy(instance, cmap), emptyStringArray, children.toArray,
+      val (static, instance) = structure(c, enclPkg, cmap)
+      val cls = new api.ClassLike(name, acc, mods, annots, tpe, Empty, instance, emptyStringArray, children.toArray,
         topLevel, typeParameters(typeParameterTypes(c)))
       val clsDef = new api.ClassLikeDef(name, acc, mods, annots, typeParameters(typeParameterTypes(c)), tpe)
-      val stat = new api.ClassLike(name, acc, mods, annots, Module, lzyS(Empty), lzy(static, cmap), emptyStringArray, emptyTypeArray,
+      val stat = new api.ClassLike(name, acc, mods, annots, Module, Empty, static, emptyStringArray, emptyTypeArray,
         topLevel, emptyTypeParameterArray)
       val statDef = new api.ClassLikeDef(name, acc, mods, annots, emptyTypeParameterArray, Module)
       val defs = cls :: stat :: Nil
@@ -120,18 +120,12 @@ object ClassToAPI {
 
   @inline private[this] def lzyS[T <: AnyRef](t: T): xsbti.api.Lazy[T] = SafeLazyProxy.strict(t)
   @inline final def lzy[T <: AnyRef](t: => T): xsbti.api.Lazy[T] = SafeLazyProxy(t)
-  private[this] def lzy[T <: AnyRef](t: => T, cmap: ClassMap): xsbti.api.Lazy[T] = {
-    val s = lzy(t)
-    cmap.lz += s
-    s
-  }
 
   private val emptyStringArray = new Array[String](0)
   private val emptyTypeArray = new Array[xsbti.api.Type](0)
   private val emptyAnnotationArray = new Array[xsbti.api.Annotation](0)
   private val emptyTypeParameterArray = new Array[xsbti.api.TypeParameter](0)
   private val emptyDefArray = new Array[xsbti.api.ClassDefinition](0)
-  private val lzyEmptyTpeArray = lzyS(emptyTypeArray)
   private val lzyEmptyDefArray = lzyS(emptyDefArray)
 
   private def allSuperTypes(t: Type): Seq[Type] =
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -109,8 +109,8 @@ object ClassToAPI {
     if (!Modifier.isPrivate(c.getModifiers))
       cmap.inherited ++= parentJavaTypes.collect { case parent: Class[_] => c -> parent }
     val parentTypes = types(parentJavaTypes)
-    val instanceStructure = new api.Structure(lzyS(parentTypes), lzyS(all.declared.toArray), lzyS(all.inherited.toArray))
-    val staticStructure = new api.Structure(lzyEmptyTpeArray, lzyS(all.staticDeclared.toArray), lzyS(all.staticInherited.toArray))
+    val instanceStructure = new api.Structure(parentTypes, all.declared.toArray, lzyS(all.inherited.toArray))
+    val staticStructure = new api.Structure(emptyTypeArray, all.staticDeclared.toArray, lzyS(all.staticInherited.toArray))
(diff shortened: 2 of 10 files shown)


internal/zinc-compile-core/src/test/scala/sbt/internal/inc/javac/JavaCompilerSpec.scala
```diff
@@ -95,7 +95,7 @@ class JavaCompilerSpec extends UnitSpec {
         val (result, _) = compile(local, Seq(input), Seq("-d", out.getAbsolutePath))
         result shouldBe true
         val clazzz = new URLClassLoader(Array(out.toURI.toURL)).loadClass("hasstaticfinal")
-        ClassToAPI(Seq(clazzz))
+        ClassToAPI(Seq(clazzz)).sortBy(_.definitionType.hashCode)
```
romanowski: Why do we need to sort it now?

internal/compiler-interface/src/main/contraband/definition.json
```diff
@@ -101,11 +101,11 @@
...
           "fields": [
             { "name": "definitionType",        "type": "DefinitionType" },
-            { "name": "selfType",              "type": "lazy Type"      },
-            { "name": "structure",             "type": "lazy Structure" },
+            { "name": "selfType",              "type": "Type"           },
+            { "name": "structure",             "type": "Structure"      },
             { "name": "savedAnnotations",      "type": "String*"        },
             { "name": "childrenOfSealedClass", "type": "Type*"          },
-            { "name": "topLevel",              "type": "Boolean"        },
+            { "name": "topLevel",              "type": "boolean"        },
```
ttreyer, line 108: `Boolean` became lower case, see sbt/contraband#80

internal/zinc-compile-core/src/test/scala/sbt/internal/inc/javac/JavaCompilerSpec.scala
jvican, line 98: I don't understand this, can you elaborate?

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -120,18 +120,13 @@ object ClassToAPI {
 
   @inline private[this] def lzyS[T <: AnyRef](t: T): xsbti.api.Lazy[T] = SafeLazyProxy.strict(t)
   @inline final def lzy[T <: AnyRef](t: => T): xsbti.api.Lazy[T] = SafeLazyProxy(t)
-  private[this] def lzy[T <: AnyRef](t: => T, cmap: ClassMap): xsbti.api.Lazy[T] = {
```
jvican, line 123: Glad to see this going away.

internal/compiler-interface/src/main/contraband/definition.json
jvican: Glad you found that issue, we need to be pretty careful with equality on these fields.

jvican (approved):
Hey @zekhae, great to see you again!
  
I have a small question, otherwise good job :smile:.

jvican:
@romanowski I received some of your replies via email, but they somehow got deleted? Don't be afraid of reviewing stuff, comments and help are very much welcome :wink:.

jvican:
Something is off in Travis land. Will try restarting the Travis CI again tomorrow morning.

romanowski:
@jvican it is super strange since I approved this PR..

I've got question similar to yours (`I don't understand this, can you elaborate?`) but I later on I decided to remove it.

romanowski (approved):
I really like this simplification (it was pain for debugging).

I think we may see some performance gains from it (we don't need to create tons on Lazy objects)

jvican:
Good point, I didn't think this could improve performance. Hope it does!

internal/zinc-compile-core/src/test/scala/sbt/internal/inc/javac/JavaCompilerSpec.scala
ttreyer: Removing laziness from the fields of `ClassLike` has the side effect to randomize the order of the result from `compileWithPrimitive()` method (which returns a `Seq` containing a `class` and an `object`).
Sorting the result ensure we have an order consistent with the expected result of the test.

eed3si9n: > Removing laziness from the fields of ClassLike has the side effect to randomize the order of the result 

Does the result become non-deterministic with this change? If so would that affect change detection in computing (transitive) incremental compilation? iirc, Zinc first compiles whatever that it first detected as source change, and then it subsequently uses API difference to figure out if it needs to do more work.

jvican: > Removing laziness from the fields of ClassLike has the side effect to randomize the order of the result from compileWithPrimitive() method (which returns a Seq containing a class and an object).
  
Can you diagnose why? This is strange.
  
> Sorting the result ensure we have an order consistent with the expected result of the test.
  
In my opinion, we should not sort here just for testing purposes. Our core logic does not require it, it has no effect on Zinc changes analysis and it's not a contract of our API. Can we solve it in the tests side?

romanowski: AFAIK order should not matter in computing APIs and hashes (I am almost sure that we have a tests for that). Maybe we need to call `toSet` in checks later on?

internal/compiler-interface/src/main/contraband/definition.json
jvican: Since you're at this, could you check that `boolean` is used everywhere in the json contraband sources? @zehkae

internal/zinc-compile-core/src/test/scala/sbt/internal/inc/javac/JavaCompilerSpec.scala
ttreyer: `ClassToAPI` uses a `Set` to keep track of the APIs, then returns its result converted into a `Seq`.
Contraband generate the `hashCode()` method in a different way, depending on the class having lazy fields or not. Since all the lazy fields have been removed, it computes the hash by combining the hash of all the fields together instead of just calling `super.hashCode()` (See [JavaCodeGen.scala](https://github.com/sbt/contraband/blob/master/library/src/main/scala/sbt/contraband/JavaCodeGen.scala#L332-L345)). This should explain why we now have a random order.

Since we zip the `leftAPI` and `rightAPI` together, sorting the result seems to be the apropriate way to ensure we call `SameAPI` with the correct pairs.

ttreyer:
I also tried to remove `lazy` for the `inherited` field of `Structure` (in "type.json") and the tests didn't report any error.
Unfortunately this field has to be `lazy` for now and here is a test that shows that. It will fail with a stack overflow if `inherited` is strict.

internal/zinc-compile-core/src/test/scala/sbt/internal/inc/javac/JavaCompilerSpec.scala
jvican: @zehkae Thanks for the explanation!
  
So I have two questions:
  
1. Could this issue be fixed in Contraband?
2. When you mention `leftAPI` and `rightAPI`, do you mean that their order alter the behaviour of the API change detection algorithm?
  
To me, this change seems to be necessary only for the testing suite. I'm slightly reluctant of sorting this only for testing purposes because order should not be a factor that we should take into account if it really doesn't alter semantics :smile:.

eed3si9n (approved):

jvican:
@eed3si9n Why have you merged? I was waiting for some feedback on the `sorted` scenario. I don't agree that change was necessary to this PR.

eed3si9n:
Sorry I'd forgotten that the discussion was still midway.
The sort is done in the `JavaCompilerSpec`, so in my head the PR was ready to merge once the nondetermistic output was considered ok.

jvican:
Let's forget about it, I don't think it's that important either (/cc @zehkae). But next time, please make sure there's consensus before merging.

internal/zinc-compile-core/src/test/scala/sbt/internal/inc/javac/JavaCompilerSpec.scala
ttreyer: 1. There is no issue with Contraband. Changing how `hashCode()` is generated based on the laziness of the fields is intended and prevents infinite recursion (see the test added in 436828c).
  2. The compilation done by `compileWithPrimitive()` can produce a *class* and an *object*, but their order is not defined since they're in a `Set`. So `leftApi` will contain a *class* and an *object* from a first compilation and `rightApi` will contain a *class* and an *object* from a second compilation. Since we want to compare the *class* of `leftApi` with the *class* of `rightApi` (same for the *object*) we need to form the correct pairs for the comparison and not compare a *class* with an *object*. Sorting ensure we have the same order in `leftApi` and in `rightApi`. There may be other ways to ensure we get the correct pairs, the sort could be done elsewhere, but sorting there seemed to be a simple fix for the problem. Finally the sort is done only in the tests, it shouldn't be needed elsewhere.

Sorry for the late response. Even though the PR has been merged, I hope these answers help you understand the necessity of the sort in those tests.

jvican: Yes, they do. Thanks.

PR #287, opened 2017-04-06
Fix #102: Better main class detection

### LINKED ISSUE #102: Simpler main class detection
The main class detection is handled by https://github.com/sbt/zinc/blob/1.0/internal/zinc-apiinfo/src/main/scala/xsbt/api/Discovery.scala which looks for a `main` method with the correct signature in the extracted API. This is imperfect because it relies on `ExtractAPI` dealiasing types which it shouldn't do since aliases can affect type checking (example: https://gist.github.com/smarter/2982126fd17b9b58b820f3b50751a8e5). A simpler solution would be for `ExtractAPI` to directly record which classes have a main method (using [hasJavaMainMethod](https://github.com/scala/scala/blob/93f209dd65d5c05fc2cb61916a850940499c9261/src/reflect/scala/reflect/internal/Definitions.scala#L555-L557)) and return this information in some way.

smarter:
Previously, the main class detection was handled by
https://github.com/sbt/zinc/blob/1.0/internal/zinc-apiinfo/src/main/scala/xsbt/api/Discovery.scala
which looks for a main method with the correct signature in the
extracted API. This is imperfect because it relies on ExtractAPI
dealiasing types (because Discovery will look for a main method with a
parameter type of `java.lang.String` and won't recognize
`scala.Predef.String`), dealiasing means that the extracted API looses
information and thus can lead to undercompilation.

This commit partially fixes this by adding a new callback to AnalysisCallback:
```java
    void mainClass(File sourceFile, String className)
```
that is used to explicitly register main entry points. This way, tools
do not need to interpret the extracted API, this is much better since it
makes it easier for zinc to evolve the API representation.

This commit does not actually changes ExtractAPI to not dealias, this
can be done in a later PR.

Note that there is another usecase for xsbt.api.Discovery that this PR
does not replace: discovering tests. This is more complicated because
different test frameworks have different ways to discover tests. For
more information, grep for "Fingerprint" in https://github.com/sbt/sbt
and https://github.com/sbt/junit-interface

Note also that the added scripted test does not actually test that this
PR is correct since sbt itself needs to be updated to use this new API,
the branch at
https://github.com/smarter/sbt/commits/fix/main-class-detection does
this, but cannot be merged in sbt until a new zinc is published.

Fix #102

### DIFF
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -22,14 +22,16 @@ object ClassToAPI {
   def apply(c: Seq[Class[_]]): Seq[api.ClassLike] = process(c)._1
 
   // (api, public inherited classes)
-  def process(classes: Seq[Class[_]]): (Seq[api.ClassLike], Set[(Class[_], Class[_])]) = {
+  def process(
+      classes: Seq[Class[_]]): (Seq[api.ClassLike], Seq[String], Set[(Class[_], Class[_])]) = {
     val cmap = emptyClassMap
     classes.foreach(toDefinitions(cmap)) // force recording of class definitions
     cmap.lz.foreach(_.get()) // force thunks to ensure all inherited dependencies are recorded
     val classApis = cmap.allNonLocalClasses.toSeq
+    val mainClasses = cmap.mainClasses.toSeq
     val inDeps = cmap.inherited.toSet
     cmap.clear()
-    (classApis, inDeps)
+    (classApis, mainClasses, inDeps)
   }
 
   // Avoiding implicit allocation.
@@ -55,7 +57,8 @@ object ClassToAPI {
       private[sbt] val memo: mutable.Map[String, Seq[api.ClassLikeDef]],
       private[sbt] val inherited: mutable.Set[(Class[_], Class[_])],
       private[sbt] val lz: mutable.Buffer[xsbti.api.Lazy[_]],
-      private[sbt] val allNonLocalClasses: mutable.Set[api.ClassLike]
+      private[sbt] val allNonLocalClasses: mutable.Set[api.ClassLike],
+      private[sbt] val mainClasses: mutable.Set[String]
   ) {
     def clear(): Unit = {
       memo.clear()
@@ -67,6 +70,7 @@ object ClassToAPI {
     new ClassMap(new mutable.HashMap,
                  new mutable.HashSet,
                  new mutable.ListBuffer,
+                 new mutable.HashSet,
                  new mutable.HashSet)
 
   def classCanonicalName(c: Class[_]): String =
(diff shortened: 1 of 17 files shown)


jvican:
> This commit does not actually changes ExtractAPI to not dealias, this
can be done in a later PR.
  
I'm unsure this is possible at all. API detection needs dealiased types to work correctly.
  
@smarter I'll review this one, thanks for the PR.

smarter:
> I'm unsure this is possible at all. API detection needs dealiased types to work correctly.

API detection of what?

jvican:
> API detection.
  
`ExtractAPI.scala`.

smarter:
Yes, but why do you say that ExtractAPI needs it to work correctly? After https://github.com/sbt/zinc/pull/87 and https://github.com/sbt/zinc/pull/239 zinc should handle abstract types in the API correctly, so not expanding type aliases should work too.

smarter:
I fixed the added test to:
- actually work (I forgot that zinc scripted tests cannot define commands in build.sbt anymore, you have to add them to `IncHandler` instead)
- actually test zinc itself, no matter what sbt uses for main class discovery (I added a `checkMainClasses` command to `IncHandler`, that behaves similarly to existing `check*` commands in there and uses the internal API).

smarter:
Ping for review.

zinc/src/sbt-test/apiinfo/main-discovery/test
```diff
@@ -0,0 +1,4 @@
+> compile
```
jvican, line 1: These tests look great!

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -94,6 +96,16 @@ object ClassToAPI {
       val defsEmptyMembers = clsDef :: statDef :: Nil
       cmap.memo(name) = defsEmptyMembers
       cmap.allNonLocalClasses ++= defs
+
```
jvican: 👍

jvican (approved):
LGTM. Excellent job!
  
It's true that we need to put the hooks (or provide the necessary information) to detect main classes for tests. I'll look into the best way to do so.

jvican:
@smarter Sorry, but we merged a formatting PR that invalidates the contents of this one. Can you please update this PR?
  
@eed3si9n @dwijnand  Can you have a quick look to this? I want to merge it. If you don't have time, tell me so and I'll merge it.

eed3si9n (changes requested):
> This is imperfect because it relies on ExtractAPI
dealiasing types (because Discovery will look for a main method with a
parameter type of java.lang.String and won't recognize
scala.Predef.String), dealiasing means that the extracted API looses
information and thus can lead to undercompilation.

Is it possible to fix the root cause rather than adding special case for main class detection?

smarter:
> Is it possible to fix the root cause rather than adding special case for main class detection?

The root cause is the use of Discovery, the fix is to not use Discovery and the same should be done for tests, and Discovery should be removed. There is no way to make Discovery understand all the semantics of Scala type system.

smarter:
Updated, phew! Hopefully this gets merged before the next major reformatting/refactoring of the codebase.

eed3si9n (approved):

PR #288, opened 2017-04-06
Fix #97: Avoid spurious recompilations when unrelated constructor changes

### LINKED ISSUE #97: Avoid spurious recompilations when unrelated constructor changes
`A.scala`:

``` scala
class A(a: Int) // change Int to String, B.scala should not be recompiled
object A { val x = 3 }
```

`B.scala`:

``` scala
class B { val y = A.x }
```

`B` does not use the constructor of `A`, but every constructor has the same name `<init>`, so if we change a constructor in a class, then all dependencies of that class are inevitably recompiled. This should be easy to fix: just special case constructors and prepend the class name to the method name, like `A<init>`, `B<init>`.

(Note that this change will break value-class-underlying if #95 gets merged (see the PR description), this can be fixed by including the underlying val of a value class in its name hash)

smarter:
The name hashing algorithm is designed to take implicit conversions into
account: when a name "foo" is changed somewhere in a dependency of X,
you have to recompile X if it uses the name "foo", even if the usage of
"foo" in X is completely unrelated, just because this might have an
effect on available implicit conversions. However, there is one case
where we can be sure that implicit conversions will not kick in: when we
call a constructor. A constructor name is always "&lt;init&gt;", this PR now
replaces this name by "pkgA;pkgB;className;init;", this mean that we no
longer recompile classes when an unrelated constructor in a used class
changed (see the new test `constructors-unrelated` for an example).


jvican (approved):
LGTM. This is a great change!

Thank you for the docs, too. Let's request another review from the Ligthbend team. In the meanwhile, I'll try to see the interaction of this PR with potential corner cases, just to be sure this is ready to go!

eed3si9n (approved):
The change itself looks good.
I wonder if we should start a markdown file documenting mangling rules.

jvican:
@smarter Give me until next Monday to think about potential interactions, and we merge :wink:. I just want to be sure.

smarter:
Ping for review.

eed3si9n:
I already gave my 👍  so we're waiting on @jvican?

jvican:
Yes @eed3si9n, I told @smarter in person that I'll be reviewing this by the end of the week. Sorry for the delay.

jvican:
This change looks great, I've confirmed this will not have a bad interaction with anything. I was worried about the format representation and wanted to double check that it was unique.
  
I am resolving the conflicts caused by the formatting PR manually, and when the light is green I'm merging.

jvican:
Travis was blocked, that's why I closed and opened again. I'll merge this when it's green.

PR #301, opened 2017-05-17
Fix #114: Remove unnecessary use of lazy in classes defined by compiler-interface

### LINKED ISSUE #114: Remove unnecessary use of lazy in classes defined by compiler-interface

ttreyer:
As stated in sbt/zinc#114, some types don't need to be declared lazy and had been strictly computed in the dotty compiler bridge. This PR is a new attempt at backporting those observations in Zinc itself.

I've removed laziness from the following fields:
  * `ClassLike.selfType`
  * `Structure.declared`
  * `Structure.inherited`

I was not able to remove laziness for `ClassLike.structure`, the tests in `apiinfo/circular-structure` fail when removing it. `Structure.parents` must also stay lazy.
We should come up with new tests for `Structure.declared` and `Structure.inherited` to prevent any regression, especially for `Structure.inherited` which was not strictly computed in the dotty bridge.


jvican:
Hey @zehkae, thanks for following up on this one.
  
Right now, we're preparing for a major milestone of Zinc and sbt and we're a little bit swamped in work. However, I plan to come back to this PR when that's over and study carefully your changes to ensure that the previous regression doesn't happen again, this lazy construct is quite difficult to predict!
  
> We should come up with new tests for Structure.declared and Structure.inherited to prevent any regression, especially for Structure.inherited which was not strictly computed in the dotty bridge.
  
Could you come up with such tests to ensure that these changes are safe?

jvican:
ping @zehkae

typesafe-tools:
To validate this pull request against other sbt modules, visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/301/head&deploy_to_bintray=false">THIS LINK.</A>

ttreyer:
I've added a test for circular references in both Scala and Java.
It passes without issue, so that's a good sign. But still, please test the patch with a real-world project.
Typically sbt itself since it is already known for triggering regressions in the previous patch.

jvican:
@zehkae Thanks for following through this. This patch is a little bit dangerous. I'm afraid of merging this and have a bazillion of regressions, so I'm closing this. After Zinc 1.0, I'll reopen and consider merging it.
  
Thanks for your work!

Issue #302, opened 2017-05-17
Replace all instances of `Maybe` by `Optional`


jvican:
We have already had chats over the requirement of JDK8 to use Zinc 1.0. This has been discussed in the Gitter channels and other mediums, and the general feeling is that it's a reasonable thing to do, and Pants has confirmed this would not be a blocker.
  
Therefore, **before we release RC1 for Zinc**, we should remove `Maybe` and replace it with `java.util.Optional`.

Issue #286, opened 2017-04-06
Reorder subproject dependencies


jvican:
The current state of the subproject dependencies affect us in two ways:
  
1. CI is slow because we cannot easily parallelize compilation and test execution.
2. Cross-compilation of projects is inconsistent.
  
```
[zinc Root@1.0]> crossScalaVersions
[info] zinc/*:crossScalaVersions
[info]  List(2.11.8, 2.12.1)
[info] zincTesting/*:crossScalaVersions
[info]  List(2.11.8, 2.12.1)
[info] zincBenchmarks/*:crossScalaVersions
[info]  List(2.11.8, 2.12.1)
[info] zincCore/*:crossScalaVersions
[info]  List(2.11.8, 2.12.1)
[info] zincCompileCore/*:crossScalaVersions
[info]  List(2.11.8, 2.12.1)
[info] compilerBridge/*:crossScalaVersions
[info]  List(2.12.1, 2.11.8, 2.10.6)
[info] zincClassfile/*:crossScalaVersions
[info]  List(2.12.1, 2.11.8, 2.10.6)
[info] zincPersist/*:crossScalaVersions
[info]  List(2.11.8, 2.12.1)
[info] zincClasspath/*:crossScalaVersions
[info]  List(2.12.1, 2.11.8, 2.10.6)
[info] compilerInterface/*:crossScalaVersions
[info]  List(2.12.1)
[info] zincCompile/*:crossScalaVersions
[info]  List(2.11.8, 2.12.1)
[info] zincScripted/*:crossScalaVersions
[info]  List(2.11.8, 2.12.1)
[info] zincIvyIntegration/*:crossScalaVersions
[info]  List(2.11.8, 2.12.1)
[info] zincApiInfo/*:crossScalaVersions
[info]  List(2.12.1, 2.11.8, 2.10.6)
[info] zincRoot/*:crossScalaVersions
[info]  List(2.11.8, 2.12.1)
```
  
@dwijnand has provided this graph:
![projects-graph](https://cloud.githubusercontent.com/assets/2462974/24748244/6925b260-1abf-11e7-8700-af6af96000e1.png)

  
In it, we can see that the compiler bridge tests depend on `zincApiInfo` and `zincClassfile`, and that's the reason why those projects have to cross-compile to 2.10.
  
A way to solve this issue would be to disentagle the dependencies between the projects by moving the necessary bits to independent projects or even to the dependent projects, depending on the nature of the code.
  
This issue was pointed out by @smarter.

[...]

Issue #268, opened 2017-03-16
Use protobuf for serialization/deserialization


jvican:
Serialization and deserialization of Analysis files is important for:
  
* Performance. The faster, the better.
* Disk usage.
* Correctness. We want serialised formats to be read by Zinc even if they were done years and years ago. This is super important if we're going to ship analysis files and cache incremental compilation, a project @romanowski is working on in [Hoarder](https://github.com/romanowski/hoarder).
  
The serialisation that we're using now is far than ideal: it's relatively slow, it consumes a lot of space and it does not ensure binary compatibility of serialised files.
  
Because of that, I'd like to propose to use Protobuf. It addresses the three problems at the same time:
  
* It's [the fastest serialisation library out there in the Scala community](https://medium.com/@dkomanov/scala-serialization-419d175c888a#.exqx4eeyf).
* Disk usage is diminished because it outputs binaries. These binaries are small so we can then decide whether we can avoid the compressing/decompressing step based on the resulting sizes, which may bring is some speedup.
* Binary compatibility is taken seriously. The whole protocol is architected around it, so this has the plus benefit that we won't ever break the binaries if serious changes to the API happen.

romanowski:
Our Analysis holds 2 things:
* incremental compiler information
* apis

Latter is mostly used in SBT and I feel that those APIs soon become obsolete (can be replaced with e.g. scala.meta, but I don't want to go into details here).

So we should mostly focus on first part. IMO we need to first migrate parts that use java serialization 
(most can be found [here](https://github.com/sbt/zinc/blob/1.0/internal/zinc-persist/src/main/scala/sbt/internal/inc/TextAnalysisFormat.scala#L34-L48), excluding sbinary ones).

The only concern I have about Protobufs is that debugging may be harder then zipped text files but a) this not apply for thing that are serialized using sbinary/java serialization b) this is only zinc developers (and there is probably a few who reads Analysis manually).

jvican:
No matter how Analysis files are used, they should stay for now always binary compatible. We cannot use the semantic database now, so it's better not to talk about that since it's a decision that needs to be taken at least one year in the future when everything stabilises.
  
> The only concern I have about Protobufs is that debugging may be harder then zipped text files but a) this not apply for thing that are serialized using sbinary/java serialization b) this is only zinc developers (and there is probably a few who reads Analysis manually).
  
[...]

Issue #218, opened 2017-02-04
Make Analysis content machine-independent


eed3si9n:
There are number of items in Analysis currently that's machine dependent.
We could for example, keep the machine dependent parts such as base directory in-memory, and store relative paths in Analysis. Similar treatment could be done for library JARs, Java Home etc.

Making Analysis content machine-independent allows cached compilation.

jvican:
This is related to #214.
  
I agree this is a nice have. Though I'm not sure we should push for it for the next stable release since it does not affect key functionality in Scala devs workflow. I would suggest marking it as "contributions welcome" or "community" so that people like @gabro could have a stab at it. WDYT @eed3si9n ?

gabro:
@jvican I think it's a good idea, although it'd be great if you guys should provide some extra context/analysis, so that a person not familiar with the code base (like myself) could pick this up.
For instance, do you already have a list of the machine-dependent items?

jvican:
Only paths that I know of. The idea described by @eed3si9n is to have an internal representation that is absolute but only keep relative parts in the serialized analysis instances (and therefore API). An internal conversion relative -> absolute -> relative would need to be performed (the first one to read analysis and the second one to write it).
  
Classpaths would need special handling (we need to be careful, some of the paths in classpaths are absolute and therefore machine-dependent). Those should be left unchanged. We should only change paths that affect user-defined class files or jars.

eed3si9n:
Lightbend and @romanowski (VirtusLab) has spent some time thinking about this problem, so I think it's something one of us should tackle. Analysis structure is among the most critical aspect of Zinc to get right.

jvican:
Good @eed3si9n, that makes sense! I'll focus for now on the dependency analysis, but if I have time left I'll try to get it out. Though I would prefer if someone else (perhaps @romanowski) is interested in tackling this soon.

romanowski:
First of all we need to make both *Analysis* and *MInisetup* machine independent. 
Zinc (in current shape) is IMO way too low level to handle relative classpath or even relative source roots.

Classpath is good example here since build tool (e.g. sbt) knows what makes given classpath entry. We can map e.g. jar back to its declaration (so e.g. /my/absolute/path/lib.jar becomes ogranization%name%version%jar that we can share across machines).

However zinc can allows clients to make analysis machine independent. All what is required to achieve that is already implemented in #216 as mappers (so client can map part of Analysis based of its setup).

[...]

PR #304, opened 2017-05-22
Add the first iteration of the Zinc API 1.0

jvican:
This is just the first iteration of the Zinc API.
  
This iteration's goal is to make safe functionality of Zinc public to allow Scala and Java parties (Gradle, Pants, Maven, et cetera) use Zinc with binary compatibility guarantees. This task needs another iteration given:
  
1. The big surface of the API; and,
2. The fact that the whole Zinc API was made private for 1.0 (!).

The last iteration will happen before RC1 (end of June - beginning of July).
  
The new public API makes it easier to:
  
* Get a compiled bridge for a given scala instance (via ivy).
* Get a scala instance (via ivy).
* Instantiate a Zinc compiler.
* Provide utilities for both Java and Scala users to get default implementations that do not expose the internals, placed in `sbt.internal.inc`.
  
Aside from these, this API also addresses some issues that have arisen in the past:
  
* It makes it easier to customize the classfile manager, as required by Scalajs and Scala Native. https://github.com/sbt/zinc/issues/125
* It replaces any instance of Maybe with Optional. https://github.com/sbt/zinc/issues/302
 
I find that in general, these commits make Zinc easier to use and improve the consistency of the whole API. The most remarkable change is the `CompileAnalysis` API, which now provides useful information for build tools like file stamps, compilations and source infos.
  
Note that all the utilities made public only expose functionality that is a product of the incremental compiler. No internal details of the incremental analysis are exposed to the users.
  
There are some Java files that have been moved out of the compiler interface because they required static method implementations in the projects on which the implementations of certain classes were provided.
  
In the future, I feel that these are the things that should be worked on to finish off the Zinc API:
  
1. Reorder project structure. Simplify it and make it more granular and easier to test. https://github.com/sbt/zinc/issues/286
2. Decide what and how to use to persist analysis files with appropriate benchmarking. https://github.com/sbt/zinc/issues/268 (protobuf is not possible, so we need to find a way to either use sbinary or a similar solution that does not harm (or even improve!) speed of compiler iterations).
3. Create utilities to instantiate:
  3.1. Empty `Analysis`.
  3.2. `DefinesClass`.
  3.3. `AnalysisStore` (up to discussion for next iteration).
  3.4. `JavaTool` et al, necessary to create `Compilers`.
  3.5. `PerClasspathLookupEntry`.
4. Make analysis content machine-independent: #218
  
I have more ideas, but those will be discussed when time's up.

The breaking changes are documented in every commit. I suggest that reviewers go through the commits one by one, though I don't expect them to do so (given the big diff size). As this is just the first iteration of the redesign and targets a milestone, I am not writing any migration notes for now. I'll wait until everything is finished in our first RC to write the migration notes so that other build tools (other than sbt) can start using it.
  
As a reminder, the public API is `xsbti`. Scala implementations live in `sbt.internal.inc`.
  
This is a good time to gather feedback. So that's very much welcome.


eed3si9n:
As a test, could we put `zinc/src/test/scala/sbt/inc/*Spec.scala` into another package name like `com.example`, and see if it's able to initialize everything without invoking the internals?

While I agree with getting rid of `Maybe` and using `Optional` it's a bit distracting to review it because it spans across everywhere. How difficult do you think it would be to split Maybe change into a another PR that precedes this?

internal/zinc-scripted/src/main/contraband-scala/sbt/internal/inctest/Build.scala
```diff
@@ -1,32 +0,0 @@
-/**
- * This code is generated using [[http://www.scala-sbt.org/contraband/ sbt-contraband]].
- */
-
-// DO NOT EDIT MANUALLY
-package sbt.internal.inctest
-final class Build private (
```
eed3si9n: What happened to my build.json?

internal/compiler-interface/src/main/java/xsbti/compile/CompilerBridgeProvider.java
```diff
@@ -0,0 +1,20 @@
...
+public interface CompilerBridgeProvider {
+
+    /**
+     * Get the location of the Scala bridge sources.
+     *
+     * @param scalaInstance The Scala instance for which the bridge should be retrieved for.
+     * @param logger The logger.
+     * @return The jar (more typically) or directory where the sources are stored.
+     */
+    File getBridgeSources(ScalaInstance scalaInstance, Logger logger);
```
eed3si9n: Getter shouldn't have side effects.

internal/zinc-ivy-integration/src/main/scala/sbt/internal/inc/IfMissing.scala
```diff
@@ -0,0 +1,14 @@
+package sbt.internal.inc
+
+sealed trait IfMissing
```
eed3si9n, line 10: Not sure why this needs to be on its own file.

internal/compiler-interface/src/main/java/xsbti/compile/CompilerBridgeProvider.java
```diff
@@ -22,7 +22,7 @@
     static CompilerBridgeProvider constant(File file) {
         return new CompilerBridgeProvider() {
             @Override
-            public File getBridgeSources(ScalaInstance scalaInstance, Logger logger) {
+            public File getCompiledBridge(ScalaInstance scalaInstance, Logger logger) {
```
eed3si9n: Same as before. Getter shouldn't have side effects.

eed3si9n (changes requested):
Some minor comments.

internal/zinc-ivy-integration/src/main/scala/sbt/internal/inc/IfMissing.scala
jvican: We may want to explore to make this stable in the next iteration. That's why I split it up, there was too much logic in its place of origin.

internal/compiler-interface/src/main/java/xsbti/compile/CompilerBridgeProvider.java
jvican: I'll call it `fetchBridgeSources`.

jvican:
> As a test, could we put zinc/src/test/scala/sbt/inc/*Spec.scala into another package name like com.example, and see if it's able to initialize everything without invoking the internals?
  
This is a good idea. I'ld like to do this in the second iteration, since there are still some things that must be got from `internal`. Those I specify in the PR description, but I may be missing more.
  
> While I agree with getting rid of Maybe and using Optional it's a bit distracting to review it because it spans across everywhere. How difficult do you think it would be to split Maybe change into a another PR that precedes this?
  
The diff for the maybe migration is small, see https://github.com/sbt/zinc/pull/304/commits/686b6bb174fe76390646a7e7770914906fefc2f1. For time-consuming reasons, I would prefer to get it merged in this PR.

internal/zinc-scripted/src/main/contraband-scala/sbt/internal/inctest/Build.scala
jvican: This has been recovered.

jvican:
Feedback has been addressed @eed3si9n.

jvican:
Tests are green.

eed3si9n (approved):

eed3si9n:
Thanks for all the work on this.

stuhood:
With regard to additional private/internal APIs that are currently in use, can browse through https://github.com/pantsbuild/pants/tree/master/src/scala/org/pantsbuild/zinc (as of zinc `1.0.0-X7`) ... I'm currently working on updating to include this patch.

jvican:
@stuhood I've been browsing through that for a while, but thanks for the heads up. Are you in a hurry to do this? I'll be tackling the second iteration of the API soon before the release candidate and the API freeze, so maybe you want to wait until that's out?

eed3si9n:
The change I made to sbt is https://github.com/sbt/sbt/pull/3228

stuhood:
@jvican : It would probably make sense for me to continue to attempt this, just in order to find any additional issues that you might not have seen. As an example, yesterday I noticed that it's not currently possible to extend `LoggerReporter` in a useful way in `X16`, because various methods were made private. So https://github.com/pantsbuild/pants/blob/master/src/scala/org/pantsbuild/zinc/logging/Reporters.scala#L28 is no longer possible. A theoretical `FilteredReporter` wrapper might make more sense anyway though.

jvican:
> So https://github.com/pantsbuild/pants/blob/master/src/scala/org/pantsbuild/zinc/logging/Reporters.scala#L28 is no longer possible. A theoretical FilteredReporter wrapper might make more sense anyway though.
  
Great, I'll make this (or something similar, I have to look into the details) happen to unblock that implementation.

stuhood:
@jvican: A few more details here:
1. the Reporter API seems to require a `ManagedLogger`, but it's not clear how (or whether it would be a good idea) to create one of those, versus passing in an `xsbti.Logger` or `sbt.util.Logger`
2. Because a `Reporter` is a required parameter for `IncrementalCompilerImpl.compile` (and there doesn't appear to be a way to create a `LoggerReporter` without a `ManagerLogger`), it does not currently seem to possible to compile using the 18 arg form.

I'm going to try porting to usage of `xsbti.compile.Inputs` instead.

stuhood:
Alright, I ported to `xsbti.compile.Inputs`, and it cleared things up considerably. But the above issue with `LoggerReporter` and `ManagedLogger` is still an issue.

Also, it doesn't look like it's possible to use a "default" Reporter (as it is with `Progress`, which is `Optional`)... it's conceivable that at some point in the future we'll move the log filtering out of the Reporter, at which point being able to use a default by specifying `None` would be good.

jvican:
@stuhood: Yes, it seems that the entanglement between `LoggerReporter` and `ManagedLogger` is due to sbt. It's an API where sbt's design decisions leak too much.
  
To overcome this issue, some of Zinc's test suites create a `ConsoleReporter` that is a direct subclass of `Reporter` and just `println`s, but that's far from desirable. I think I am going to create a `ManagedLogger` for Zinc and expose it somewhere, or at least make it trivial to create a Zinc reporter. I am working on it as we speak.
  
With regards to making the reporter `Optional`, I have my doubts this is the best choice. It's such a key component of the compilation that I would prefer that users explicitly pass it. Also, depending on the use case, it's likely to be changed. What is a safe assumption for the user? Will Zinc output to my console? Since it's not obvious, I believe users may be better off being more explicit about what they want. I guess that you need for this would disappear if we add methods to instantiate zinc reporters that you can pass in. WDYT?

jvican:
For the record, if you want to move forward and need a `ManagedLogger`, you can do `sbt.util.LoggerExchange.logger(...)`.

stuhood:
> What is a safe assumption for the user? Will Zinc output to my console? Since it's not obvious, I believe users may be better off being more explicit about what they want. I guess that you need for this would disappear if we add methods to instantiate zinc reporters that you can pass in. WDYT?

Yea, that makes sense. I'm still thinking of zinc as a CLI tool, but in the context of a library, being explicit about output location is important... ditto logging.

jvican:
Hey @stuhood, I've been fixing all the issues you've reported regarding logging. I think I have come up with some elegant solutions you may want to have a look at.

This is the most important change: https://github.com/scalacenter/zinc/commit/a9a9c2faf3b511c0f95b19fab87eea2e67346408.
  
All the changes are in this branch: https://github.com/sbt/zinc/compare/1.0...scalacenter:fix-loggers.
  
I can release these changes as part of a temporary `scalacenter/zinc` release to unblock you and get more feedback. These changes will need, of course, to be merged upstream in sbt/zinc.
  
In the new interface, I have added a filtered reporter in Zinc and the normal `LoggerReporter` (now renamed to `LoggedReporter`, see https://github.com/scalacenter/zinc/commit/cac014918ec6d4ff20b257d8363256739d5645e1). `LoggedReporter` now accepts `xsbti.Logger`, meaning that you can pass in whatever logging mechanism you want so long as it extends that interface. These reporter implementations are internal, but you can get a reporter backed up by a logger via the public `ReporterUtil`: https://github.com/scalacenter/zinc/commit/60820b18b7d56f0b7c9e9a9ae267c2e5bc50b59c.
  
Let me know if this public API works for you.

stuhood:
Thanks @jvican . The API looks great (with one exception mentioned below): thanks for incorporating file/msg based filtering.

I should be able to sanity check incremental compile before the end of Monday.

There is a bit of inconsistency between `java.util.function.Function` and `xsbti.F1` for positionMappers in `ReporterUtil` and `CompileOptions`. pants currently only requires JDK7, but I'm floating the idea of switching to JDK8 now... `Function` should be fine, but consistency would be good.

jvican:
@stuhood Happy that you mention that, I opened a PR in sbt/util to address that issue three days ago. The idea is to remember any trace of the old Java-friendly classes for bincompat (`F0`, `F1` and `Maybe`) and replace them by the default Java ones (`Supplier`, `Function` and `Optional`).
https://github.com/sbt/util/pull/84

stuhood:
The new API looks good from a logging/usability perspective (other than #317). I have things working well enough to have reproduced https://github.com/pantsbuild/pants/issues/4477, so I will begin debugging that tomorrow.

stuhood:
@jvican : If you have a PR for the `Reporters`/`Logger` change, I'd shipit. I'm getting close to needing an `X17` release to move forward.

jvican:
Shipping it tomorrow morning. It's ready but kept it local because I'm changing more things in the API.

PR #315, opened 2017-06-13
Speed up CI by 7x

jvican:
Recycle the compiler bridge and the scala instance (with its parent
classloader) for every scripted test that has the same options (the
scala version). This way, we avoid resolving the bridge and classloading
it every time per scripted test.


jvican:
https://github.com/sbt/zinc/pull/315/commits/c2cbf0da36eaf2f622269a5cdc0b5a664f862367 alone makes the execution of scripted more than twice faster.
  
Before: `[error] Total time: 533 s, completed Jun 13, 2017 2:21:06 PM`.
After: `[error] Total time: 237 s, completed Jun 13, 2017 2:09:28 PM`.

jvican:
This is so frustrating. For some reason, Travis is not running the tests...

jvican:
**Summary of changes**:

The following commits make the CI 7x faster than before. This is achieved
through two main improvements:

* Scripted tests are not resolving the compiler and class-loading all the jars
per test (we only do it once per `scripted` invocation).
* It uses Drone instead of Travis, which does not impose any constrain on
resources. Tests are run in a machine with 48 cores and 228GB of memory.
  
Parallelization at the build level has been removed because modules are too
intertwined with each other, and would require compilation of almost all the
modules before proceeding to test things. Therefore, .drone.yml runs everything
together.

zinc/src/sbt-test/source-dependencies/compactify/incOptions.properties
```diff
@@ -0,0 +1 @@
+scalac.options = -Xmax-classfile-name 240
```
jvican, line 1: This is necessary to work around an issue with large file paths in Docker AUFS, limit is 240 bytes. It only happens with 2.11.x because the compactify test makes an intensive use of synthesized classes with huge names.

jvican:
Closing and opening again to update status of PR, that should be passing.

eed3si9n (approved):

bin/run-ci.sh
```diff
@@ -0,0 +1,11 @@
...
+set -eu
+set -o nounset
+SCALA_VERSION="$1"
+
+sbt -Dfile.encoding=UTF-8 \
+  -J-XX:ReservedCodeCacheSize=256M \
+  -J-Xmx3046M -J-Xms3046M -J-server \
+  zincRoot/test:compile scalafmtCheck \
+  "publishBridgesAndSet $SCALA_VERSION" \
+  crossTestBridges zincRoot/test zincRoot/scripted
```
Duhemm: Could you put each distinct command on a separate line? It makes it easier to read and less scary to configure the CI.

internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
```diff
@@ -41,94 +41,83 @@ import scala.collection.mutable
...
-        val p = buildStructure("root")
-        onIncInstance(i, p) { x: IncInstance =>
-          // This does not do aggregation.
-          commands(cmd)(p, arguments, x)
-        }
-      case _ =>
-        sys.error(s"$command")
+  override def apply(command: String,
+                     arguments: List[String],
+                     state: Option[IncInstance]): Option[IncInstance] = {
```
Duhemm: `state: State` and returns a `State` too

Duhemm (changes requested):
2 small details, otherwise LGTM! Thanks for making CI great again 👍🏻

internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
jvican: It's true, using `State` here makes it waaaay clearer.

bin/run-ci.sh
jvican: Sure.

Duhemm (approved):
LGTM!

jvican:
Merging, travis hook has been disabled.

PR #321, opened 2017-06-26
In-source the modules of sbt/util required in sbt/zinc

dwijnand:
Run git filter-branch as such:

    cd /s
    git clone /d/sbt-util /d/sbt-util-4-lm
    git filter-branch -f --index-filter \
      'git rm --cached -qr -- . && git reset -q $GIT_COMMIT -- util/control internal/util-control util/relation internal/util-relation cache/tracking internal/util-tracking internal/util-scripted' \
      --prune-empty
    git reset --hard
    git gc --aggressive
    git prune

Then that branch was merged into the 'eat-util' branch here, the build was
configured, and the source auto-formatted.


jvican:
What's the reason for this change?

dwijnand:
To un-split the repositories. These are the subset of modules in sbt/util that are required by sbt/zinc but not by sbt/lm.

jvican:
Yes, I've just understood it after I saw the other PRs. My only question is: would it make sense to change the name of the logging facilities from `lmLogging` to `utilLogging`? The logging API on which the compiler bridge depends comes from there, so I think a more generic name can help avoid confusion: "Why is the bridge depending on logging facilities from lm?".
  
Happy to LGTM after build passes.

dwijnand:
We can change it back, sure. The reason I changed it like so is for a few reasons:

* make it obvious that the logging capabilities come from a module in sbt/librarymanagement
* make it obvious that that module is versioned using lmVersion (utilVersion is gone).

jvican:
Thanks for the explanation. I'm happy to leave it as `lmLogging` if it's too much trouble *and* the librarymanagement does indeed depend on the Java logging interfaces (it's not clear to me this is the case, we may be able to move them to Zinc).
  
The only thing I propose is to explain the reasons for this in both code and the CONTRIBUTING guide just to avoid confusion.

dwijnand:
See https://github.com/sbt/util/issues/58#issuecomment-312242348.

Issue #285, opened 2017-04-05
Port API diff utils to Zinc to make it work again


smarter:
http://www.scala-sbt.org/0.13/docs/Understanding-Recompilation.html#Debugging+an+interface+representation explains how to get API diffs to debug incremental compilation, but that does not seem to work with zinc currently:
1. Enable debug info for scripted tests: https://github.com/sbt/zinc/issues/283
2. Add/edit the `incOptions.properties` file inside a scripted test directory (https://github.com/sbt/zinc/issues/284), for example `zinc/src/sbt-test/source-dependencies/as-seen-from-a/` to contain:
```scala
apiDebug = true
```
3. `wget https://repo1.maven.org/maven2/com/googlecode/java-diff-utils/diffutils/1.3.0/diffutils-1.3.0.jar`
4. Run `sbt -Dsbt.extraClasspath=diffutils-1.3.0.jar`
5. Run `> scripted source-dependencies/as-seen-from-a`
6. Get:
```scala
[error] You have api debugging enabled but DiffUtils library cannot be found on sbt's classpath (out-1)
```
7. Be very sad. Consider all the things you could do in life instead of trying to debug sbt. Go outside.

smarter:
By the way maybe instead of this insane roundabout way of doing diffs you could consider doing something natively in Scala, feel free to steal https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/util/DiffUtil.scala

jvican:
Hey @smarter, thanks for reporting and the way to reproduce it!
  
I would very much prefer that. We could also reuse some of Nicola's work on that front, I remember his coloured diff was super useful.

smarter:
Yes, Nicolas' stuff is in https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/util/DiffUtil.scala, we also adapted it recently to do line diffs so that we can use it when comparing a run test actual output with expected output.

jvican:
The goal of this issue is to port [DiffUtil](https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/util/DiffUtil.scala) to Zinc, replacing the current diff infrastructure. The current infrastructure is in the `zincCore` project: https://github.com/sbt/zinc/blob/1.0/internal/zinc-core/src/main/scala/sbt/internal/inc/APIDiff.scala.
  
Why is this ticket useful?
  
[...]

PR #346, opened 2017-07-07
ApiDiff implementation not relying on diffutils.

Krever:
This solves #285. 

I have copied the Dotty implementation directly and put it as `private object` inside existing `APIDiff` , so the namespace/api is not polluted. Let me now if you find some other approach more convenient(e.g. putting code directly or removing unused parts).

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/APIDiff.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/APIDiff.scala
@@ -18,35 +18,8 @@ import xsbti.api.Companions
 /**
  * A class which computes diffs (unified diffs) between two textual representations of an API.
  *
- * Internally, it uses java-diff-utils library but it calls it through reflection so there's
- * no hard dependency on java-diff-utils.
- *
- * The reflective lookup of java-diff-utils library is performed in the constructor. Exceptions
- * thrown by reflection are passed as-is to the caller of the constructor.
- *
- * @throws ClassNotFoundException if difflib.DiffUtils class cannot be located
- * @throws LinkageError
- * @throws ExceptionInInitializerError
  */
 private[inc] class APIDiff {
-
-  import APIDiff._
-
-  private val diffUtilsClass = Class.forName(diffUtilsClassName)
-  // method signature: diff(List<?>, List<?>)
-  private val diffMethod: Method =
-    diffUtilsClass.getMethod(diffMethodName, classOf[JList[_]], classOf[JList[_]])
-
-  private val generateUnifiedDiffMethod: Method = {
-    val patchClass = Class.forName(patchClassName)
-    // method signature: generateUnifiedDiff(String, String, List<String>, Patch, int)
-    diffUtilsClass.getMethod(generateUnifiedDiffMethodName,
-                             classOf[String],
-                             classOf[String],
-                             classOf[JList[String]],
-                             patchClass,
-                             classOf[Int])
-  }
 
   /**
    * Generates an unified diff between textual representations of `api1` and `api2`.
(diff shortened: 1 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/346/head&deploy_to_bintray=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/APIDiff.scala
```diff
@@ -57,29 +30,202 @@ private[inc] class APIDiff {
...
-    val diff /*: Patch*/ = diffMethod.invoke(null, f1Lines, f2Lines)
-    val unifiedPatch: JList[String] = generateUnifiedDiffMethod
-      .invoke(null, fileName, fileName, f1Lines, diff, (contextSize: java.lang.Integer))
-      .asInstanceOf[JList[String]]
-    unifiedPatch.asScala.mkString("\n")
-  }
+  /**
+   * This class was directly copied from Dotty [[https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/util/DiffUtil.scala]]
+   */
+  private object DiffUtil {
```
jvican: FTR: This implementation does not depend on java-diff-utils, that's why it can be removed.

jvican (dismissed):
LGTM, and thank you again! This is a really useful contribution (/cc @smarter that may be interested in the fix).
  
I'm really excited about this change, it does really improve the way people will figure out what changes triggered Zinc compiles.

Krever:
The output that was produced by `scripted source-dependencies/as-seen-from-a`:

snapshot1

Duhemm (changes requested):
Thanks again @Krever!

The changes LGTM, I would just like to get confirmation from someone who's more knowledgeable than me about licensing to confirm that this PR is okay as it is.

My understanding is that we should put the actual copyright notice from dotty before the code that we borrowed?

So I'm requesting changes until someone can confirm that it's okay...

jvican:
I think crediting is enough. The compiler reporter we have is also copy pasted from Scalac, and has the same license as Dotty.

I think this is completely safe to merge because Zinc used the Scala license (BSD 3-way license). That's why we don't have to add a new license to the repo.

Duhemm (dismissed):

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/346/head&deploy_to_bintray=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/APIDiff.scala
```diff
@@ -57,29 +30,202 @@ private[inc] class APIDiff {
...
-    val f2Lines = asJavaList(f2.lines)
-    //val diff = DiffUtils.diff(f1Lines, f2Lines)
-    val diff /*: Patch*/ = diffMethod.invoke(null, f1Lines, f2Lines)
-    val unifiedPatch: JList[String] = generateUnifiedDiffMethod
-      .invoke(null, fileName, fileName, f1Lines, diff, (contextSize: java.lang.Integer))
-      .asInstanceOf[JList[String]]
-    unifiedPatch.asScala.mkString("\n")
-  }
+  /**
+   * This class was directly copied from Dotty [[https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/util/DiffUtil.scala]]
```
dwijnand: Hey @Krever,

Could you please copy Dotty's copyright notice here? It's a condition of BSD-3 source code redistribution.

Also could you use a more stable identifier than `master`. Looks like Dotty only has 1 tag, so perhaps:

    https://github.com/lampepfl/dotty/blob/0.1.2-RC1/compiler/src/dotty/tools/dotc/util/DiffUtil.scala

Thanks!

dwijnand (changes requested):

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/346/head&deploy_to_bintray=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/APIDiff.scala
Krever: Done! :)

internal/zinc-core/src/main/scala/sbt/internal/inc/APIDiff.scala
```diff
@@ -57,29 +30,211 @@ private[inc] class APIDiff {
...
+    * 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+    * 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
+    * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+   */
```
Duhemm: That's nitpicking, but could you please format this comment so that lines are 100 columns long? (don't worry about the link if it's too long)

Duhemm (changes requested):
Just a really minor comment. Thanks for the follow up on the PR!

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/346/head&deploy_to_bintray=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/APIDiff.scala
Krever: Not a problem at all :) I just assumed no one will really read it anyway. Fixed now.

Duhemm (approved):

jvican (approved):
And... Merging! :tada:

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/346/head&deploy_to_bintray=false">this link.</A>

jvican (approved):

PR #350, opened 2017-07-10
Add protobuf-based analysis binary store

jvican:
This pull request replaces the text-based analysis format by a binary format
powered by Protobuf. The new format provides us with two main advantages:

1. Backwards and forwards binary compatibility at the analysis format level.
2. Faster serialization/deserialization of the analysis file.
3. Provides a much better way to make the analysis file machine-independent.

The purpose of this pull request was to add the benefits of a
backwards/forwards binary compatible analysis file without impacting the speed
of the text-based analysis format. In my local experiments, this goal has been
met. The new format is on par performance-wise with the text-based analysis
format for small analysis files, and it's much faster (~1.5-2x) when increasing
the Scalacheck generators (`maxSources` and `maxRelatives`) from 10 to 40, and
even more if the number increases.

With regards to the public API, the PR defines the public API in zinc persist,
which was all internal. This API is required by every piece of software that
integrates with Zinc and needs to modify the read/write steps in a sane way.

The API adds the following interfaces:

1. `ReadMappers`. To allow customisation before reading the analysis files.
2. `WriteMappers`. To allow customisation before writing the analysis files.
3. `ReadWriteMappers`. To wrap the read and write mappers.
4. `AnalysisStore`. To allow users to implement their own way to back analysis
   files.
5. `AnalysisContents`. To allow API evolution of the return type of `set` in
   `AnalysisStore`. It replaces the previous tuple.

These APIs all provide ways to get the default implementations from Java.

The most relevant implementations to discuss are the ones for `ReadMappers` and
`WriteMappers`. These implementations expose a way to relativize (make
machine-independent) an analysis file. Why would someone want to do that? To
ship analysis files to clusters and associate them with commits to avoid first
compilations when checking out a concrete commit.

These implementations, however, are not meant to cover all the possible use
cases for all build tools, they only provide a simple implementation for those
users that can ensure that all paths in the analysis file start with a concrete
prefix. This is usually not the case (read docs in the interfaces), and
therefore build tools will need to implement their own, taking into account all
the paths that can be present: ivy cache, coursier cache, binary dirs, product
dirs, et cetera. This approach was taken because it's impossible to produce an
elegant implementation that satisfies all build tools' requirements.

I've shipped today a release that @stuhood is using to integrate with these
changes and the ones I made to the loggers interface. It is accessible
[here](https://bintray.com/scalacenter/releases/zinc/1.0.0-X19-M1).

The implementation has been done iteratively so that every commit compiles,
making the final switch in the last commits.

I recommend that the people with most experience in this part of the codebase
(@eed3si9n and @Duhemm) review this PR. Also a ping for @stuhood. @dwijnand
Please feel free to have a view on it, too, if you'd like.

The way I recommend to review this PR is to have a look at the resulting diff.
It may look like huge, but it's actually conceptually really simple:

* Changes that affect `zincPersist` basically add:
  * Converters from the Zinc API => protobuf and viceversa.
  * Mappers implementations.
  * `FileAnalysisStore` implementation.
* Changes that don't affect the `zincPersist` module have been done to adjust
  the use sites of the new introduced APIs. `AnalysisStore` and `AnalysisFile`
  are defined in the Java sources of the `zinc` module.
* Remove all the tools used in the previous implementation.
* Remove sbinary dependency and add protobuf dependency.

Therefore, I suggest you have pay attention to the following parts:
* Public binary API.
* Protobuf converters.

Performance-wise, note that the implementation can be much faster in the future
by streaming instead of converting first to the protobuf IR (or the other way
around) and then reading/writing. However, this is not worth it for now, in my
opinion, because of two reasons:

1. We don't want complexity in this part of the codebase. The dumber, the better.
2. The time spent on reading/writing analysis files account for 1-3% of the
   whole compile time. This, of course, may be significant for some *big*
   projects but we should not introduce complexity until that's proven with
   data. And it seems that no one complained about the text-based format; so
   being the binary format faster than that one as it currently, I don't
   consider it's worth it do streaming.

When it comes to readability, the new format provides ways to read the code via
ScalaPB's CLI tool and also javapb. From within the code, doing `toString` in
any class of the Protobuf API suffices to show the JSON code of the protos. I
will add instructions on how to run this to the CONTRIBUTING guide in a bit.

I'm adding two commits very shortly, but they won't affect the contents of this
PR. The pull request is up for review. Let me know if there's something I haven't covered yet.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/350/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/350/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/350/head&deploy_to_bintray=false">this link.</A>

jvican:
Weird, it seems CI is not in sync with Github and does not find the ref for the PR. I'm closing and opening a new PR.

Issue #320, opened 2017-06-20
Make `AnalysisMapper` public


stuhood:
`AnalysisMapper` looks like the right way to [avoid needing to be aware of zinc analysis content](https://github.com/sbt/zinc/issues/218), but currently it is an `sbt.internal.inc` API.

jvican:
The whole analysis mappers API is going to change drastically. It's highly tied to the serialization mechanism. The mechanism that I envision is more high-level operating only on the API classes and mapping from one class to the same class.
  
However, I must add that I'm working on making the content of analysis machine-independent. So this won't be necessary in theory. I'm just adding these hooks in case users need to perform more changes before serializing/deserializing.

stuhood:
Yea, these seem challenging to use correctly... not clear yet which of the methods need overrides in order to catch all relevant cases. Our implementation currently looks like https://github.com/pantsbuild/pants/compare/master...twitter:stuhood/zinc-upgrade-2017-06#diff-297b1b6a94d202511b784fd43284b6ebR166 ... but I'm not able to (easily) test portability until after `X17` is available.

romanowski:
@stuhood [Here](https://github.com/romanowski/hoarder/blob/master/0.13/hoarder/src/main/sbt_0.13/org.romanowski.hoarder/core/SbtAnalysisMapper.scala) is an example how this is done inside hoarder (code is ported to work with sbt 0.13.x but hopefully you will be able to get the main idea).

If you like I can try to explain what and how need to be mapped.

stuhood:
@romanowski : Thanks. Looks similar.

https://github.com/sbt/zinc/pull/351 is relevant. I'm hoping that @jvican agrees that a `getMachineIndependentMapper` implementation that supports multiple bases to relativize to is worth including... it would be really great if both hoarder and pants could use the default implementation.

jvican:
I'm happy to include such implementation. I'm just not sure how the interface should look like. If we want to make it generic, we'll need some assumptions that downstream users can agree on.
  
@stuhood Can you try to implement a rough prototype of `ReadMapper` and `WriteMapper` so that I get an idea of what you mean and we extrapolate a generic implementation from there?

jvican:
[...]

Issue #309, opened 2017-05-28
Rename any use of the term "Pimp" to "Enrich"


eed3si9n:
There's been a number of discussion on Twitter about how horrible the choice of term "Pimp" was to the impression of Scala:

- https://twitter.com/coda/status/93003343965851648
- https://twitter.com/coda/status/93049114106920961

> Refactored: plz don't use the "pimp" metaphor; it has unintended connotations which have offended and alienated potential Scala programmers.

The standard replacement is "Enrich", "Enrichment", or "RichX".
I sometimes call them "method injection" or "static monkey patching", but never pimp.

jvican:
I've never stopped to think about what pimp actually means. So I'm in for this change 👍 Good idea.

PR #351, opened 2017-07-10
Add protobuf-based analysis binary store

jvican:
This pull request replaces the text-based analysis format by a binary format
powered by Protobuf. The new format provides us with three main advantages:

1. Backwards and forwards binary compatibility at the analysis format level.
2. Faster serialization/deserialization of the analysis file.
3. Provides a much better way to make the analysis file machine-independent.

The purpose of this pull request was to add the benefits of a
backwards/forwards binary compatible analysis file without impacting the speed
of the text-based analysis format. In my local experiments, this goal has been
met. The new format is on par performance-wise with the text-based analysis
format for small analysis files, and it's much faster (~1.5-2x) when increasing
the Scalacheck generators (`maxSources` and `maxRelatives`) from 10 to 40, and
even more if the number increases.

With regards to the public API, the PR defines the public API in zinc persist,
which was all internal. This API is required by every piece of software that
integrates with Zinc and needs to modify the read/write steps in a sane way.

The API adds the following interfaces:

1. `ReadMappers`. To allow customisation before reading the analysis files.
2. `WriteMappers`. To allow customisation before writing the analysis files.
3. `ReadWriteMappers`. To wrap the read and write mappers.

These APIs all provide ways to get the default implementations from Java.

The most relevant implementations to discuss are the ones for `ReadMappers` and
`WriteMappers`. These implementations expose a way to relativize (make
machine-independent) an analysis file. Why would someone want to do that? To
ship analysis files to clusters and associate them with commits to avoid first
compilations when checking out a concrete commit.

These implementations, however, are not meant to cover all the possible use
cases for all build tools, they only provide a simple implementation for those
users that can ensure that all paths in the analysis file start with a concrete
prefix. This is usually not the case (read docs in the interfaces), and
therefore build tools will need to implement their own, taking into account all
the paths that can be present: ivy cache, coursier cache, binary dirs, product
dirs, et cetera. This approach was taken because it's impossible to produce an
elegant implementation that satisfies all build tools' requirements.

I've shipped today a release that @stuhood is using to integrate with these
changes and the ones I made to the loggers interface. It is accessible
[here](https://bintray.com/scalacenter/releases/zinc/1.0.0-X19-M1).

The implementation has been done iteratively so that every commit compiles,
making the final switch in the last commits.

I recommend that the people with most experience in this part of the codebase
(@eed3si9n and @Duhemm) review this PR. Also a ping for @stuhood. @dwijnand
Please feel free to have a view on it, too, if you'd like.

The way I recommend to review this PR is to have a look at the resulting diff.
It may look like huge, but it's actually conceptually really simple:

* Changes that affect `zincPersist` basically add:
  * Converters from the Zinc API => protobuf and viceversa.
  * Mappers implementations.
  * `FileAnalysisStore` implementation.
* Changes that don't affect the `zincPersist` module have been done to adjust
  the use sites of the new introduced APIs. `AnalysisStore` and `AnalysisFile`
  are defined in the Java sources of the `zinc` module.
* Remove all the tools used in the previous implementation.
* Remove sbinary dependency and add protobuf dependency.

Therefore, I suggest you have pay attention to the following parts:
* Public binary API.
* Protobuf converters.

Performance-wise, note that the implementation can be much faster in the future
by streaming instead of converting first to the protobuf IR (or the other way
around) and then reading/writing. However, this is not worth it for now, in my
opinion, because of two reasons:

1. We don't want complexity in this part of the codebase. The dumber, the better.
2. The time spent on reading/writing analysis files account for 1-3% of the
   whole compile time. This, of course, may be significant for some *big*
   projects but we should not introduce complexity until that's proven with
   data. And it seems that no one complained about the text-based format; so
   being the binary format faster than that one as it currently, I don't
   consider it's worth it do streaming.

When it comes to readability, the new format provides ways to read the code via
ScalaPB's CLI tool and also javapb. From within the code, doing `toString` in
any class of the Protobuf API suffices to show the JSON code of the protos. I
will add instructions on how to run this to the CONTRIBUTING guide in a bit.

I'm adding two commits very shortly, but they won't affect the contents of this
PR. The pull request is up for review. Let me know if there's something I haven't covered yet.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

eed3si9n:
I would add @xuwei-k as a reviewer as well since he's an expert in protobuf.

internal/compiler-interface/src/main/java/xsbti/api/Modifiers.java
```diff
@@ -37,6 +37,25 @@ public Modifiers(boolean isAbstract, boolean isOverride, boolean isFinal, boolea
 		);
 	}
 
+	/**
+     * Allow to set the modifiers from a flags byte where:
```
Duhemm, line 41: Formatting is off

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

internal/compiler-interface/src/main/java/xsbti/api/Modifiers.java
```diff
@@ -37,6 +37,25 @@ public Modifiers(boolean isAbstract, boolean isOverride, boolean isFinal, boolea
...
+	 *
+	 *   1. The first bit tells if has an abstract modifier.
+	 *   2. The first bit tells if has an override modifier.
+	 *   3. The first bit tells if has an final modifier.
+	 *   4. The first bit tells if has an sealed modifier.
+	 *   5. The first bit tells if has an implicit modifier.
+	 *   6. The first bit tells if has an lazy modifier.
+	 *   7. The first bit tells if has an macro modifier.
+	 *   8. The first bit tells if has an super accessor modifier.
```
Duhemm: first / second / etc.

internal/zinc-core/src/main/java/xsbti/compile/AnalysisContents.java
```diff
@@ -0,0 +1,18 @@
+package xsbti.compile;
+
+/**
+ * Defines an analysis file that contains information about every incremental compile.
```
Duhemm: > about every incremental compile

Is that not "about the last incremental compilation"?

internal/zinc-core/src/main/java/xsbti/compile/AnalysisContents.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * This information can be persisted using an {@link AnalysisStore}.
+ */
+public interface AnalysisContents {
+    /**
+     * @return An instance of {@link CompileAnalysis}.
```
Duhemm: How about
```
@return The `CompileAnalysis` held by this analysis file
```
?

internal/zinc-core/src/main/java/xsbti/compile/AnalysisContents.java
```diff
@@ -0,0 +1,18 @@
...
+     */
+    CompileAnalysis getAnalysis();
+
+    /**
+     * @return An instance of {@link MiniSetup}.
```
Duhemm: How about
```
@return The configuration of the compiler, a `MiniSetup`, that produced this analysis
```
?

internal/zinc-core/src/main/java/xsbti/compile/AnalysisStore.java
```diff
@@ -0,0 +1,37 @@
...
+    Optional<AnalysisContents> get();
+
+    /**
+     * Sets an {@link AnalysisContents} to the underlying store.
+     *
+     * The contents of the analysis file are necessary for subsequent incremental compiles
+     * given that the analysis files contains information about the previous incremental
+     * compile and lets the incremental compiler decide what needs or needs not to be recompiled.
+     *
+     * This method is called after every incremental compile.
```
Duhemm: After every successful incremental compilation

internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisStore.scala
```diff
@@ -9,30 +9,34 @@ package sbt
...
-}
+import xsbti.compile.{ AnalysisContents, AnalysisStore }
 
 object AnalysisStore {
-  def cached(backing: AnalysisStore): AnalysisStore = new AnalysisStore {
-    private var last: Option[(CompileAnalysis, MiniSetup)] = None
-    def set(analysis: CompileAnalysis, setup: MiniSetup): Unit = {
-      backing.set(analysis, setup)
-      last = Some((analysis, setup))
+  def cached(backing: AnalysisStore): AnalysisStore = new CachedAnalysisStore(backing)
```
Duhemm: While we're changing this, maybe add some doc or change to name to say that it's an in memory version of `backing`?

internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisStore.scala
```diff
@@ -9,30 +9,34 @@ package sbt
...
+  private final class CachedAnalysisStore(backing: AnalysisStore) extends AnalysisStore {
+    private var lastStore: Optional[AnalysisContents] = Optional.empty()
```
Duhemm: Not the best name. `lastContent`? `previous`?

internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisStore.scala
```diff
@@ -9,30 +9,34 @@ package sbt
...
+    override def set(analysisFile: AnalysisContents): Unit = {
+      backing.set(analysisFile)
+      lastStore = Optional.of(analysisFile)
     }
   }
-  def sync(backing: AnalysisStore): AnalysisStore = new AnalysisStore {
-    def set(analysis: CompileAnalysis, setup: MiniSetup): Unit = synchronized {
-      backing.set(analysis, setup)
+
+  def sync(backing: AnalysisStore): AnalysisStore = new SyncedAnalysisStore(backing)
```
Duhemm: I'm more reading this as in synchronous I/O rather than `synchronized` threads

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
```diff
@@ -58,10 +59,9 @@ trait WithPattern { protected def Pattern: Regex }
 import java.lang.{ Long => BoxedLong }
 
 /** Define the hash of the file contents. It's a typical stamp for compilation sources. */
-final class Hash(val value: Array[Byte]) extends StampBase {
-  val hexHash = IOHash.toHex(value)
+final class Hash(val hexHash: String) extends StampBase {
```
Duhemm: What's the advantage of `String` over `Array[Byte]` here?

internal/zinc-persist/src/main/java/sbt/inc/ReadMapper.java
```diff
@@ -0,0 +1,114 @@
...
+import sbt.internal.inc.mappers.RelativeReadMapper;
+import xsbti.compile.MiniSetup;
+import xsbti.compile.analysis.Stamp;
+
+import java.io.File;
+import java.nio.file.Path;
+
+/**
+ * Defines a reader-only mapper interface that is used by Zinc before read
+ * the contents of the analysis files from the persistent storage.
```
Duhemm: I would say it is used when reading the analysis.

internal/zinc-persist/src/main/java/sbt/inc/ReadMapper.java
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * This interface is useful to make the analysis file machine-independent and
+ * allow third parties to distribute these files around.
+ */
+public interface ReadMapper extends GenericMapper {
+
+    /**
+     * Defines a mapper that reads from a machine-independent analysis file.
+     *
+     * An analysis file is machine independent if all the paths are relative and no
```
Duhemm: Relative to the project's root?

internal/zinc-persist/src/main/java/sbt/inc/WriteMapper.java
```diff
@@ -0,0 +1,125 @@
...
+     * Defines a mapper that writes a machine-independent analysis file.
+     *
+     * An analysis file is machine independent if all the paths are relative and no
+     * information about the machine that produced it is stored -- only information
+     * about the structure of the project from a given build tool is persisted.
+     *
+     * The mapper makes sure that the analysis file looks like if it was generated by
+     * the local machine it's executed on.
+     *
+     * <b>Important note<b>: The assumption that all paths can be made relative to a concrete
```
Duhemm: The closing tag `</b>` is broken.

internal/zinc-persist/src/main/java/sbt/inc/WriteMapper.java
```diff
@@ -0,0 +1,125 @@
...
+     * position is not always correct. There is no guarantee that the paths of the caches, the
+     * artifacts, the classpath entries, the compilation inputs and outputs, et cetera share
+     * the same prefix. If this is not the case this reader will fail at runtime.
+     *
+     * Such assumption is broken in our test infrastructure: `lib_managed` does not share
```
Duhemm: I don't think this needs to appear in the documentation for the consumers of this API.

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
```diff
@@ -0,0 +1,99 @@
...
+      extends AnalysisStore {
+
+    private final val format = new BinaryAnalysisFormat(readWriteMappers)
+    private final val TmpEnding = ".tmp"
+
+    /**
+     * Get `CompileAnalysis` and `MiniSetup` instances for current `Analysis`.
+     */
+    override def get: Optional[AnalysisContents] = {
+      import JavaInterfaceUtil.PimpOption
```
Duhemm: I think we removed all mention of `Pimp` with something else. See #309

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
```diff
@@ -0,0 +1,99 @@
...
+import com.google.protobuf.{ CodedInputStream, CodedOutputStream }
+import sbt.inc.ReadWriteMappers
+import sbt.io.{ IO, Using }
+import xsbti.compile.{ AnalysisContents, AnalysisStore }
+
+import scala.util.control.Exception.allCatch
+
+object FileAnalysisStore {
+  private final val analysisFileName = "inc_compile.txt"
+  private final val companionsFileName = "api_companions.txt"
```
Duhemm: Change or remove the extensions? Not sure how Windows will survive binary files that end in `.txt`.

internal/zinc-persist/src/main/scala/sbt/internal/inc/ConcreteOutputs.scala
```diff
@@ -0,0 +1,21 @@
...
+ * This software is released under the terms written in LICENSE.
+ */
+
+package sbt.internal.inc
+
+import java.io.File
+
+import xsbti.compile.OutputGroup
+
+class ConcreteSingleOutput(val getOutputDirectory: File) extends xsbti.compile.SingleOutput {
```
Duhemm: Any reason for not making those `final`?

internal/zinc-persist/src/main/scala/sbt/internal/inc/cached/CacheAwareStore.scala
```diff
@@ -8,23 +8,28 @@
...
+    val previous = localStore.get().toOption.map(f => (f.getAnalysis, f.getMiniSetup))
     val cache = cacheProvider.findCache(previous)
     val cachedResult = cache.flatMap(_.loadCache(projectLocation))
-
-    cachedResult.orElse(previous)
+    val optResult = cachedResult.orElse(previous)
+    optResult match {
+      case Some((analysis, setup)) => Optional.of(ConcreteAnalysisContents(analysis, setup))
+      case None                    => Empty
+    }
```
Duhemm: I don't understand why we need this. It uses a `CacheProvider` to find in which file to write and read based on the value of `projectLocation`, and it takes care of the rebasing of the `Analysis`. (Is that correct?) How is that different from `FileBasedStore` + mappers?

internal/zinc-persist/src/main/scala/sbt/internal/inc/converters/InternalApiProxy.scala
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Defines a proxy to the Java compiler interface to create different utils.
+ *
+ * This proxy is required for an efficient deserialization of the analysis files.
+ * It exposes implementation details and uses protected methods to create new
+ * instances of other classes.
+ *
+ * Even though this proxy is public, Do not depend on it, it has no binary compatibility
+ * guarantee and can be broken in any minor release.
```
Duhemm: Can we move that to the `internal` namespace? Also, the path doesn't match the package name currently.

internal/zinc-persist/src/main/scala/sbt/internal/inc/converters/ProtobufReaders.scala
```diff
@@ -0,0 +1,643 @@
...
+    java.nio.file.Paths.get(path).toFile
+  }
+
+  def fromStampType(stampType: schema.Stamps.StampType): Stamp = {
+    import sbt.internal.inc.{ EmptyStamp, LastModified, Hash }
+    stampType.`type` match {
+      case schema.Stamps.StampType.Type.Empty            => EmptyStamp
+      case schema.Stamps.StampType.Type.Hash(h)          => new Hash(h.hash)
+      case schema.Stamps.StampType.Type.LastModified(lm) => new LastModified(lm.millis)
+      // ^ TODO: Double check that we recompute millis when reading this in certain conditions
```
Duhemm: Can you elaborate?

internal/zinc-persist/src/main/scala/sbt/internal/inc/converters/ProtobufReaders.scala
```diff
@@ -0,0 +1,643 @@
...
+import xsbti.compile.{ CompileOrder, FileHash, MiniOptions, MiniSetup, Output, OutputGroup }
+import xsbti.compile.analysis.{ Compilation, SourceInfo, Stamp }
+import sbt.internal.inc.converters.ProtobufDefaults.Feedback.StringToException
+import sbt.internal.inc.converters.ProtobufDefaults.Feedback.{ Readers => ReadersFeedback }
+import sbt.internal.inc.converters.ProtobufDefaults.{ Classes, ReadersConstants }
+import sbt.internal.util.Relation
+import xsbti.api._
+
+final class ProtobufReaders(mapper: ReadMapper) {
+  def fromPathString(path: String): File = {
```
Duhemm: The name doesn't really match what the function does

zinc/src/test/scala/sbt/inc/cached/CommonCachedCompilation.scala
```diff
@@ -86,6 +111,7 @@ abstract class CommonCachedCompilation(name: String)
 
     val prefix = tempDir.toPath.toString
 
+    // TODO(jvican): Add files that are missing...
```
Duhemm: Does this need to be addressed?

Duhemm (changes requested):
Amazing work!

I have really minor comments, mostly naming and docs... 

More importantly, it looks like `ReadMapper` and `WriteMapper` have the same interface. Why don't we have just one `AnalysisMapper`?

internal/zinc-persist/src/main/scala/sbt/internal/inc/converters/ProtobufReaders.scala
jvican: This can be removed. Good catch.

zinc/src/test/scala/sbt/inc/cached/CommonCachedCompilation.scala
jvican: This is just a comment to check not only the stamps, but also the files. Will fix in next commit. Good catch, forgot about this.

internal/zinc-persist/src/main/scala/sbt/internal/inc/converters/ProtobufReaders.scala
jvican: Why? It converts a path (represented by a string) to a file. The name only describes `from` not `to`. It's the convention all over the place.

internal/zinc-persist/src/main/scala/sbt/internal/inc/converters/InternalApiProxy.scala
jvican: No, it cannot. The thing is that the constructor that accepts a byte is protected, so only the things in the same namespace can use it. That's why it's defined in that namespace.

internal/zinc-persist/src/main/scala/sbt/internal/inc/cached/CacheAwareStore.scala
jvican: This is just some internal data structures that @gkossakowski added but that they don't really belong to this project (though they are good for testing). What this does is that it enables to do more "custom" behaviour as a build tool would do, have a look at `ProjectRebase`.
  
These utils kept being private on purpose, and have not been removed in this PR to keep the tests that use the mappers under the hood.
  
The way to go is the public API, that is `FileAnalysisStore` + mappers.

internal/zinc-persist/src/main/scala/sbt/internal/inc/ConcreteOutputs.scala
jvican: Nope, they can very well be :smile:. TOFIX.

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
jvican: True... need to rebase this. TOFIX.

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
jvican: True. Those are never exposed, they are only internal to the zip file though. Do you prefer the `.bin` extension?

internal/zinc-persist/src/main/java/sbt/inc/WriteMapper.java
jvican: Where do you think this should go? It has to go in a public place, otherwise Java users will not read it (the docs are also in the implementation, but consumers of this API may not have access to it).

internal/zinc-persist/src/main/java/sbt/inc/WriteMapper.java
jvican: Hehe, good catch. TOFIX.

internal/zinc-persist/src/main/java/sbt/inc/ReadMapper.java
jvican: TOFIX.

internal/zinc-persist/src/main/java/sbt/inc/ReadMapper.java
jvican: TOFIX.

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
jvican: See commit message. It removes an unnecessary transformation when reading/writing.

internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisStore.scala
jvican: Would you use `SynchronizedAnalysisStore` as a better name then? What's your suggestion?

internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisStore.scala
jvican: True. TOFIX.

internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisStore.scala
jvican: :+1:. TOFIX.

internal/zinc-core/src/main/java/xsbti/compile/AnalysisStore.java
jvican: TOFIX.

internal/zinc-core/src/main/java/xsbti/compile/AnalysisContents.java
jvican: TOFIX.

internal/zinc-core/src/main/java/xsbti/compile/AnalysisContents.java
jvican: TOFIX.

internal/zinc-core/src/main/java/xsbti/compile/AnalysisContents.java
jvican: Actually, it's about *one* incremental compile. Original docs are less than precise. TOFIX.

internal/compiler-interface/src/main/java/xsbti/api/Modifiers.java
jvican: True. TOFIX.

internal/compiler-interface/src/main/java/xsbti/api/Modifiers.java
jvican: TOFIX.

jvican:
Thanks for the prompt review, you're awesome @Duhemm.

> More importantly, it looks like ReadMapper and WriteMapper have the same interface. Why don't we have just one AnalysisMapper?
  
We cannot have one `AnalysisMapper` because we want to distinguish what's meant to be used to read and what's meant to be used to write. Why is this distinction useful? Because mappers for reading and writing do different things, and we don't want our users to pass in invalid mappers.

internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisStore.scala
Duhemm: Sounds good.

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
Duhemm: Sure.

internal/compiler-interface/src/main/java/xsbti/api/Modifiers.java
jvican: Apparently this is not off... It's how GitHub shows it.

internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisStore.scala
jvican: @Duhemm Actually, this is not an in-memory backing. There is no guarantee on what the underlying `backing` is.

jvican: I'm adding docs to `cached`, though. And also adding these options to the public API.

internal/zinc-persist/src/main/java/sbt/inc/ReadMapper.java
jvican: It is really not used "when" reading analysis, but "after" reading the contents.

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
jvican: I'm fixing this in the pertinent commit that introduced the change.

internal/zinc-persist/src/main/scala/sbt/internal/inc/converters/ProtobufReaders.scala
jvican: @Duhemm ping.

Duhemm: Oh I read `fromPathToString`, sorry. It's perfect the way it is then 👍

build.sbt
```diff
@@ -197,13 +197,14 @@ lazy val zincCompile = (project in file("zinc-compile"))
...
 
-// Persists the incremental data structures using SBinary
+// Persists the incremental data structures using Protobuf
 lazy val zincPersist = (project in internalPath / "zinc-persist")
   .dependsOn(zincCore, zincCore % "test->test")
   .configure(addBaseSettingsAndTestDeps)
   .settings(
     name := "zinc Persist",
-    libraryDependencies += sbinary
+    compileOrder := sbt.CompileOrder.Mixed,
```
eed3si9n: Why was this necessary?

build.sbt
```diff
@@ -197,13 +197,14 @@ lazy val zincCompile = (project in file("zinc-compile"))
...
+    PB.targets in Compile := List(scalapb.gen() -> (sourceManaged in Compile).value)
```
eed3si9n: Where are the protobuf binding code? Could you put them in src like I do for Contraband?

internal/zinc-persist/src/main/java/sbt/inc/GenericMapper.java
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Defines a generic interface to map the values of the analysis file 1-to-1.
+ */
+public interface GenericMapper {
+    /**
+     * @param sourceFile The source file to compiled.
+     * @return A valid, modified source file.
+     */
+    public File mapSourceFile(File sourceFile);
```
eed3si9n: Would it make sense to switch to using NIO Path wherever Files are being used?

internal/zinc-core/src/main/java/xsbti/compile/AnalysisStore.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.Optional;
+
+/**
+ * Defines a store interface that provides analysis read and write capabilities to users.
+ *
+ * The store is a backend-independent interface that allows implementors to decide how
+ * the analysis stores are read and written before or after every incremental compile.
+ */
+public interface AnalysisStore {
```
eed3si9n: As far as I can tell, the impact to sbt/sbt is around the relatively small changes around the storage.
Do you think you can split it up into another PR along with the ripple PRs?

eed3si9n (changes requested):
Thanks for the contribution.

Although I am excited about the outlook for performance improvements, I am genuinely concerned about the size and scale of this PR.
1. Could we split up AnalysisStore related change to another PR? That can be contained on its own, correct?
2. What do you think about potentially providing both mechanism in Zinc 1 so downstream sbt users can potentially fallback to the old and trusty implementation, while early adopters can enjoy the speedup?

build.sbt
jvican: This is the way to configure it.

build.sbt
jvican: Because Java sources depend on Scala sources, and viceversa. It's pretty common in Zinc modules.

internal/zinc-persist/src/main/java/sbt/inc/GenericMapper.java
jvican: I didn't want to make this change in this PR.

internal/zinc-core/src/main/java/xsbti/compile/AnalysisStore.java
jvican: I can split it. The reason why this is in the same PR is because the protobuf PR touches all these utilities. Breaking it in another PR is going to force me to depend on this, rebase until it's done, and then create an extra PR in sbt/sbt to synchronize with this. Is this worth it?

jvican:
> What do you think about potentially providing both mechanism in Zinc 1 so downstream sbt users can potentially fallback to the old and trusty implementation, while early adopters can enjoy the speedup?
  
I think this is a bad idea and goes against the whole purpose of this PR; the creation of a binary backwards and forwards format that Zinc 1.0 can evolve over time. I see no benefit in doing this. We want everyone to use the binary format so that we don't depend on the constraints of the text-based format if we evolve the API.

internal/zinc-core/src/main/java/xsbti/compile/AnalysisStore.java
jvican: Note that this change is just one commit.

jvican:
Also, on another note:
> Although I am excited about the outlook for performance improvements
  
This PR is not only about performance improvements :smile:. It's about allowing organizations to ship analysis files with the guarantee that any Zinc 1.x version will be able to read/write (backwards and forwards).

build.sbt
eed3si9n: We usually try to isolate the Java interfaces into subprojects like xsbti. If the API layer is depending on Scala, that's a smell.

eed3si9n:
> This PR is not about performance improvements 😄. It's about allowing organizations to ship analysis files with the guarantee that any Zinc 1.x version will be able to read/write (backwards and forwards).

If so we can do that using JSON.

jvican:
> If so we can do that using JSON.
  
Not without a performance penalty. This PR improves both: performance and binary compatibility.
  
Also, JSON does not allow forwards binary compatibility, and has problems since it serialises the names of the members. Protobuf does not. Protobuf supports both forwards and backwards binary compatibility via index.

build.sbt
jvican: The protobuf binding code is in `sbt-protoc`.

eed3si9n:
> Also, JSON does not allow forwards binary compatibility, and has problems since it serialises the names of the members. Protobuf does not. Protobuf supports both forwards and backwards binary compatibility.

JSON uses key names, and you can make any fields optional.
Protobuf uses key numbers, and it also allows you to define optional fields. There's nothing magical difference between the two.

jvican:
> JSON uses key names, and you can make any fields optional. Protobuf uses key numbers, and it also allows you to define optional fields. There's nothing magical difference between the two.
  
There *is* one difference: Protobuf allows you to modify the names of the members, JSON does not. This is great to improve APIs, allowing you to rename reserved names without polluting the API.
  
That and the performance benefits that Protobuf brings to the projects is a good improvement over sjson, IMO. This is why this PR targets Protobuf over the json alternative.

xuwei-k:
> There's nothing magical difference between the two.


> There is one difference: Protobuf allows you to modify the names of the members, JSON does not. 

I thinks there are some **more** important difference.

protobuf version3 has only 4 types without deprecated types, it means protobuf has better **compatibility keepability** compared to json.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

internal/zinc-core/src/main/java/xsbti/compile/AnalysisStore.java
jvican: I have just removed this commit from the PR.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

jvican:
CI is finally passing :tada:. I've published sbt-protoc myself until https://github.com/thesamet/sbt-protoc/pull/27 gets merged and released.

build.sbt
jvican: Java interfaces depend on Scala code to provide the default implementations of certain entities. This is extensively used in all the Zinc modules and it's a legit use case. This is the only way to provide a Java-friendly API.

jvican:
This PR addresses the following tickets, which are strictly coupled:
1. https://github.com/sbt/zinc/issues/218.
2. https://github.com/sbt/zinc/issues/320
3. https://github.com/sbt/zinc/issues/268

tayama0324:
I discussed with @xuwei-k offline. Let me explain in detail about protobuf and JSON compatibility.

> There is one difference: Protobuf allows you to modify the names of the members, JSON does not.

Another difference is compatibility of array type. Protobuf fields can be replaced with `repeated` fields of same type (and vice versa) without breaking anything, while JSON can not, that is, `[ "foo" ]` and `"foo"` are incompatible.

The magic here is data structure of protobuf binary format. Fields of a message are represented as a sequence of `(tag number, value)`, and deserializers are requested to preserve only lastly read value of non-repeated fields. That is why protobuf fields can be `repeated` in backward and forward compatible manner.

Some more differences you may be interested in are as follows:

- In protobuf `int`s and `bool`s are compatible, while JSON is not.
- In protobuf `bytes` and `string`s are compatible, while JSON is not.
- In JSON integers and floating point numbers are compatible, while protobuf is not.

You may think compatibility rule of protobuf binary is tricky. However, you will find that it is very simple and flexible if once you would understand underlying binary format and (de)serialization mechanism. IMO, protobuf binary is preferred choice than JSON if you need compatibility.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
dwijnand: .. at the cost of type-safety (not all strings are hex strings; note the use of the name`hexString` rather than plain `value`). Previously this would fail-fast on bad data (e.g `"hash(gobbledygook)"`).

Also given each byte is transformed into 2 chars I wonder about the equality performance gain claim. Maybe String.equals is special-cased.

internal/zinc-persist/src/main/java/xsbti/compile/analysis/GenericMapper.java
```diff
@@ -0,0 +1,93 @@
...
+
+import xsbti.compile.MiniSetup;
+
+import java.io.File;
+
+
+/**
+ * Defines a generic interface to map the values of the analysis file 1-to-1.
+ */
+public interface GenericMapper {
```
stuhood, line 18: As previously mentioned, ReadMapper and WriteMapper being separate interfaces extending GenericMapper feels redundant. At some point it's necessary to actually construct one of these, and given that they will always be perfect inverses, it's not clear that their implementations would be so obviously different that this would catch bugs.

While updating to this branch, I ended up doing: https://github.com/pantsbuild/pants/compare/master...twitter:stuhood/zinc-upgrade-2017-06-update-to-X19#diff-3c0fb6f9440b619e7d350cfa7c3b6228R62 ... ie, implementing a GenericMapper and then creating two subclasses... which just feels like boilerplate. Not a huge deal though.

internal/zinc-persist/src/main/java/xsbti/compile/analysis/ReadMapper.java
```diff
@@ -0,0 +1,125 @@
...
+     * Defines a mapper that reads from a machine-independent analysis file.
+     *
+     * An analysis file is machine independent if all the paths are relative to a path
+     * and no information about the machine that produced it is stored -- only
+     * information about the structure of the project from a given build tool is persisted.
+     *
+     * The mapper makes sure that the analysis file looks like if it was generated by
+     * the local machine it's executed on.
+     *
+     * <b>Important note</b>: The assumption that all paths can be made relative to a concrete
```
stuhood, line 36: I think that this is a very important point. If it is possible to shrink or completely eliminate the `*Mappers` API by implementing `getMachineIndependentMapper` in a way that works for all consumers, this will be much less fragile. As it stands, zinc is punting to consumers to have them figure out which portions of the analysis internals are not portable... when zinc itself is best positioned to do that.

As previously discussed, I'd be OK with leaving the `*Mappers` API public and marked experimental, but it would be great to have an explicit goal of making getMachineIndependentMapper a default that will work for 95% of consumers.

internal/zinc-persist/src/main/protobuf/schema.proto
```diff
@@ -0,0 +1,444 @@
+syntax = "proto3";
+
+package sbt.internal.inc;
+
+///////////////////////////////////////////////////////////////////////////////////////////////
+///////////////////////////////////////// ANALYSIS API ////////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+message Hash {
+    string hash = 1;
```
stuhood, line 10: As alluded to earlier, just making this a `bytes` value might skip the whole problem of decoding/encoding.

internal/zinc-persist/src/main/protobuf/schema.proto
```diff
@@ -0,0 +1,444 @@
...
+        }
+    }
+
+    map<string, StampType> productStamps = 1;
+    map<string, StampType> binaryStamps = 2;
+    map<string, StampType> sourceStamps = 3;
+}
+
+message OutputGroup {
+    string source = 1;
```
stuhood: Since these are stringly typed, you might consider including more typeinfo in their names: `sourceFile` or `sourcePath`, etc.

internal/zinc-persist/src/main/protobuf/schema.proto
```diff
@@ -0,0 +1,444 @@
...
+message MultipleOutput {
+    repeated OutputGroup outputGroups = 1;
+}
+
+message SingleOutput {
+    string target = 1;
+}
+
+message Compilation {
+    int64 startTime = 1;
```
stuhood: Hash includes `millis` in the name of the field... useful here too probably. `startTimeMillis`, etc.

internal/zinc-persist/src/main/protobuf/schema.proto
```diff
@@ -0,0 +1,444 @@
...
+}
+
+message Tuple {
+    string first = 1;
+    string second = 2;
+}
+
+message MiniOptions {
+    repeated FileHash classpathHash = 1;
+    repeated string scalacOptions = 2;
```
stuhood, line 106: These are arrays in the public API, which seems more reasonable to me since you don't need to lex them? But maybe doesn't matter.

internal/zinc-persist/src/main/protobuf/schema.proto
```diff
@@ -0,0 +1,444 @@
...
+    map<string, AnalyzedClass> external = 2;
+}
+
+message UsedName {
+    string name = 1;
+    repeated UseScope scopes = 2;
+}
+
+message Values {
+    repeated string values = 1;
```
stuhood: I had no idea that protobuf doesn't support typedefs (bummer!)... but it would be good to add comments for a bunch of these `string` usages to indicate what they're expected to contain.

internal/zinc-persist/src/main/scala/sbt/internal/inc/binary/converters/ProtobufReaders.scala
```diff
@@ -0,0 +1,641 @@
...
+import sbt.internal.inc.binary.converters.ProtobufDefaults.{ Classes, ReadersConstants }
+import sbt.internal.util.Relation
+import xsbti.api._
+
+final class ProtobufReaders(mapper: ReadMapper) {
+  def fromPathString(path: String): File = {
+    java.nio.file.Paths.get(path).toFile
+  }
+
+  def fromStampType(stampType: schema.Stamps.StampType): Stamp = {
```
stuhood, line 29: Possible to make a bunch of these private? Only used internally.

stuhood (approved):
Thanks @jvican : this looks good to me.

I've incorporated this via your branch, and had success for simple usecases. One issue (#355) is blocking deeper integration testing, but I'm hoping that once this lands it will be easier to dig deeper.

stuhood:
Also, to be clear: strongly in favor of using protobuf/thrift/avro, or some other explicitly schema-ed and widely-used serialization format rather than json.

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
jvican: @dwijnand You're right, this assumes that the internals will instantiate `Hash` with a hexadecimal value. But this is kind of the point, as this API does not leak to the users and there is only one legit way of creating a hash stamp:
```
  val forHash = (toStamp: File) => tryStamp(new Hash(IOHash.toHex(IOHash(toStamp))))
```
  
... which uses `IOHash.toHex` and makes sure we pass in a valid string to the constructor.
  
The reason why I made this change is because I think comparing bytes is more expensive than strings. `String.==` only checks referential equality, while `sameElements` in an array of bytes causes the comparison of every chunk of data (where chunk is a byte, though it could be a little bit optimized by the JVM): `case (h1: Hash, h2: Hash) => h1.value sameElements h2.value`.
  
What we can do to fix the "type-safety" issue is to make the constructor private and ensure that the instantiation that I show before is the only one that instantiates the class. WDYT?

dwijnand: > What we can do to fix the "type-safety" issue is to make the constructor private and ensure that the instantiation that I show before is the only one that instantiates the class. WDYT?

Yes, guaranteeing by construction would restore this safety.

> `String.==` only checks referential equality

You sure?

```scala
Welcome to Scala 2.12.2 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_131).

scala> val s1 = new String("abc")
s1: String = abc

scala> val s2 = new String("abc")
s2: String = abc

scala> s1 eq s2
res0: Boolean = false

scala> s1 == s2
res1: Boolean = true
```

internal/zinc-core/src/main/scala/sbt/internal/inc/APIDiff.scala
```diff
@@ -34,33 +34,33 @@ private[inc] class APIDiff {
...
-    * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
-    * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
-    * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE
-    * GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
-    * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
-    * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
-    * OF THE POSSIBILITY OF SUCH DAMAGE.
+   * This class was directly copied from Dotty
+   * [[https://github.com/lampepfl/dotty/blob/0.1.2-RC1/compiler/src/dotty/tools/dotc/util/DiffUtil.scala]]
+   *
```
dwijnand: Why wasn't this caught by Drone in the PR that introduced it?

jvican: Good question, no idea. It seems this may be an error in `scalafmt::test`?

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
jvican: Yes, I'm sure. Scalac's backend transforms `==` into a ref equality check + value equality check. That's why your snippet shows this behaviour. So if two strings point to the same memory pointer (or they are interned) will short-circuit `==`'s implementation.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
dwijnand: Ah I see what you meant now.

jvican: Cool, I'll amend that commit to turn the constructor private. Thanks for the suggestion.

internal/zinc-persist/src/main/protobuf/schema.proto
jvican: I have addressed Dale's reviews further up this PR, the decision is to keep this a string both in the Scala internal API and the internal protobuf one. Let me know if that's ok for you.

stuhood: Fine with me.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

internal/zinc-persist/src/main/scala/sbt/internal/inc/binary/converters/ProtobufReaders.scala
jvican: It's a good idea, but the whole readers/writers are internal. The reason I decided to leave them like that is just in case some downstream user needs them at any point... That way Zinc wouldn't be blocking her/him from using them. What do you think?

internal/zinc-persist/src/main/protobuf/schema.proto
jvican: Good idea, addressing it :smile:.

internal/zinc-persist/src/main/protobuf/schema.proto
jvican: Actually, you cannot use `Array` directly with scalapb, it defaults to `Seq`... So this change is not possible.

internal/zinc-persist/src/main/protobuf/schema.proto
jvican: Done.

internal/zinc-persist/src/main/java/xsbti/compile/analysis/GenericMapper.java
jvican: I think that having this distinction at the type level is important to avoid "mistakes" by the users. A good example of implementations that differ is `RelativeReadMapper` and `RelativeWriteMapper`.
  
Have you tried defining your portable analysis mappers like this?
```scala
class PortableMapper(mapper: File => File) extends ReadMapper with WriteMapper {
  def mapSourceFile(x: File): File = mapper(x)
  def mapBinaryFile(x: File): File = mapper(x)
  def mapProductFile(x: File): File = mapper(x)
  def mapOutputDir(x: File): File = mapper(x)
  def mapSourceDir(x: File): File = mapper(x)
  def mapClasspathEntry(x: File): File = mapper(x)

  // TODO: Determine whether the rest of these need to be overridden in practice.
  def mapJavacOption(x: String): String = x
  def mapScalacOption(x: String): String = x
  def mapBinaryStamp(f: File, x: Stamp): Stamp = x
  def mapSourceStamp(f: File, x: Stamp): Stamp = x
  def mapProductStamp(f: File, x: Stamp): Stamp = x
  def mapMiniSetup(x: MiniSetup): MiniSetup = x
}
```
  
I think this way is much nicer. WDYT?

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

internal/zinc-persist/src/main/protobuf/schema.proto
jvican: Done.

internal/zinc-persist/src/main/java/xsbti/compile/analysis/ReadMapper.java
jvican: I addressed your concerns here: https://github.com/sbt/zinc/issues/320. Can you have a look?

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

jvican:
@stuhood I've addressed all your feedback, thanks for your comments! :smile:
  
This PR is ready for review again. /cc @eed3si9n @Duhemm 
Please, review only the bits I've fixed, that is the following commits:
```
@jvican	Address Martin's feedback	Verified		a2980af
 @jvican	Move implementations to independent packages  …	Verified		1cd288e
 @jvican	Make the binary format the default implementation  …	Verified		48253db
 @jvican	Rename `TestCaseGenerators` to `AnalysisGenerators`	Verified		65587e1
 @jvican	Move mappers to `xsbti.compile.analysis`  …	Verified		1fccf93
 @jvican	Remove old mappers and rework relative path utils  …	Verified		8661d12
 @jvican	Use `mapSourceFile` for sources in source infos  …	Verified		02545fc
 @jvican	Address Stu's feedback	Verified		bcab1b1
 @jvican	Add generic read and write mappers  …	Verified		6efd16c
```
  
The commits remove the previous `AnalysisMappers` utils and replaces them by the new ones. They also add generic read and write mappers that can be used by all the build tools in a better way than the existing and limited `NaiveRelative{Read,Write}Mapper`s.
  
We can merge as soon as @stuhood is happy with my proposed mappers and Eugene and Martin LGTM.

**Status update**: The Scala Center (@Duhemm and I) + sbt team (@eed3si9n and @dwijnand ) discussed this PR two days ago via Hangouts. The verdict of our discussions is that we're merging the binary analysis format proposed in this PR and making it the default, but contrary to my original goals we will not remove the text-analysis format. The argument for doing this is that downstream projects and their users have a fallback method in case something goes wrong.
  
I'm working now on opening the PRs to synchronize this in sbt/sbt.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

internal/zinc-persist/src/main/scala/sbt/internal/inc/mappers/MapperUtils.scala
```diff
@@ -0,0 +1,72 @@
...
+
+import xsbti.compile.analysis.Stamp
+
+object MapperUtils {
+  private[inc] def rebase(target: File, from: Path, to: Path): File = {
+    to.resolve(from.relativize(target.toPath)).toFile
+  }
+
+  /** Defines a marker that tells the utils that the relativized path is empty. */
+  private final val RELATIVE_MARKER = "\u2603\u2603\u2603"
```
Duhemm, line 21: I don't think that's super important, but maybe we should use something inside Unicode's PUA here?

internal/zinc-persist/src/main/scala/sbt/internal/inc/mappers/MapperUtils.scala
```diff
@@ -0,0 +1,72 @@
...
+  private final val MARKER_LENGTH = RELATIVE_MARKER.length()
+
+  private final def relativeReadError(path: String) =
+    s"Unexpected path $path was not written by a relative write mapper. Paths have to start with $RELATIVE_MARKER"
+
+  /**
+   * Makes a file relative to a path.
+   *
+   * This function writes all file paths to a relative format that has a [[RELATIVE_MARKER]]
```
Duhemm: It converts rather than write, no?

Duhemm (changes requested):
Really minor comments, and then LGTM! Awesome work 🦄

internal/zinc-persist/src/main/scala/sbt/internal/inc/mappers/MapperUtils.scala
jvican: This is the unicode symbol used internally for the synthesized macro names by sbt. It's safe to be used because the text format uses UTF8 (we force it when we write and read) and protobuf also uses unicode (https://developers.google.com/protocol-buffers/docs/proto, look for "unicode" and "UTF").
  
Given the fact that this symbol is rarely used, I think it's perfect for this scenario. Some SO questions point to it as a good way to disambiguate names or use it for "marking" string boundaries.

internal/zinc-persist/src/main/scala/sbt/internal/inc/mappers/MapperUtils.scala
jvican: Thanks!

jvican: Fixed.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/351/head&deploy_to_bintray=false">this link.</A>

Duhemm (approved):

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
```diff
@@ -58,10 +59,10 @@ trait WithPattern { protected def Pattern: Regex }
 import java.lang.{ Long => BoxedLong }
 
 /** Define the hash of the file contents. It's a typical stamp for compilation sources. */
-final class Hash(val value: Array[Byte]) extends StampBase {
-  val hexHash = IOHash.toHex(value)
+final class Hash private[inc] (val hexHash: String) extends StampBase {
```
dwijnand, line 62: `private[inc]` is no guarantee - as you see `Stamp.fromString` invokes it without doing any validation:

```scala
scala> sbt.internal.inc.Stamp.fromString("hash(gobbledygook)")
res0: xsbti.compile.analysis.Stamp = hash(gobbledygook)
```

dwijnand (changes requested):

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
jvican: Yeah, I know. But it's the best we can do. There's no way to define it `private` and instantiate it in the same file.

typesafe-tools:
A dbuild validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/90/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: 1.0.x-pull/351/head-1.0-1.0-1.0-1.0.x)
dbuild <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/90/">has checked</A> the following projects against Scala 2.12:

  sbt               | <A HREF="https://github.com/sbt/sbt/tree/1.0.x">1.0.x               </A> | sbt/sbt@716d664d64d183384408734fa570e63e1a21f99e
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/351">pull/351/head       </A> | sbt/zinc@690b7a4336756812e8ed53470ee660c64f4f7880
  io                | <A HREF="https://github.com/sbt/io/tree/1.0">1.0                 </A> | sbt/io@f55849d988b94941309d316c43533b0af423f244
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/1.0">1.0                 </A> | sbt/librarymanagement@f8d3718d283cacdc11c6d4a9f101c675f687dc49
  util              | <A HREF="https://github.com/sbt/util/tree/1.0">1.0                 </A> | sbt/util@fe7743878fb590765582aec86d0e45df6ae0a936
  website           | <A HREF="https://github.com/sbt/website/tree/1.0.x">1.0.x               </A> | sbt/website@7e644bc8fc623589152e98bee75afd44893690e3

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild?sbt=1.0.x&zinc=pull/351/head&io=1.0&librarymanagement=1.0&util=1.0&website=1.0.x&deploy_to_bintray=false">(restart)</A>

typesafe-tools:
A dbuild validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/95/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: 1.0.x-pull/351/head-1.0-74d8a3835d535d3d787e9294539a0a3efc47674d-1.0-1.0.x)
dbuild <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/95/">has checked</A> the following projects against Scala 2.12:

  sbt               | <A HREF="https://github.com/sbt/sbt/tree/1.0.x">1.0.x               </A> | sbt/sbt@716d664d64d183384408734fa570e63e1a21f99e
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/351">pull/351/head       </A> | sbt/zinc@690b7a4336756812e8ed53470ee660c64f4f7880
  io                | <A HREF="https://github.com/sbt/io/tree/1.0">1.0                 </A> | sbt/io@f55849d988b94941309d316c43533b0af423f244
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/74d8a3835d535d3d787e9294539a0a3efc47674d">74d8a3835d535d3d787e9294539a0a3efc47674d</A> | sbt/librarymanagement@74d8a3835d535d3d787e9294539a0a3efc47674d
  util              | <A HREF="https://github.com/sbt/util/tree/1.0">1.0                 </A> | sbt/util@fe7743878fb590765582aec86d0e45df6ae0a936
  website           | <A HREF="https://github.com/sbt/website/tree/1.0.x">1.0.x               </A> | sbt/website@7e644bc8fc623589152e98bee75afd44893690e3

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild?sbt=1.0.x&zinc=pull/351/head&io=1.0&librarymanagement=74d8a3835d535d3d787e9294539a0a3efc47674d&util=1.0&website=1.0.x&deploy_to_bintray=false">(restart)</A>

typesafe-tools:
A dbuild validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/96/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: 1.0.x-pull/351/head-1.0-pull/122/head-1.0-1.0.x)
dbuild <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/96/">has checked</A> the following projects against Scala 2.12:

  sbt               | <A HREF="https://github.com/sbt/sbt/tree/1.0.x">1.0.x               </A> | sbt/sbt@716d664d64d183384408734fa570e63e1a21f99e
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/351">pull/351/head       </A> | sbt/zinc@690b7a4336756812e8ed53470ee660c64f4f7880
  io                | <A HREF="https://github.com/sbt/io/tree/1.0">1.0                 </A> | sbt/io@f55849d988b94941309d316c43533b0af423f244
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/pull/122">pull/122/head       </A> | sbt/librarymanagement@e3891c8732dc248bd3ff3be9d5c796454d66bc3a
  util              | <A HREF="https://github.com/sbt/util/tree/1.0">1.0                 </A> | sbt/util@fe7743878fb590765582aec86d0e45df6ae0a936
  website           | <A HREF="https://github.com/sbt/website/tree/1.0.x">1.0.x               </A> | sbt/website@7e644bc8fc623589152e98bee75afd44893690e3

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild?sbt=1.0.x&zinc=pull/351/head&io=1.0&librarymanagement=pull/122/head&util=1.0&website=1.0.x&deploy_to_bintray=false">(restart)</A>

jvican:
@eed3si9n This is ready for review. The validator fails with the expected errors and this PR doesn't require any sbt/sbt PR because it doesn't affect the public API.

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
dwijnand: Sure there is, define it `private[this]` and constructed it from Hash's companion object, after validating the input.

jvican: No, I don't think that's possible. `Hash` is instantiated from two different objects: `Stamp` and `Stamper`, and wrapping them into Hash's companion object is not a valid solution.

dwijnand: Wrapping them? You define a safe method in Hash's companion object that does the validation and then call that method from Stamp and Stamper.

jvican: The validation is in the types. `Stamp` needs to pass in a string (which is the deserialized format). Adding a method in Hash's companion object does not solve anything: instead of exposing the constructor with string you will expose a method that has the same signature. People can still call that from anywhere in Zinc.

eed3si9n (approved):
LGTM
Thanks for the keeping the text around

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
dwijnand: What does "The validation is in the types" mean? I'd just like it be impossible to create illegal Hash objects. Currently it's trivial.

eed3si9n: This looks like a minor detail we can come back later, so I am going to merge this PR as is.

jvican: > What does "The validation is in the types" mean? I'd just like it be impossible to create illegal Hash objects. Currently it's trivial.
  
What I mean is that it's not possible to prevent the creation of illegal hash objects because `Stamp` needs access to a constructor that takes a string. Whether that is provided by a constructor or a method in the companion object produces the same result: it has to be accessible from `Stamp`, which disqualifies it from being defined as `private[this]`. The only way this can be supported is by wrapping all the definitions inside an object or a package, and then fixing all the imports. Too big of a change to do it here. I'm happy to do open a ticket to do it.
  
All in all, with the current organisation of the code, `private[this]` cannot be used and that's why we use `private[inc]`, which at least gives us something.

jvican:
:shipit:

jvican:
Note for future archaeologists: the plan is to remove the text-based analysis format in Zinc 1.1.

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
dwijnand: Sure you can prevent the creation. You can pass back an `Option[Hash]`, or an `Either[String, Hash]`, or you can stick to what zinc was doing previous which is throw an exception. Either way you prevent the creation of an illegal hash object.

jvican: Either of those alternatives have a runtime performance penalty (returning `Option` and `Either` require validation), which goes against the goal of the commit that introduced this change.
  
IMO, the best to solve this issue is to put all those stamp utils in its own package, and then do `private[stamps]`.

jvican:
@eed3si9n Are we waiting for something to merge this? I need to rebase in my other PR.

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
eed3si9n: `fromString` is already doing regex pattern matching. If you want to tighten it you should change `final val Pattern = """hash\((\w+)\)""".r` in object Hash to accept hex only.

dwijnand: Good idea, validation is being done there. I'll fix it with that.

dwijnand (approved):

internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
jvican: Yeah, good idea. I'll replace it with `"""hash\(([0-9a-fA-F]+)\).r"""`.

PR #361, opened 2017-07-14
Guarantee Hash contains a hex string by construction

dwijnand:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/361/head&deploy_to_bintray=false">this link.</A>

eed3si9n (dismissed):

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/361/head&deploy_to_bintray=false">this link.</A>

jvican (changes requested):
I don't think this is necessary. I don't consider necessary to check that a hash is correct with `fromString` and pay the price of checking all the source stamps against a regex.
  
There's even less value in doing this if from `ProtobufReaders` we are going to fail with a plain exception.
  
The readers and writers of the analysis file have fair assumptions. Assuming that the stamp we're reading is a hexadecimal value is a fair assumption, honestly. I fail to see this giving us any value.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/361/head&deploy_to_bintray=false">this link.</A>

jvican (approved):
I've seen your changes. If you only want to enforce the regex in the text analysis format, LGTM.

dwijnand:
Thank you for approving.

PR #370, opened 2017-07-17
Add Windows CI support

### LINKED ISSUE #367: Add CI for Windows
I was thinking it would be wise to add a CI hook for AppVeyor so that Zinc is checked in Windows. This will give us more guarantees when we make releases and help us predict any Windows-related error before it happens in production. Some of our users run it in Windows, so it's important we cover this scenario in our testing infrastructure.
  
AppVeyor is free for open-source projects. To avoid making the CI slow, we can constrain the amount of tests that we run -- cherry-picking the most important ones perhaps. But first we need to experience a slowdown in comparison with our Drone CI. Note that this proposal is about having the two enabled.

jvican:
Fixes #367

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,38 @@
+environment:
+  global:
+    # Unlike bin/run-ci.sh, AppVeyor does not execute `crossTestBridges` nor the benchmark tests (which fail because they depend on *nix systems).
+    COMPILE_AND_TEST: ";zincRoot/test:compile; scalafmt::test; test:scalafmt::test; publishBridgesAndSet %SCALA_VERSION%; set test in zincBenchmarks := Def.task(()); zincRoot/test; zincRoot/scripted"
+  matrix:
+    - SCALA_VERSION: 2.12.2
+install:
+  - ps: |
+      Add-Type -AssemblyName System.IO.Compression.FileSystem
+      if (!(Test-Path -Path "C:\sbt\sbt" )) {
+        (new-object System.Net.WebClient).DownloadFile(
+          'https://github.com/sbt/sbt/releases/download/v1.0.0-M6/sbt-1.0.0-M6.zip',
+          'C:\sbt-1.0.0-M6.zip'
+        )
+        [System.IO.Compression.ZipFile]::ExtractToDirectory("C:\sbt-1.0.0-M6.zip", "C:\sbt")
+      }
+  - ps: |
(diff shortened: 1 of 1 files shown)


dwijnand (dismissed):
LGTM

jvican:
File is inspired from https://github.com/ensime/scala-debugger/blob/master/appveyor.yml. My understanding of Windows is limited, to say the least, but sounds like this should do it.
  
As AppVeyor is quite slow, I've removed `crossTestBridges`, which is the command running the test of the bridges, and left the unit and scripted tests. The reason why I'm removing the test of the bridges is speed. AppVeyor won't keep up with our intensive pipeline and these tests are the more likely to run successfully in Windows (since they are just a compiler plugin).  It goes without saying that the bridges are actually tested in scripted too, so that should be safe-ish.
  
Objections?

dwijnand:
I'm happy with some, initial Windows testing over the current none. We can tweak over time.

jvican:
I don't know if to be happy or not :cry: :
```
[error] (zincRoot/*:scripted) Failed tests:
[error] 	source-dependencies / canon
[error] 	source-dependencies / packageobject-and-traits
[error] 	source-dependencies / recorded-products
[error] Total time: 256 s, completed Jul 17, 2017 9:37:20 P
```

jvican:
The most impressing part is that a pending test for Linux and OSX is passing in Windows: `packageobject-and-traits`. LOL.
  
**EDIT**: It must be psasing because of an unexpected IO error, since the tests expects `compile` to fail.

jvican:
Status:
- [ ] `source-dependencies/canon` -> Looks like a scalac classpath error. Need to reproduce and report.
- [ ] `source-dependencies/recorded-products` -> A simple error that assumes `/` instead of `\`.
- [ ] `source-dependencies/packageobjects-and-traits` -> Looks like an uncovered IO error. To double-check.

jvican:
@eed3si9n In one of my last comments, I mention one example in which our less precise way of testing integration tests via high-level tasks like `compile` bites us. I've experienced other similar cases personally. This is just a data point.

eed3si9n:
@jvican Cool. Thanks for pointing it out. These data points would indeed help in future discussions.

dwijnand:
I'll close this for now (just to be sure not to lose the effort here I've cross-linked it in #367). Revival welcome!

PR #374, opened 2017-07-20
Fix #355: Handle malformed class name

### LINKED ISSUE #355: Java analysis extraction fails to `getCanonicalClassname` for nested object
While attempting to compile Java code that depends on `com.twitter.io.Buf.Composite.Impl` ([defined here](https://github.com/twitter/util/blob/44128deb7ad989be36f825612cc958599cfe3af7/util-core/src/main/scala/com/twitter/io/Buf.scala#L475)), zinc crashes with a stack like the following:
```
[error] Malformed class name
[error] java.lang.Class.getSimpleName(Class.java:1330)
[error] java.lang.Class.getCanonicalName(Class.java:1399)
[error] sbt.internal.inc.ClassToAPI$.classCanonicalName(ClassToAPI.scala:77)
[error] sbt.internal.inc.ClassToAPI$.toDefinitions(ClassToAPI.scala:80)
[error] sbt.internal.inc.ClassToAPI$.$anonfun$structure$5(ClassToAPI.scala:148)
[error] scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:241)
[error] scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:32)
[error] scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:29)
[error] scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:38)
[error] scala.collection.TraversableLike.flatMap(TraversableLike.scala:241)
[error] scala.collection.TraversableLike.flatMap$(TraversableLike.scala:238)
[error] scala.collection.AbstractTraversable.flatMap(Traversable.scala:104)
[error] sbt.internal.inc.ClassToAPI$.merge(ClassToAPI.scala:369)
[error] sbt.internal.inc.ClassToAPI$.structure(ClassToAPI.scala:150)
[error] sbt.internal.inc.ClassToAPI$.x$2$lzycompute$1(ClassToAPI.scala:92)
[error] sbt.internal.inc.ClassToAPI$.x$2$1(ClassToAPI.scala:92)
[error] sbt.internal.inc.ClassToAPI$.instance$lzycompute$1(ClassToAPI.scala:92)
[error] sbt.internal.inc.ClassToAPI$.instance$1(ClassToAPI.scala:92)
[error] sbt.internal.inc.ClassToAPI$.$anonfun$toDefinitions0$1(ClassToAPI.scala:99)
[error] xsbti.api.SafeLazyProxy$$anon$1.apply(SafeLazyProxy.scala:24)
[error] xsbti.api.SafeLazy$Impl.get(SafeLazy.java:55)
[error] sbt.internal.inc.ClassToAPI$.$anonfun$process$2(ClassToAPI.scala:29)
[error] scala.collection.immutable.List.foreach(List.scala:389)
[error] scala.collection.generic.TraversableForwarder.foreach(TraversableForwarder.scala:35)
[error] scala.collection.generic.TraversableForwarder.foreach$(TraversableForwarder.scala:35)
[error] scala.collection.mutable.ListBuffer.foreach(ListBuffer.scala:44)
[error] sbt.internal.inc.ClassToAPI$.process(ClassToAPI.scala:29)
[error] sbt.internal.inc.javac.AnalyzingJavaCompiler.readAPI$1(AnalyzingJavaCompiler.scala:142)
[error] sbt.internal.inc.javac.AnalyzingJavaCompiler.$anonfun$compile$20(AnalyzingJavaCompiler.scala:160)
[error] sbt.internal.inc.classfile.Analyze$.readInheritanceDependencies$1(Analyze.scala:134)
[error] sbt.internal.inc.classfile.Analyze$.$anonfun$apply$16(Analyze.scala:150)
[error] sbt.internal.inc.classfile.Analyze$.$anonfun$apply$16$adapted(Analyze.scala:67)
[error] scala.collection.TraversableLike$WithFilter.$anonfun$foreach$1(TraversableLike.scala:789)
[error] scala.collection.mutable.HashMap.$anonfun$foreach$1(HashMap.scala:138)
[error] scala.collection.mutable.HashTable.foreachEntry(HashTable.scala:236)
[error] scala.collection.mutable.HashTable.foreachEntry$(HashTable.scala:229)
[error] scala.collection.mutable.HashMap.foreachEntry(HashMap.scala:40)
[error] scala.collection.mutable.HashMap.foreach(HashMap.scala:138)
[error] scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:788)
[error] sbt.internal.inc.classfile.Analyze$.apply(Analyze.scala:67)
[error] sbt.internal.inc.javac.AnalyzingJavaCompiler.$anonfun$compile$19(AnalyzingJavaCompiler.scala:160)
[error] sbt.internal.inc.javac.AnalyzingJavaCompiler.$anonfun$compile$19$adapted(AnalyzingJavaCompiler.scala:158)
[error] scala.collection.TraversableLike$WithFilter.$anonfun$foreach$1(TraversableLike.scala:789)
[error] scala.collection.immutable.List.foreach(List.scala:389)
[error] scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:788)
[error] sbt.internal.inc.javac.AnalyzingJavaCompiler.$anonfun$compile$17(AnalyzingJavaCompiler.scala:158)
[error] scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
[error] sbt.internal.inc.javac.AnalyzingJavaCompiler.timed(AnalyzingJavaCompiler.scala:195)
[error] sbt.internal.inc.javac.AnalyzingJavaCompiler.compile(AnalyzingJavaCompiler.scala:158)
[error] sbt.internal.inc.MixedAnalyzingCompiler.$anonfun$compile$4(MixedAnalyzingCompiler.scala:104)
[error] scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
[error] sbt.internal.inc.MixedAnalyzingCompiler.timed(MixedAnalyzingCompiler.scala:132)
[error] sbt.internal.inc.MixedAnalyzingCompiler.compileJava$1(MixedAnalyzingCompiler.scala:89)
[error] sbt.internal.inc.MixedAnalyzingCompiler.compile(MixedAnalyzingCompiler.scala:115)
[error] sbt.internal.inc.IncrementalCompilerImpl.$anonfun$compileInternal$1(IncrementalCompilerImpl.scala:304)
[error] sbt.internal.inc.IncrementalCompilerImpl.$anonfun$compileInternal$1$adapted(IncrementalCompilerImpl.scala:304)
[error] sbt.internal.inc.Incremental$.doCompile(Incremental.scala:101)
[error] sbt.internal.inc.Incremental$.$anonfun$compile$4(Incremental.scala:82)
[error] sbt.internal.inc.IncrementalCommon.recompileClasses(IncrementalCommon.scala:118)
[error] sbt.internal.inc.IncrementalCommon.cycle(IncrementalCommon.scala:64)
[error] sbt.internal.inc.Incremental$.$anonfun$compile$3(Incremental.scala:84)
[error] sbt.internal.inc.Incremental$.manageClassfiles(Incremental.scala:129)
[error] sbt.internal.inc.Incremental$.compile(Incremental.scala:75)
[error] sbt.internal.inc.IncrementalCompile$.apply(Compile.scala:70)
[error] sbt.internal.inc.IncrementalCompilerImpl.compileInternal(IncrementalCompilerImpl.scala:308)
[error] sbt.internal.inc.IncrementalCompilerImpl.$anonfun$compileIncrementally$1(IncrementalCompilerImpl.scala:266)
[error] sbt.internal.inc.IncrementalCompilerImpl.handleCompilationError(IncrementalCompilerImpl.scala:157)
[error] sbt.internal.inc.IncrementalCompilerImpl.compileIncrementally(IncrementalCompilerImpl.scala:236)
[error] sbt.internal.inc.IncrementalCompilerImpl.compile(IncrementalCompilerImpl.scala:67)
[error] org.pantsbuild.zinc.compiler.Main$.main(Main.scala:109)
[error] org.pantsbuild.zinc.compiler.Main.main(Main.scala)
```

~~This _may_ be a red herring, because I was working in a branch that combines both https://github.com/sbt/zinc/pull/325 and https://github.com/sbt/zinc/pull/351 (a prerelease version: [1.0.0-X19-M1](https://bintray.com/scalacenter/releases/zinc/1.0.0-X19-M1))~~ EDIT: Confirm on `X20` as well, but the location of the crash looks completely unrelated to those changes, so thought it would be worth posting now.

jvican:
The binary name encoding for Scala has been broken for a while. It does
not follow the Java source name rules because it does not append a `$`
after the name of an object, which already ends in `$`.

Zinc works around this fix in the only way it can: it gets the enclosing
class, which is obtained from the `InnerClasses` java classfile
metadata, and it reconstructs the path from there recursively. It
circumvents, by design, the core issue: it doesn't call Java's name
parser at all.

This issue has been fixed in the JDK9, but unfortunately it's going to
torment us for a while because 2.13 does not target Java 9, and
we're still at 2.12.x.

This is the fix in JDK9:
http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8057919

In order to fully fix this issue, the Scala compiler should do the
desired behaviour in 2.13.x. 2.12.x is already lost since this fix is
binary incompatible. The appropriate JIRA ticket is here:
scala/bug#2034.


jvican:
To all reviewers (@dwijnand, @eed3si9n, @Duhemm): this is a blocker for the next RC, can you please have a look at this self-contained change? I would like to get this thing in as soon as possible.

dwijnand:
Could you add a unit test for the full spectrum of combinations as in SI-2034?

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -73,8 +73,50 @@ object ClassToAPI {
...
+            if malformedError.getMessage.contains("Malformed class name") =>
+          val enclosingClass = c.getEnclosingClass
+          val enclosingName = enclosingClass.getName
+          val restOfName = c.getName.stripPrefix(enclosingName)
+          handleMalformedNameOf(enclosingClass, true) + restOfName + atEnd
+      }
+    }
+  }
+
+  def classCanonicalName(c: Class[_]): String = handleMalformedNameOf(c)
```
Duhemm, line 118: Considering this:

A.scala:
```scala
object A {
  def foo = B.C$.x
}
```

B.java:
```java
public class B {
    public static class C$ {
        public static int x = 42;
    }
}
```

The compilation will fail with this patch applied:

```
Caused by: java.lang.RuntimeException: Failed to find name hashes for B.C$
        at scala.sys.package$.error(package.scala:27)
        at sbt.internal.inc.AnalysisCallback.nameHashesForCompanions(Compile.scala:331)
        at sbt.internal.inc.AnalysisCallback.analyzeClass(Compile.scala:338)
        at sbt.internal.inc.AnalysisCallback.$anonfun$addProductsAndDeps$4(Compile.scala:354)
        at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
        at scala.collection.mutable.HashSet.foreach(HashSet.scala:78)
        at scala.collection.TraversableLike.map(TraversableLike.scala:234)
        at scala.collection.TraversableLike.map$(TraversableLike.scala:227)
        at scala.collection.mutable.AbstractSet.scala$collection$SetLike$$super$map(Set.scala:46)
        at scala.collection.SetLike.map(SetLike.scala:101)
        at scala.collection.SetLike.map$(SetLike.scala:101)
        at scala.collection.mutable.AbstractSet.map(Set.scala:46)
        at sbt.internal.inc.AnalysisCallback.$anonfun$addProductsAndDeps$1(Compile.scala:354)
        at scala.collection.TraversableOnce.$anonfun$foldLeft$1(TraversableOnce.scala:157)
        at scala.collection.TraversableOnce.$anonfun$foldLeft$1$adapted(TraversableOnce.scala:157)
        at scala.collection.mutable.HashSet.foreach(HashSet.scala:78)
        at scala.collection.TraversableOnce.foldLeft(TraversableOnce.scala:157)
        at scala.collection.TraversableOnce.foldLeft$(TraversableOnce.scala:155)
        at scala.collection.AbstractTraversable.foldLeft(Traversable.scala:104)
        at scala.collection.TraversableOnce.$div$colon(TraversableOnce.scala:151)
        at scala.collection.TraversableOnce.$div$colon$(TraversableOnce.scala:151)
        at scala.collection.AbstractTraversable.$div$colon(Traversable.scala:104)
        at sbt.internal.inc.AnalysisCallback.addProductsAndDeps(Compile.scala:350)
        at sbt.internal.inc.AnalysisCallback.get(Compile.scala:300)
        at sbt.internal.inc.Incremental$.doCompile(Incremental.scala:102)
        at sbt.internal.inc.Incremental$.$anonfun$compile$4(Incremental.scala:82)
        at sbt.internal.inc.IncrementalCommon.recompileClasses(IncrementalCommon.scala:118)
        at sbt.internal.inc.IncrementalCommon.cycle(IncrementalCommon.scala:64)
        at sbt.internal.inc.Incremental$.$anonfun$compile$3(Incremental.scala:84)
        at sbt.internal.inc.Incremental$.manageClassfiles(Incremental.scala:129)
        at sbt.internal.inc.Incremental$.compile(Incremental.scala:75)
        at sbt.internal.inc.IncrementalCompile$.apply(Compile.scala:70)
        at sbt.internal.inc.IncrementalCompilerImpl.compileInternal(IncrementalCompilerImpl.scala:309)
        at sbt.internal.inc.IncrementalCompilerImpl.$anonfun$compileIncrementally$1(IncrementalCompilerImpl.scala:267)
        at sbt.internal.inc.IncrementalCompilerImpl.handleCompilationError(IncrementalCompilerImpl.scala:158)
        at sbt.internal.inc.IncrementalCompilerImpl.compileIncrementally(IncrementalCompilerImpl.scala:237)
        at sbt.internal.inc.IncrementalCompilerImpl.compile(IncrementalCompilerImpl.scala:68)
        at sbt.internal.inc.ProjectStructure.compile(IncHandler.scala:386)
        at sbt.internal.inc.IncHandler$$anonfun$commands$lzycompute$1.applyOrElse(IncHandler.scala:133)
        at sbt.internal.inc.IncHandler$$anonfun$commands$lzycompute$1.applyOrElse(IncHandler.scala:133)
        at sbt.internal.inc.IncHandler.applyOrElse(IncHandler.scala:197)
        at sbt.internal.inc.IncHandler.$anonfun$noArgs$1(IncHandler.scala:195)
        at sbt.internal.inc.IncHandler.$anonfun$noArgs$1$adapted(IncHandler.scala:195)
        at sbt.internal.inc.IncHandler.$anonfun$apply$1(IncHandler.scala:97)
        at sbt.internal.inc.IncHandler.$anonfun$apply$1$adapted(IncHandler.scala:97)
        at sbt.internal.inc.IncHandler.onIncInstance(IncHandler.scala:104)
        at sbt.internal.inc.IncHandler.apply(IncHandler.scala:98)
        at sbt.internal.inc.IncHandler.apply(IncHandler.scala:52)
        at sbt.internal.scripted.ScriptRunner.processStatement$1(ScriptRunner.scala:18)
        ... 52 more
```

Duhemm (dismissed):
See my comment.

Maybe that `xsbt.ClassName.flatclassName` must be fixed too so that their results match?

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: This case is handled automatically by the spec of both Scala and Java: you cannot use `$` in names neither in Java nor Scala:
* [Scala spec](http://www.scala-lang.org/files/archive/spec/2.12/01-lexical-syntax.html)
* [Java spec](https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.8)

jvican: In fact, note that this patch does the right thing: it preserves the ending '$' of the original 'B.C$', where a low tech solution wouldn't have handled that :smile:.

Duhemm: From the Java spec you linked:
> The $ sign should be used only in mechanically generated source code or, rarely, to access pre-existing names on legacy systems.

My reading is that it's perfectly legal for a source code generator to emit names that use the $ sign.

The Scala spec is more ambiguous. However, in the above example, changing `def foo = B.C$.x` to `def foo = new B` produces the same crash.

Duhemm (dismissed):
I guess that Java sources with `$` in identifiers are less frequent than doubly nested classes, so this PR is a win. 
Could you please add that test as a pending scripted test?

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: Yes, code generators can, but it's not clear to me whether the `$` in user-defined names is escaped or not. I'm unsure at this point, I have to check.
  
Yes, `def foo = new B` clashes because those classes with `$` are referenced in the constant pool. Let me have a closer look to whether this misbehavior can (or should) be fixed.

jvican:
@dwijnand The tests you asked me to add don't even compile:
```scala
> $ scalac SI2038.scala 
SI2038.scala:4: error: name clash: class o2 defines object o3
and its companion object o2 also defines class o3
    class o2  { object o3; class o3 }
                       ^
SI2038.scala:4: error: name clash: class o2 defines class o3
and its companion object o2 also defines class o3
    class o2  { object o3; class o3 }
                                 ^
SI2038.scala:7: error: name clash: class o1 defines object o2
and its companion object o1 also defines class o2
    object o2 { object o3; class o3 }
           ^
SI2038.scala:8: error: name clash: class o1 defines class o2
and its companion object o1 also defines class o2
    class o2  { object o3; class o3 }
          ^
SI2038.scala:8: error: name clash: class o2 defines object o3
and its companion object o2 also defines class o3
    class o2  { object o3; class o3 }
                       ^
SI2038.scala:8: error: name clash: class o2 defines class o3
and its companion object o2 also defines class o3
    class o2  { object o3; class o3 }
                                 ^
6 errors found
```
  
The source file to compile is:
```scala
package s {
  object o1 {
    object o2 { object o3; class o3 }
    class o2  { object o3; class o3 }
  }
  class o1 {
    object o2 { object o3; class o3 }
    class o2  { object o3; class o3 }
  }
}
```
  
My conclusion: for some reason, scalac stopped compiling this code and I don't find any way to test what Paul had in mind back in the day because the key is that classes are linked to their companion (module classes) in their constant pool. If we change the naming of any of those names, like this:
```scala
package s {
  object o11 {
    object o22 { object o3; class o3 }
    class o2  { object o3; class o3 }
  }
  class o1 {
    object o22 { object o3; class o3 }
    class o2  { object o3; class o3 }
  }
}
```
  
We won't be able to test the objects because, for instance, `o22` won't be a companion of `o2` anymore. So unless you have another idea, adding these tests won't bring anything to the table.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: @Duhemm I've had a deeper look. Are you sure that this case was not failing before my patch? I'm able to reproduce, but the case fails when compiling the java source first, which does not use `ClassToAPI`. It fails because of https://github.com/sbt/zinc/issues/292.

jvican: For the record, I'm intending to fix this issue. Though if this issue has no relation with this change, i would like to fix it in an independent PR.

Duhemm: @jvican See this scripted test: https://github.com/Duhemm/incrementalcompiler/commit/76bce1d1a10bc9957a84d337f18ce71599731b5d

Duhemm (dismissed):
Thanks for adding the test!

LGTM 👍

dwijnand:
Oh yeah, d'oh, that code never compiled in any released version of Scala: https://github.com/scala/bug/issues/2034#issuecomment-292367544.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: Okay, so for some reason Zinc was not showing the double `$` that triggered this issue:
```
Caused by: java.lang.RuntimeException: Failed to find name hashes for B.C$
```
  
Last commit fixes it. I've squashed it with the previous "Fix #..." commit.

jvican:
I think I've addressed all the feedback when it was possible. Can you guys have a look again? @dwijnand @Duhemm

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -73,8 +73,49 @@ object ClassToAPI {
...
+   * in 2.13.x. 2.12.x is already lost since this fix is binary incompatible. The appropriate
+   * JIRA ticket is here: https://github.com/scala/bug/issues/2034.
+   *
+   * @return The canonical name if not null, the full name otherwise.
+   */
+  def handleMalformedNameOf(c: Class[_], isRecursive: Boolean = false): String = {
+    if (c == null) "" // Return nothing if it hits the top-level class
+    else {
+      val className = c.getName
+      // Adds a `dollar` if it's a recursive call or if the original class is an object
```
Duhemm: That comment is outdated, isn't it?

jvican:
@dwijnand I hope we're able to fix this in 2.13.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -73,8 +73,49 @@ object ClassToAPI {
...
+   * <a href="http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8057919">This is the fix in Java 9.</a>
+   *
+   * In order to fully fix this issue, the Scala compiler should do the desired behaviour
+   * in 2.13.x. 2.12.x is already lost since this fix is binary incompatible. The appropriate
+   * JIRA ticket is here: https://github.com/scala/bug/issues/2034.
+   *
+   * @return The canonical name if not null, the full name otherwise.
+   */
+  def handleMalformedNameOf(c: Class[_], isRecursive: Boolean = false): String = {
+    if (c == null) "" // Return nothing if it hits the top-level class
```
eed3si9n, line 98: Either use `Option(c)` or `c eq null`?

Duhemm (dismissed):
Not sure if the comment needs to be updated.

If it doesn't, then LGTM!

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: It does need to be updated.

jvican: Outdated. Review again :wink:.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -73,8 +73,49 @@ object ClassToAPI {
...
+    else {
+      val className = c.getName
+      // Adds a `dollar` if it's a recursive call, else nothing
+      val atEnd: String = if (isRecursive) "$" else ""
+      try {
+        val canonicalName = c.getCanonicalName
+        if (canonicalName == null) className
```
eed3si9n, line 105: Same here. `eq` or `Option`.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -73,8 +73,49 @@ object ClassToAPI {
...
+            if malformedError.getMessage.contains("Malformed class name") =>
+          val enclosingClass = c.getEnclosingClass
+          val enclosingName = enclosingClass.getName
+          val restOfName = c.getName.stripPrefix(enclosingName)
+          handleMalformedNameOf(enclosingClass, true) + restOfName + atEnd
+      }
+    }
+  }
+
+  def classCanonicalName(c: Class[_]): String = handleMalformedNameOf(c)
```
eed3si9n, line 118: Is it possible to write unit that checks the expected output for the nested classes etc?

eed3si9n (changes requested):

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: Unit tests only check java and scala compilation independently. For this issue to be reproducible, it has to be produced by Scala compilation followed by Java compilation. This is exactly what scripted does, and it's the reason why it's an integration test.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: There's no point in doing it, the Scala compiler already does it for us. Also, this is the way it's being written across all the `ClassToAPI`, so I prefer to keep it consistent.

jvican: Proof. Compile `if (a == null) println("BLA") else println("BLO")` and print the tree. You'll see:
```
    def <init>($outer: B.type): B$delayedInit$body = {
      if ($outer.eq(null))
        throw null
      else
        B$delayedInit$body.this.$outer = $outer;
      B$delayedInit$body.super.<init>();
      ()
    }
```

eed3si9n (approved):

eed3si9n:
LGTM

dwijnand:
2 approvals

jvican:
Thanks for the prompt review guys.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: Proof at the bytecode level (see ifnotnull):
```
    Code:
      stack=2, locals=1, args_size=1
         0: aload_0
         1: ldc           #71                 // String
         3: putfield      #66                 // Field a:Ljava/lang/String;
         6: aload_0
         7: invokevirtual #73                 // Method a:()Ljava/lang/String;
        10: ifnonnull     24
        13: getstatic     #78                 // Field scala/Predef$.MODULE$:Lscala/Predef$;
        16: ldc           #80                 // String HEY
        18: invokevirtual #84                 // Method scala/Predef$.println:(Ljava/lang/Object;)V
        21: goto          32
        24: getstatic     #78                 // Field scala/Predef$.MODULE$:Lscala/Predef$;
        27: ldc           #86                 // String HOY
        29: invokevirtual #84                 // Method scala/Predef$.println:(Ljava/lang/Object;)V
        32: return
```
  
Compiling with `eq null` produces the same bytecode (`ifnonnull`).

Issue #375, opened 2017-07-21
Don't publish the benchmark project


xuwei-k:
http://repo1.maven.org/maven2/org/scala-sbt/benchmarks-of-zinc-and-the-compiler-bridge_2.12/

jvican:
Mind sending a fix?

jvican:
I'm surprised this is happening because `zincBenchmarks` already has:
```scala
    publish := {},
    publishLocal := {}
```
  
Anyway, I'm adding a fix to this on top of https://github.com/sbt/zinc/pull/377.

PR #377, opened 2017-07-25
Fix #263 and minor build errors

### LINKED ISSUE #263: Nuke the compiler bridge cache when we publish the bridge
Copy from dotty:

```scala
      cleanSbtBridge := {
        val dottySbtBridgeVersion = version.value
        val dottyVersion = (version in `dotty-compiler`).value
        val classVersion = System.getProperty("java.class.version")

        val sbtV = sbtVersion.value
        val sbtOrg = "org.scala-sbt"
        val sbtScalaVersion = "2.10.6"

        val home = System.getProperty("user.home")
        val org = organization.value
        val artifact = moduleName.value

        IO.delete(file(home) / ".ivy2" / "cache" / sbtOrg / s"$org-$artifact-$dottySbtBridgeVersion-bin_${dottyVersion}__$classVersion")
        IO.delete(file(home) / ".sbt"  / "boot" / s"scala-$sbtScalaVersion" / sbtOrg / "sbt" / sbtV / s"$org-$artifact-$dottySbtBridgeVersion-bin_${dottyVersion}__$classVersion")
      },
      publishLocal := (publishLocal.dependsOn(cleanSbtBridge)).value,
```

https://github.com/lampepfl/dotty/blob/c84480960cf618c29705dbaab9332d304a081524/project/Build.scala#L534-L550

Idea from @smarter.

jvican:
This does three simple things:

* Updates the version.
* Nukes the compiler bridge caches.
* Formats the build sources.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -316,6 +316,15 @@ lazy val compilerInterface = (project in internalPath / "compiler-interface")
   )
   .configure(addSbtUtilInterface)
 
+val cleanSbtBridge = taskKey[Unit]("Cleans the sbt bridge.")
+
+def wrapIn(color: String, content: String): String = {
+  import sbt.internal.util.ConsoleAppender
+  if (!ConsoleAppender.formatEnabled) content
+  else color + content + scala.Console.RESET
+}
+
+
 // Compiler-side interface to compiler that is compiled against the compiler being used either in advance or on the fly.
 //   Includes API and Analyzer phases that extract source API and relationships.
 lazy val compilerBridge: Project = (project in internalPath / "compiler-bridge")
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -89,8 +89,17 @@ def altPublishSettings: Seq[Setting[_]] =
     }
   )
 
+val noPublish: Seq[Setting[_]] = List(
+  publish := {},
+  publishLocal := {},
+  publishArtifact in Compile := false,
```
dwijnand, line 95: You probably need to redefine `publishSigned` to fix #375.

dwijnand (approved):

build.sbt
eed3si9n: publishSigned respects `skip in publish` so that's going to be the fix I think.

build.sbt
```diff
@@ -89,8 +89,17 @@ def altPublishSettings: Seq[Setting[_]] =
...
+  publishArtifact in Test := false,
+  publishArtifact := false,
+  skip in publish := true,
```
eed3si9n, line 98: 👍

eed3si9n (approved):

Duhemm (approved):
LGTM

build.sbt
jvican: Can we actually remove the `publish*` and `publishArtifact` redefinitions if `skip in publish` is already being respected?

PR #373, opened 2017-07-20
Use latest sbt 1.0.0-RC2

dwijnand:
Fixes #372

PR #378, opened 2017-07-26
Use Scala 2.12.3

jvican:
Follows up on https://github.com/sbt/zinc/pull/373. Thanks to Dale for his initial work.
  
* Upgrades to RC2.
* Uses Scala 2.12.3.
* Adds and configures MiMa.


jvican:
It seems the protoc plugin is not working for the following cases:
1. https://github.com/scalapb/ScalaPB/issues/314
2. https://github.com/scalapb/ScalaPB/issues/313
  
This PR is therefore conservative until we do solve this situation:
1. It does not upgrade the protoc plugin.
2. It does not use the compiler shaded version (**IMPORTANT**).
  
Review by @dwijnand and @Duhemm.

jvican:
Note that this adds sbt-mima.

jvican:
This PR is failing in the Ci because 2.12.3 is not yet released to Central. When it is, the CI will be run and if successful can be merged.

.drone.yml.sig
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+eyJhbGciOiJIUzI1NiJ9.cGlwZWxpbmU6CiAgIyBGZXRjaCBmb2xkZXJzIGZyb20gZGlzdHJpYnV0ZWQgY2FjaGUKICBzZnRwX2NhY2hlX3Jlc3RvcmU6CiAgICBpbWFnZTogcGx1Z2lucy9zZnRwLWNhY2hlCiAgICByZXN0b3JlOiB0cnVlCiAgICBtb3VudDoKICAgICAgLSAvZHJvbmUvLml2eTIKICAgICAgLSAvZHJvbmUvLmNvdXJzaWVyLWNhY2hlCiAgICAgIC0gL2Ryb25lLy5zYnQKICAgICAgLSAvZHJvbmUvLmdpdAogICAgd2hlbjoKICAgICAgZXZlbnQ6IFtwdXNoLCBwdWxsX3JlcXVlc3QsIHRhZywgZGVwbG95bWVudF0KCiAgdGVzdHM6CiAgICBpbWFnZTogc2NhbGFjZW50ZXIvc2NhbGE6MS4wCiAgICBjb21tYW5kczoKICAgICAgLSBnaXQgZmV0Y2ggLS10YWdzICYmIGdpdCBsb2cgfCBoZWFkIC1uIDIwCiAgICAgIC0gLi9iaW4vcnVuLWNpLnNoICR7Q0lfU0NBTEFfVkVSU0lPTn0KCiAgIyBTYXZlIGZvbGRlcnMgaW4gZGlzdHJpYnV0ZWQgY2FjaGUKICBzZnRwX2NhY2hlX3JlYnVpbGQ6CiAgICBpbWFnZTogcGx1Z2lucy9zZnRwLWNhY2hlCiAgICByZWJ1aWxkOiB0cnVlCiAgICBtb3VudDoKICAgICAgLSAvZHJvbmUvLml2eTIKICAgICAgLSAvZHJvbmUvLmNvdXJzaWVyLWNhY2hlCiAgICAgIC0gL2Ryb25lLy5zYnQKICAgICAgLSAvZHJvbmUvLmdpdAogICAgd2hlbjoKICAgICAgZXZlbnQ6IFtwdXNoLCBwdWxsX3JlcXVlc3QsIHRhZywgZGVwbG95bWVudF0KCm1hdHJpeDoKICBDSV9TQ0FMQV9WRVJTSU9OOgogICAgLSAyLjExLjExCiAgICAtIDIuMTIuMwo.usuc6GVWQgjD1XYQW9Fd8U8N7Qj1b6WfmGdtIn44Q-8
```
dwijnand, line 1: What does this do and why did it change?

build.sbt
```diff
@@ -1,18 +1,26 @@
...
+    if (crossPaths.value) CrossVersion.binary else CrossVersion.disabled)
+  )
+)
+
 def commonSettings: Seq[Setting[_]] = Seq(
   scalaVersion := scala212,
   // publishArtifact in packageDoc := false,
   resolvers += Resolver.typesafeIvyRepo("releases"),
   resolvers += Resolver.sonatypeRepo("snapshots"),
+  resolvers += Resolver.sonatypeRepo("staging"),
```
dwijnand: I'd prefer not adding another resolver - we already have more than we should.

dwijnand (changes requested):
There's no enforcement of MiMa in CI

build.sbt
jvican: Agreed, should be removed. It slipped through the cracks.

.drone.yml.sig
jvican: This is necessary every time the drone.yml is modified, so that attackers cannot say what your Ci does. It's a security measure.

build.sbt
jvican: Oh, no, actually this is necessary for now until Scala 2.12.3 has hit Central. I will remove it then.

jvican:
Feedback addressed.

.drone.yml.sig
dwijnand: How do I generate it?

bin/run-ci.sh
```diff
@@ -3,12 +3,16 @@ set -eu
...
+# scalafmt::test \
+# test:scalafmt::test \
+
 sbt -Dfile.encoding=UTF-8 \
   -J-XX:ReservedCodeCacheSize=256M \
   -J-Xmx3046M -J-Xms3046M -J-server \
   zincRoot/test:compile \
-  scalafmt::test \
-  test:scalafmt::test \
+  mimaReportBinaryIssues \
```
dwijnand: How does this validate our 2.10 artifacts?

jvican: Good catch, forgot to `+`.

jvican: Check if my solution works.

.drone.yml.sig
jvican: From the `drone` command line application. It's in the docs, v0.5.

jvican:
Rebased on top of #379, it only adds the 2.12.3 version.

dwijnand (approved):
Ah, the test is failing because Scala 2.12.3 isn't in Maven Central yet. So we can re-run this when it is to get it green and merge it.

So, LGTM. Thanks Jorge.

eed3si9n (approved):

PR #379, opened 2017-07-27
Upgrade to sbt 1 & add, configure & enable MiMa

dwijnand:
Re-take on #378, which looks to be running into an issue with also trying to upgrade to Scala 2.12.3.

I **really** want to have MiMa up and running on the 1.0.0 branch to check and enforce our binary API, so I de-scoped the Scala upgrade for right now. I also cleaned up the commits - thank you Jorge for helping figure out WTH was going on with sbt-protoc/scalapb/protobuf.

### DIFF
--- a/bin/run-ci.sh
+++ b/bin/run-ci.sh
@@ -3,12 +3,15 @@ set -eu
 set -o nounset
 SCALA_VERSION="$1"
 
+# drop scalafmt on the 1.0.0 branch to dogfood 1.0.0-RC2 before there is a sbt 1.0 of new-sbt-scalafnt
+#  see https://github.com/lucidsoftware/neo-sbt-scalafmt/pull/34
+# scalafmt::test \
+# test:scalafmt::test \
+
 sbt -Dfile.encoding=UTF-8 \
   -J-XX:ReservedCodeCacheSize=256M \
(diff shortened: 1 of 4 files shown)


bin/run-ci.sh
```diff
@@ -6,9 +6,12 @@ SCALA_VERSION="$1"
 sbt -Dfile.encoding=UTF-8 \
   -J-XX:ReservedCodeCacheSize=256M \
   -J-Xmx3046M -J-Xms3046M -J-server \
+  +mimaReportBinaryIssues \
   zincRoot/test:compile \
-  scalafmt::test \
-  test:scalafmt::test \
+# drop scalafmt on the 1.0.0 branch to dogfood 1.0.0-RC2 before there is a sbt 1.0 of new-sbt-scalafnt
```
jvican: This doesn't work, comments make the script non-executable and `crossTestBridges` is not found. You have to move them out of the commands section.

dwijnand: Oh wow! Interesting. I didn't realise bash interpreted that way. Thank you.

jvican (approved):
Thanks for getting this thing to the finish line. Let's merge and I'll rebase my other PR.

PR #380, opened 2017-07-27
ClassToAPI.classCanonicalName spec

dwijnand:

PR #381, opened 2017-07-28
Tweak ClassToAPI.classCanonicalName

dwijnand:
Builds on the spec defined in #380, proposing a change in behaviour.

### DIFF
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -73,49 +73,45 @@ object ClassToAPI {
                  new mutable.HashSet,
                  new mutable.HashSet)
 
+  @deprecated("No longer used. Use classCanonicalName instead.", "1.0.0")
+  def handleMalformedNameOf(
+      c: Class[_],
+      @deprecated("unused", "1.0.0") isRecursive: Boolean = false
+  ): String =
+    classCanonicalName(c)
+
   /**
-   * The binary name encoding for Scala has been broken for a while. It does not
-   * follow the Java source name rules because it does not append a `$` after the
-   * mirror class of an object, which already ends in `$`.
-   *
-   * Zinc works around this fix in the only way it can: it gets the enclosing class,
-   * which is obtained from the `InnerClasses` java classfile metadata, and it
-   * reconstructs the path from there. It circumvents, by design, the core issue:
-   * it doesn't call Java's name parser at all.
-   *
-   * This issue has been fixed in the JDK9, but unfortunately it's going to be tormenting
-   * us for a while because 2.13 does not target Java 9, and we're still at 2.12.x.
-   *
-   * <a href="http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8057919">This is the fix in Java 9.</a>
-   *
-   * In order to fully fix this issue, the Scala compiler should do the desired behaviour
-   * in 2.13.x. 2.12.x is already lost since this fix is binary incompatible. The appropriate
-   * issue is here: https://github.com/scala/bug/issues/2034.
-   *
-   * @return The canonical name if not null, the blank string otherwise.
+   * Returns a string to identify the class. Ideally the class's canonical name, but if that's
+   * not available (e.g it's `null` or an error is thrown) then the full qualified name is returned.
+   * Returns the empty string if the class is `null`.
    */
-  def handleMalformedNameOf(c: Class[_], isRecursive: Boolean = false): String = {
-    if (c == null) "" // Return nothing if it hits the top-level class
-    else {
-      val className = c.getName
-      // Adds a `dollar` if it's a recursive call, else nothing
-      val atEnd: String = if (isRecursive) "$" else ""
+  def classCanonicalName(c: Class[_]): String = {
+    if (c == null) ""
+    else
       try {
         val canonicalName = c.getCanonicalName
-        if (canonicalName == null) className
-        else canonicalName + atEnd
+        if (canonicalName == null) c.getName
+        else canonicalName
       } catch {
-        case malformedError: java.lang.InternalError
-            if malformedError.getMessage.contains("Malformed class name") =>
-          val enclosingClass = c.getEnclosingClass
-          val enclosingName = enclosingClass.getName
-          val restOfName = c.getName.stripPrefix(enclosingName)
-          handleMalformedNameOf(enclosingClass, true) + restOfName + atEnd
+        case e: InternalError if e.getMessage == "Malformed class name" =>
+          // Class#getCanonicalNames use Class#getSimpleName, and Class#getSimpleName's behaviour
+          // relies on JLS 13.1 "The Form of a Binary":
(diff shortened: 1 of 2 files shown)


internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -90,33 +90,29 @@ object ClassToAPI {
...
-        else canonicalName + atEnd
+        else canonicalName
       } catch {
         case malformedError: java.lang.InternalError
             if malformedError.getMessage.contains("Malformed class name") =>
           val enclosingClass = c.getEnclosingClass
           val enclosingName = enclosingClass.getName
           val restOfName = c.getName.stripPrefix(enclosingName)
-          handleMalformedNameOf(enclosingClass, true) + restOfName + atEnd
+          classCanonicalName(enclosingClass).stripSuffix("$") + "." + restOfName
```
jvican: There's a strong difference between my previous PR and yours. Mine was not pretending that the borked binary names had `$$`, it was instead using the original `$` separation to have a consistent way to represent these borked names.
  
The patch that you propose does pretend that the borked binary names are valid. There's nothing wrong with this a priori, but needs a deeper study.
  
For example, what happens if there indeed exists a valid binary name `A$$B$` (it could come from a future version of Scala or another JVM-based programming language)? IT means we'll have a duplicated representation for the borked and the non-borked one.

internal/zinc-apiinfo/src/test/scala/sbt/internal/inc/ClassCanonicalNameSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+// C-C-C means class { class { class } }
+// etc..
+// Because of scala/bug#2034 we can't put these all in one package (you get "name clash" errors)
+// So instead we'll split them in 4 nice and even packages
+package p1 { object x { object y { object z; class z } } }
+package p2 { object x { class y { object z; class z } } }
+package p3 { class x { object y { object z; class z } } }
+package p4 { class x { class y { object z; class z } } }
+
+class ClassCanonicalNameSpec extends UnitSpec {
```
jvican: The test suite is excellent, as I said before.

jvican (changes requested):
One important question.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
dwijnand: By "duplicate representation" do you mean name-clashing for distinct classes?

jvican: Yes.

dwijnand: Then your concern is of the status quo, not specifically this proposed change. Because currently (as demonstrated in the spec as of #380) you can't tell the difference between:

+ OO and CO: both "x.y$"
+ OC and CC: both "x.y"
+ OCO and CCO: both "x.y.z$"
+ OCC and CCC: both "x.y.z"

and btw, these are coming direct from getCanonicalClass.

If Zinc depends on the invariant that distinct classes have distinct names (as returned by classCanonicalName) then there's more work to do here, starting with not depending on native getCanonicalClass.

Shall I open a ticket?

jvican: > Then your concern is of the status quo, not specifically this proposed change. Because currently (as demonstrated in the spec as of #380) you can't tell the difference between:
  
Not really. Zinc assumes that both class and companion class map to the same name. This is an invariant embedded in all the logic that handles incrementality. It not only happens in `getCanonicalClass`, but also in the bridge, where the names for both are the same.
  
This change, however, creates a clash because two different class files (one with a valid name and another one with an invalid name) will map to the same name. Those don't have any relationship between them, as in the case of class and companion class.
  
I'm not saying this patch is wrong. I like the fact that the notation is more Scala-like. But we must first understand the consequences of such a change.

dwijnand: > Zinc assumes that both class and companion class map to the same name. 

I presume you mean for the enclosing classes, given it's not true for top level class pairs (e.g O and C), or leaf class pairs (e.g OO and OC).

> This change, however, creates a clash because two different class files (one with a valid name and another one with an invalid name) will map to the same name. Those don't have any relationship between them, as in the case of class and companion class.

The relationship between OOO and OCO is a class and companion class relationship at the middle tier.

(btw I've said "relationship" enough time that at this point I feel like I'm commenting about a TV soap opera.. "Jorge, why can't OCO love OOO, just like he loves CCO?")

jvican: Don't OOO and OCO have the same name because you're stripping out the prefixed package name, which is different? Precisely because of this mapping you mention, you cannot write this code in Scala. You'll get the compile error we've mentioned before in previous PRs.

dwijnand: You're right, I/we shouldn't forget about the package prefix. And in this whole thread I forgot.

I'll come back to this thread after a break :) 🍻

jvican: Perfect. Enjoy!

dwijnand:
So I was wondering. Is there any particular reason we're trying so hard to name the given class when we get a malformed class name error? What's wrong with Class.getName?

jvican:
I'll come back to this on Wednesday. I'm on vacation for now.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/381/head&deploy_to_bintray=false">this link.</A>

eed3si9n:
I think we should clarify if we want a fully qualified name (String that is able to identify a type) or a real _canonical_ name (one and only name name that identifies a type).

https://docs.oracle.com/javase/specs/jls/se8/html/jls-6.html#jls-6.7:

> The difference between a fully qualified name and a canonical name can be seen in code such as:
>
> ```java
> package p;
> class O1 { class I {} }
> class O2 extends O1 {}
> ```
>
> Both `p.O1.I` and `p.O2.I` are fully qualified names that denote the member class `I`, but only `p.O1.I` is its canonical name.

For the purpose of extracting API Info, I'm guessing that fully qualified name is ok.

dwijnand:
@eed3si9n That's interesting info, thank you. However I'm left confused - could you clarify what change exactly are you suggesting, please?

eed3si9n:
Here's a straw-man table to simplify the situation. Correct me if the values are wrong here:

|                                             |  Canonical     |    Fully qualified |
|---------------------------| ------------- | ---------------- |
| Scala support                     | Not really?     | Yes                      |
| Needed for API extraction | No?                | Yes (as long as it's consisntent) |
| Backward compat              | Yes                 | Maybe?              |

We've already shipped of the name extraction using canonical name, but for the problematic ones we can fall back to fully qualified name, I guess is gist of my argument.

PR #314, opened 2017-06-09
Fix ConsoleInterface binding things properly.

dwijnand:
In reference to https://github.com/sbt/sbt/issues/2884 I'm seeing the
console helpers (cpHelpers) being statically 'Object', and therefore not
being that helpful:

    scala> cpHelpers
    res0: Object = sbt.internal.ConsoleProject$Imports@610be000

    scala> cpHelpers.taskKeyEvaluate
    <console>:37: error: value taskKeyEvaluate is not a member of Object
           cpHelpers.taskKeyEvaluate
                     ^

    scala> cpHelpers.asInstanceOf[sbt.internal.ConsoleProject.Imports].taskKeyEvaluate _
    res3: sbt.TaskKey[Nothing] => sbt.internal.ConsoleProject.Evaluate[Nothing] = $$Lambda$4294/1575143649@5a54d62c

This is because I misinterpreted the Scala 2.8 compatibility layer I
tore out in 1abf6ca3bfe0628321e1562a9b4cfe58e19ab7b7.

PR #328, opened 2017-06-30
Drop util-testing in favour of existing UnitSpec

dwijnand:

PR #386, opened 2017-07-31
Fix ConsoleInterface binding things properly^2

dwijnand:
Follow-up on #314 - I _still_ misinterpreted..

Turns out the ".asInstanceOf[AnyRef].getClass.getName" implementation
was the _original_ implementation. Then Mark switched to using bindValue
in sbt/sbt@4b8f0f3f941d5b3970fa24dfd9a34508d6974345.

Since Scala 2.11.0 (scala/scala#1648 in particular) bindValue was
removed. So we'll use NamedParam and quietBind, both which exist since
Scala 2.9.0.

Fixes sbt/sbt#2884, tested with local releases.

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/ConsoleInterface.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/ConsoleInterface.scala
@@ -8,7 +8,7 @@
 package xsbt
 
 import xsbti.Logger
-import scala.tools.nsc.interpreter.{ ILoop, IMain, InteractiveReader }
+import scala.tools.nsc.interpreter.{ ILoop, IMain, InteractiveReader, NamedParam }
 import scala.tools.nsc.reporters.Reporter
 import scala.tools.nsc.{ GenericRunnerCommand, Settings }
 
@@ -54,7 +54,7 @@ class ConsoleInterface {
           super.createInterpreter()
 
         for ((id, value) <- bindNames zip bindValues)
-          intp.beQuietDuring(intp.bind(id, value.asInstanceOf[AnyRef].getClass.getName, value))
+          intp.quietBind(NamedParam.clazz(id, value))
 
         if (!initialCommands.isEmpty)
           intp.interpret(initialCommands)


jvican (dismissed):
LGTM.

jvican:
Why is this targetting 1.0.1?

dwijnand:
Because @eed3si9n targeted the issue to 1.0.1: https://github.com/sbt/sbt/issues/2884#issuecomment-318778799

eed3si9n:
If it's not going to be part of `1.0.0`, then I don't think we should merge it in there.
Should we create a branch for `1.0.x`?

dwijnand:
I thought about that, but I'd rather not get into the business of more persistent branches.

I think we should merge it into the 1.0.0 branch and here are my reasons:

1. I think the bug fix here is clean and simple, and doesn't impact anything outside of console.

2. If we _do_ discover some release-blocking regression that forces our hand to do an RC4 release, this bug fix can go with it, imo. If we don't then this fix can ship as 1.0.1 or 1.1.0 at a later date.

dwijnand:
Oh and at this point in time the only thing in the 1.x branch that's not in the 1.0.0 branch is this inconsequential test classpath change: https://github.com/sbt/zinc/pull/328.

jvican:
Let's then aim for 1.0.0. I'm fine merging it and releasing RC4.

eed3si9n:
@dwijnand 

> I think we should merge it into the 1.0.0 branch and here are my reasons ...

So you are saying that RC-3 is still on its way to become 1.0.0 final, but you still want to merge this into `1.0.0` release branch? `1.0.0` is effectively a mutable tag that we are saying it will become 1.0.0 eventually.

If we plan to release 1.0.1 with bug fixes only, then we should create `1.0.x`. Otherwise, we should merge this to `1.x` and ship in `1.1.0` (in a few months?).

@jvican 

> Let's then aim for 1.0.0. I'm fine merging it and releasing RC4.

Releasing RC-4 means delaying the entire sbt 1.0 and Zinc 1.0 for another week or two.
I personally don't think this is worth it.

jvican:
> Releasing RC-4 means delaying the entire sbt 1.0 and Zinc 1.0 for another week or two.
I personally don't think this is worth it.
  
Hmmm... if releasing a trivial fix is going to delay sbt 1.0, then we have a real problem with our tooling. Personally, I don't like making wrong decisions because of constraints on the tools I use -- if the current repo setup is causing too much pain, we should have a monorepo or something that emulates it.

eed3si9n:
> Hmmm... if releasing a trivial fix is going to delay sbt 1.0, then we have a real problem with our tooling. 

It's a matter of policy, not tooling. Even if we had a monorepo e.g. sbt/sbt, I would still push the final date by a week or two if we made _any_ change to the code before calling it 1.0.0.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?zinc=pull/386/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/386/head&deploy_to_bintray=false">this link.</A>

eed3si9n (approved):

PR #391, opened 2017-08-10
Merge 1.0.0 into 1.x

dwijnand:


jvican (dismissed):

jvican:
Last merge commit (the one by Eugene) is failing.

eed3si9n:
```
		::::::::::::::::::::::::::::::::::::::::::::::
		::          UNRESOLVED DEPENDENCIES         ::
		::::::::::::::::::::::::::::::::::::::::::::::
		:: org.scala-sbt#zinc-classfile_2.12;1.0.0-X20: configuration not public in org.scala-sbt#zinc-classfile_2.12;1.0.0-X20: 'test'. It was required from org.scala-sbt#zinc-apiinfo_2.12;1.0.0-X20 test
		::::::::::::::::::::::::::::::::::::::::::::::
```

Where is 1.0.0-X20 coming from? Since Maven Central won't have Test configuration, the error itself seems reasonable.

eed3si9n:
a88096d is tagged to be 1.0.0, so I suggest you add some bogus commit to move away from it.
(This is a step I think scala/scala has in their release procedure iirc)

dwijnand:
zinc-classfile 1.0.0-X20 is coming from sbt 1.0.0-RC2 which is what zinc's build is using.

Currently I'm thinking that this is due to whatever the state of drone's cache is.

jvican:
I've removed all the Drone caches for Zinc. Try again and confirm whether this was caused by it.

jvican:
Looks like it's fixed.

jvican:
Original error is fixed, but another one has surfaced.

dwijnand:
Thanks, Jorge, I'll fix the other issue.

> I've removed all the Drone caches for Zinc.

How does one do that?

jvican:
> How does one do that?
  
Via sftp, connecting to one of our servers. For now, doing so outside of EPFL is tricky because you need the SSH key, but if these problems continue I'm happy to give you access to the server.

dwijnand:
I'm not sure if Eugene already has SSH access to EPFL, but I'd like to have access to be able to resolve these types of issues. What's the procedure? Who do I contact?

jvican:
Eugene doesn't -- these are new machines. Let's discuss this via email.

dwijnand:
Green! Omg, this PR is finally over.

Issue #389, opened 2017-08-10
Crash on null from Method.getGenericParameterTypes


stuhood:
In some cases (seemingly related to lambdas), `Method.getGenericParameterTypes` will return an array containing `null` `Type`s, which are then passed into `def defLike` to cause crashes.

This commit has a self-contained repro: https://github.com/twitter/pants/commit/aaa07c3e0d46d5750c1f4281590ec56d94ca2504 ... that crashes when compiled with zinc `1.0.0-rc3`:
```
[error] ## Exception when compiling 1 sources to current/classes
[error] null
[error] sbt.internal.inc.ClassToAPI$.reference(ClassToAPI.scala:506)
[error] sbt.internal.inc.ClassToAPI$.parameter(ClassToAPI.scala:375)
[error] sbt.internal.inc.ClassToAPI$$anonfun$16.apply(ClassToAPI.scala:351)
[error] sbt.internal.inc.ClassToAPI$$anonfun$16.apply(ClassToAPI.scala:350)
[error] scala.runtime.Tuple3Zipped$$anonfun$map$extension$1.apply(Tuple3Zipped.scala:44)
[error] scala.runtime.Tuple3Zipped$$anonfun$map$extension$1.apply(Tuple3Zipped.scala:42)
[error] scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
[error] scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:186)
[error] scala.runtime.Tuple3Zipped$.map$extension(Tuple3Zipped.scala:42)
[error] sbt.internal.inc.ClassToAPI$.defLike(ClassToAPI.scala:350)
[error] sbt.internal.inc.ClassToAPI$.methodToDef(ClassToAPI.scala:308)
[error] sbt.internal.inc.ClassToAPI$$anonfun$4.apply(ClassToAPI.scala:182)
[error] sbt.internal.inc.ClassToAPI$$anonfun$4.apply(ClassToAPI.scala:182)
[error] sbt.internal.inc.ClassToAPI$$anonfun$mergeMap$1.apply(ClassToAPI.scala:398)
[error] sbt.internal.inc.ClassToAPI$$anonfun$mergeMap$1.apply(ClassToAPI.scala:398)
[error] scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:241)
[...]

PR #393, opened 2017-08-11
Remove workarounds for sbt/sbt#1035

jvican:
The workarounds are not necessary anymore,
http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6476261 has been
fixed in JDK8, which is the JDK version that Zinc demands.

We can change the target branch after this is LGTM'ed and the forward port
for 1.0.x is finished.

### DIFF
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -548,56 +548,18 @@ object ClassToAPI {
   }.toMap
   def primitive(name: String): api.Type = PrimitiveRefs(name)
 
-  // Workarounds for https://github.com/sbt/sbt/issues/1035
-  //   these catch the GenericSignatureFormatError and return the erased type
+  private[this] def returnType(f: Field): Type = f.getGenericType
+  private[this] def returnType(m: Method): Type = m.getGenericReturnType
+  private[this] def exceptionTypes(c: Constructor[_]): Array[Type] = c.getGenericExceptionTypes
 
-  private[this] def returnType(f: Field): Type =
-    try f.getGenericType
-    catch {
-      case _: GenericSignatureFormatError => f.getType
-    }
-  private[this] def parameterTypes(c: Constructor[_]): Array[Type] =
-    try c.getGenericParameterTypes
-    catch {
-      case _: GenericSignatureFormatError => convert(c.getParameterTypes)
-    }
-  private[this] def exceptionTypes(c: Constructor[_]): Array[Type] =
-    try c.getGenericExceptionTypes
-    catch {
-      case _: GenericSignatureFormatError => convert(c.getExceptionTypes)
-    }
-  private[this] def parameterTypes(m: Method): Array[Type] =
-    try m.getGenericParameterTypes
-    catch {
-      case _: GenericSignatureFormatError => convert(m.getParameterTypes)
-    }
-  private[this] def returnType(m: Method): Type =
-    try m.getGenericReturnType
-    catch {
-      case _: GenericSignatureFormatError => m.getReturnType
-    }
-  private[this] def exceptionTypes(m: Method): Array[Type] =
-    try m.getGenericExceptionTypes
-    catch {
-      case _: GenericSignatureFormatError => convert(m.getExceptionTypes)
-    }
+  private[this] def exceptionTypes(m: Method): Array[Type] = m.getGenericExceptionTypes
+  private[this] def parameterTypes(m: Method): Array[Type] = m.getGenericParameterTypes
+  private[this] def parameterTypes(c: Constructor[_]): Array[Type] = c.getGenericParameterTypes
 
   private[this] def typeParameterTypes[T](m: Constructor[T]): Array[TypeVariable[Constructor[T]]] =
-    try m.getTypeParameters
-    catch {
-      case _: GenericSignatureFormatError => new Array(0)
-    }
+    m.getTypeParameters
   private[this] def typeParameterTypes[T](m: Class[T]): Array[TypeVariable[Class[T]]] =
-    try m.getTypeParameters
-    catch {
-      case _: GenericSignatureFormatError => new Array(0)
-    }
+    m.getTypeParameters
   private[this] def typeParameterTypes(m: Method): Array[TypeVariable[Method]] =
-    try m.getTypeParameters
-    catch {
-      case _: GenericSignatureFormatError => new Array(0)
-    }
-
-  private[this] def convert(classes: Array[Class[_]]): Array[Type] =
-    classes.asInstanceOf[Array[Type]] // ok: treat Arrays as read-only
+    m.getTypeParameters
 }


jvican:
Note the ref to https://github.com/sbt/sbt/issues/1035.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -548,56 +548,18 @@ object ClassToAPI {
...
-    }
-  private[this] def typeParameterTypes(m: Method): Array[TypeVariable[Method]] =
-    try m.getTypeParameters
-    catch {
-      case _: GenericSignatureFormatError => new Array(0)
-    }
-
-  private[this] def convert(classes: Array[Class[_]]): Array[Type] =
-    classes.asInstanceOf[Array[Type]] // ok: treat Arrays as read-only
+  @inline private[this] def returnType(f: Field): Type = f.getGenericType
```
eed3si9n: Does `@inline` annotation in Scala really help performance? - https://stackoverflow.com/questions/2709095/does-the-inline-annotation-in-scala-really-help-performance

eed3si9n (changes requested):

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
jvican: Yes, it does. Inlining is extremely important and it can cause slowdowns, but in Scala code it generally yields good benefits. For example, using `@inline` in certain places of the Scala compiler is critical for performance (an example is all the statistics infrastructure). JVM inlining is normally not enough for Scala code because it will only inline up to 135 bytes, IIRC, and this is why the careful use of `@inline` is important. I saw a tweet by Jason some weeks ago claiming that increasing this bytecode limit in the JDK helps performance for Scalac.
  
So, coming back to your question: yes, `@inline` has a real-world impact. It will not make the program 20% faster in this scenario, but it's a good practice to use `@inline` where applicable, and I think in this case it is.

jvican: Note also that the use of `@inline` does not mean the Scala compiler will inline it. It may or may not, depending on some heuristics. But this doesn't prove my point wrong.

eed3si9n: @cb372 gave a whole talk on this topic. What was the verdict on this topic? `-optimize` is better than handcrafted `@inline`? http://slides.com/cb372/inline-specialized-berlin-2016#/27

jvican: That's a benchmark of numerical computations. Those kind of benchmarks are really misleading and easy to optimize efficiently by the Hotspot, probably with not even inlining but also loop unfolding et al. To add to that, it doesn't say anything about the benchmark setup and 2.12.3 improved the optimizer and inliner, making those benchmarks (for 2.12.0-M2) useless. So I think that link has no relevance to the scenario we're dealing with here.
  
By the by, this scenario is clearly beneficial for inlining because we only have one call-site, and it's in the same class definition (unlike the benchmark you mention). The benefit is that we don't pay the cost of invoking these functions, while still making them logically independent if in the future we want to add more logic to them, like a fix for https://github.com/sbt/zinc/issues/389.
  
I think we would agree that Zinc is more similar to Scalac than to that benchmark. It is reasonable to think `@inline` here is beneficial, given how beneficial it is for the Scala compiler. If you want to keep having this discussion, I suggest you have it with the Scala team. They will give you better insights than I will.
  
That being said, it would be good that you comment on the core issue that this patch is addressing.

eed3si9n: As per reverting/removing https://github.com/sbt/sbt/commit/9984a00e25d9d6b5e7cd67e95c8d8a6123b9af7d, I am :+1: on that.

jvican: That's cool. Can you LGTM and we merge?

eed3si9n: I'd like to get Scala team's opinion as you suggested in terms of:
1. If it's effective to add `@inline` in non-optimized code, and 
2. If it's _safe_ to call reflection like `f.getGenericType` from an inline. scala/bug#6789 for example says "The one-argument Class.forName must never be inlined"

jvican: Wrt the first question, we can merge this and address that doubt later on. I don't think it's a blocker.
  
Wrt your second question, it **is** safe to inline `f.getGenericType`. The reason why you cannot inline `Class.forName` is because `this.getClass.getClassLoader` or its absence has semantics that depends on the call-site (via stack inspection). This is absolutely not the case for these methods, that rely uniquely on the reflection API and don't take any class loader as parameter.

lrytz: Inlining `getGenericType` is probably safe, https://github.com/scala/bug/issues/6789 a special case for `forName`. Note that, since 2.12.3, you have to be explicit in telling the inliner where it's allowed to inline from (https://github.com/scala/scala/pull/5964).

Inlining from the JDK is probably not a good idea in general as the bytecode potentially becomes non-portable. There might be licensing issues too. Maybe that's not a problem here because that code is only compiled locally for every user and never distributed in binary.

`@inline` is not a hint to the heuristic, but it makes the optimizer inline unconditionally (if the method can be inlined). The heuristic applies for every method invocation.

Marking small forwarder methods `@inline` is usually not beneficial, the JVM handles small monomorphic methods just fine. The only potential issue with many small methods on a hot path is that they consume from the `-XX:MaxInlineLevel` budget.

Inlining in Scala has benefits when it can make the code easier for the VM to optimize, in particular when callsites become monomorphic or escape analysis / stack allocation starts working.

> it's a good practice to use @inline where applicable

I don't agree here. For small methods like here there's no gain. For bigger methods, there's a risk of slowdown due to code size. Good practice is to use the annotation where performance benefits are measured (or known).

I once made a small experiment and removed all `@inline` annotations from the Scala compiler, and performance didn't change. It might be different now, we'd have to do it again. But I think the heuristics cover the main cases. As another data point, there's a stress test mode in the optimizer `-Yopt-inline-heuristics:everything`. If you build a compiler with that, it runs 2x slower.

In general when working on JVM performance optimization, measuring is really important. More often than not, just reasoning about it leads to changes that don't improve or even hurt performance - there are just too many factors in play.

lrytz:
> Inlining from the JDK is probably not a good idea

Oh, I got this wrong, there's no inlining from the JDK here.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
lrytz: > Inlining from the JDK is probably not a good idea

Oh, I got this wrong, there's no inlining from the JDK here.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/393/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

jvican:
This PR has been updated and can be reviewed.

eed3si9n (dismissed):
LGTM pending CI

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/393/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/393/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

jvican:
Merging as code hasn't changed since last approval.

PR #408, opened 2017-09-15
Adds external hook to filter scala compiler options for product invalidation

### LINKED ISSUE #406: Add a hook to filter out Scala options in every compile
If scala options change from one run to the other one, Zinc recompiles the whole project conservatively.
  
With the advent of compilers like Dotty, Hydra and `rsc`, the current behaviour is too restrictive because some downstream users may want to filter out scalac options that are dependent to the way compilation should occur (like number of cores). This is safe to do because downstream users know which scala flags affect the compilation and which do not.
  
You can add this hook in `ExternalHooks` https://github.com/sbt/zinc/blob/1.x/internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java and https://github.com/sbt/zinc/blob/1.x/internal/compiler-interface/src/main/java/xsbti/compile/DefaultExternalHooks.java. The method `compile` in `IncrementalCompilerImpl` already takes an instance of `IncOptions`, which inside has an instance of `ExternalHooks`, so the last thing to do is to use the external hook we've just provided in the body of `makeConfig`, inside `MixedAnalyzingCompiler`: https://github.com/sbt/zinc/blob/1.x/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala#L183-L214
  
/cc @dragos

fiadliel:
This adds `filterScalaCompilerOptions` to `ExternalHooks`, which can be overridden to remove scala compiler options which do not affect the resulting products.

Fixes #406

### DIFF
--- a/internal/compiler-interface/src/main/java/xsbti/compile/DefaultExternalHooks.java
+++ b/internal/compiler-interface/src/main/java/xsbti/compile/DefaultExternalHooks.java
@@ -45,4 +45,9 @@ public class DefaultExternalHooks implements ExternalHooks {
     public ExternalHooks withExternalLookup(ExternalHooks.Lookup externalLookup) {
         return new DefaultExternalHooks(Optional.of(externalLookup), this.getExternalClassFileManager());
     }
+
+    @Override
+    public String[] filterScalaCompilerOptions(String[] options) {
+        return options;
+    }
 }
--- a/internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java
+++ b/internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java
@@ -86,4 +86,14 @@ public interface ExternalHooks {
      * @return An instance of {@link ExternalHooks} with the specified lookup.
      */
     ExternalHooks withExternalLookup(Lookup externalLookup);
+
+    /**
+     * Alter which scala compiler options are considered for invalidation of compiled products.
+     *
+     * Expected to be used by filtering out options which do not affect the products,
+     * e.g. debug/logging flags, flags which change the number of cores used, et al.
+     *
+     * @return A String array used to calculate changes in scala options for invalidation.
+     */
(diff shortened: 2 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/408/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

jvican:
Thanks a lot Gary :smile: @fiadliel! This is a useful change.

The error in the CI is due to MiMa, but this is a backwards compatible change, so I'll update this PR with the MiMa whitelist so that CI passes.

dotta:
@fiadliel @jvican Am I correct to think that this change is going to be exposed to sbt users via [`incOptions`](https://github.com/sbt/sbt/blob/1.x/main/src/main/scala/sbt/Keys.scala#L188)?

jvican:
You're correct, `incOptions` has an instance of `ExternalTools`.

dwijnand:
How is adding an abstract method to a java interface a backwards compatible change?

dotta:
It is, see https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html

> here is a list of some important binary compatible changes that the Java programming language supports:
>
> * Adding new fields, methods, or constructors to an existing class or interface.

dwijnand:
Oh sorry, my mistake, I didn't see it was a `ReversedMissingMethodProblem`.

dwijnand (approved):

jvican:
I promise I'm updating the whitelist here soon. It was recently introduced in a PR by @eed3si9n, so all we need to do is to rebase.

dwijnand:
There's now a merge conflict here. @fiadliel are you happy to resolve that? If not I can rebase it and push that.

jvican:
Fixed in https://github.com/sbt/zinc/issues/406 instead.

PR #422, opened 2017-10-07
Names from nonexisiting objects

romanowski:
In some cases zinc reports dependencies to non-exiting objects that are taken from e.g. imports. This may later on manifest as dependencies to jar/classes from projects that should be handled as external dependencies.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/422/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

zinc/src/test/scala/sbt/inc/BinaryDepSpec.scala
```diff
@@ -0,0 +1,26 @@
+package sbt.inc
+
+import sbt.internal.inc.Analysis
+import sbt.io.IO
+
+class BinaryDepSpec extends BaseCompilerSpec{
+  it should "not depend on non-existing objects" in {
```
eed3si9n: Is this expectation true? What if you changed the macro? Shouldn't we invalidate the macro consumer?

romanowski: We do not change macro here we compile each project once. 

We depend on NoopMacro but as a external dependency and we should not depended at the same time on it as library dependency (so we don't want recompile every usage of NoopMacro when we change anything in jar where it comes from)

romanowski:
@eed3si9n @jvican @dwijnand Please let me know if you need any need anything more to review this PR :)

jvican:
Myself I just need time. I'll get back to this once the big PRs in the queue are merged.

dwijnand:
Where do we stand on this PR, @romanowski @eed3si9n @jvican?

It's green and merges cleanly. Can we land it?

jvican:
I’m trying to review it this week.

jvican (approved):
LGTM

jvican:
I'll merge when CI passes.

PR #423, opened 2017-10-07
Fix handling deeply nested java classes.

romanowski:
Without fix it reports dependencies to jar/classes produced by same project.

### DIFF
--- a/internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
+++ b/internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
@@ -50,8 +50,16 @@ private[sbt] object Analyze {
          binaryClassName = classFile.className;
          loadedClass <- load(binaryClassName, Some("Error reading API from class file"))) {
       binaryClassNameToLoadedClass.update(binaryClassName, loadedClass)
-      val srcClassName = binaryToSourceName(loadedClass)
-        .orElse(Option(loadedClass.getEnclosingClass).flatMap(binaryToSourceName))
+
+      def loadEnclosingClass(clazz: Class[_]): Option[String] = {
+        binaryToSourceName(clazz) match {
+          case None if clazz.getEnclosingClass != null =>
+            loadEnclosingClass(clazz.getEnclosingClass)
+          case other => other
+        }
+      }
+
+      val srcClassName = loadEnclosingClass(loadedClass)
 
       srcClassName match {
         case Some(className) =>
--- a/zinc/src/test/scala/sbt/inc/NestedJavaClassSpec.scala
+++ b/zinc/src/test/scala/sbt/inc/NestedJavaClassSpec.scala
@@ -1,0 +1,20 @@
+package sbt.inc
+
+import sbt.internal.inc.Analysis
+import sbt.io.IO
+
+class NestedJavaClassSpec extends BaseCompilerSpec {
+  it should "handle nested Java classes" in {
+    IO.withTemporaryDirectory { tempDir =>
+      val projectSetup =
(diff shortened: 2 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/423/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

eed3si9n (dismissed):
LGTM

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
```diff
@@ -50,8 +50,16 @@ private[sbt] object Analyze {
          binaryClassName = classFile.className;
          loadedClass <- load(binaryClassName, Some("Error reading API from class file"))) {
       binaryClassNameToLoadedClass.update(binaryClassName, loadedClass)
-      val srcClassName = binaryToSourceName(loadedClass)
-        .orElse(Option(loadedClass.getEnclosingClass).flatMap(binaryToSourceName))
+
+      def loadEnclosingClass(clazz: Class[_]): Option[String] = {
+        binaryToSourceName(clazz) match {
+          case None if loadedClass.getEnclosingClass != null =>
```
dwijnand: I think you mean `clazz.getEnclosingClass != null` here.

dwijnand (changes requested):

zinc/src/test/scala/sbt/inc/NestedJavaClassSpec.scala
```diff
@@ -0,0 +1,24 @@
+package sbt.inc
+
+import sbt.internal.inc.Analysis
+import sbt.io.IO
+
+/**
```
jvican: Can we remove this header? We don't use them anywhere in the project.

zinc/src/test/scala/sbt/inc/NestedJavaClassSpec.scala
```diff
@@ -0,0 +1,24 @@
...
+    IO.withTemporaryDirectory { tempDir =>
+      val projectSetup =
+        ProjectSetup.simple(tempDir.toPath, Seq("NestedJavaClasses.java"))
+
+      val result = projectSetup.createCompiler().doCompile()
+      result.analysis() match {
+        case analysis: Analysis =>
+          analysis.relations.libraryDep._2s
+            .filter(_.toPath.startsWith(tempDir.toPath)) shouldBe 'empty
+
```
jvican: Remove empty space?

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
jvican: How was the test case even passing with this error? This would mean that more than one level of nesting would not be handled.

jvican (changes requested):
LGTM. Thanks @romanowski. Waiting for your feedback.

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/Analyze.scala
romanowski: Tests passes since if `loadedClass` is nested class (regardless how deep) it will be always null.

The only problem with this is that in case that for some reason we don't have source mapping for top enclosing class we will get NullPointerException here. But anyway this was stupid mistake.

jvican (approved):
LGTM @romanowski, thanks for this work.

dwijnand (approved):

jvican:
I see @dwijnand has approved and @eed3si9n already had a pass through this. Merging then.

PR #427, opened 2017-10-10
Allow clients to override hashing classpath.

### LINKED ISSUE #400: We should allow clients to overriding classpath hashing
Zinc has a lot hooks to improve incremental compilation performance in case of really big projects. However when we hash classpath we just do tons of IO operations without  any way to e.g. reuse libs hashed before.

In case of 100+ projects with 100+ libs and on slower IO (e.g. windows) it takes a while **during each compilation (even no-op ones)**.

I haven't found a way to do it  without breaking binary compatibility but simple proposal how it can be done is here: https://github.com/romanowski/zinc/tree/hashing-classpath-proposal  

@stuhood I think pants may also benefit from caching classpath hashes.

romanowski:
This is useful in big builds with 100+ projects with 100+ jars on classpath (mostly the same).

Fixes #400

### DIFF
--- a/internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java
+++ b/internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java
@@ -51,6 +51,8 @@ public interface ExternalHooks {
          * @return API changes
          */
         boolean shouldDoIncrementalCompilation(Set<String> changedClasses, CompileAnalysis previousAnalysis);
+
+        Optional<FileHash[]> hashClasspath(File[] classpath);
     }
 
     /**
--- a/zinc/src/main/scala/sbt/internal/inc/LookupImpl.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/LookupImpl.scala
@@ -8,6 +8,7 @@
 package sbt.internal.inc
 
 import java.io.File
+import java.util.Optional
 
 import xsbti.compile.{ Changes, CompileAnalysis, FileHash, MiniSetup }
 
@@ -39,11 +40,8 @@ class LookupImpl(compileConfiguration: CompileConfiguration, previousSetup: Opti
 
   private val entry = MixedAnalyzingCompiler.classPathLookup(compileConfiguration)
 
-  override def lookupAnalysis(binaryClassName: String): Option[CompileAnalysis] = {
-    analyses.collectFirst {
-      case a if a.relations.productClassName._2s.contains(binaryClassName) => a
-    }
-  }
+  override def lookupAnalysis(binaryClassName: String): Option[CompileAnalysis] =
+    analyses.find(_.relations.productClassName._2s.contains(binaryClassName))
 
   override def lookupOnClasspath(binaryClassName: String): Option[File] =
     entry(binaryClassName)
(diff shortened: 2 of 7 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/427/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

jvican:
@romanowski Formatting error on CI. I also expect this to report a MiMa error. Make sure to rebase this PR with the latest change in master and add the forward mima issue to the whitelist.
  
Thank you Kryzstoff.

jvican:
Can someone in the sbt team review this?

internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java
```diff
@@ -51,6 +51,8 @@
          * @return API changes
          */
         boolean shouldDoIncrementalCompilation(Set<String> changedClasses, CompileAnalysis previousAnalysis);
+
+        Optional<FileHash[]> hashClasspath(File[] classpath);
```
eed3si9n, line 55: Kind of curious. Does it matter if this takes `File[]` or just `hash(File)`?

eed3si9n (changes requested):
Added a question.

internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java
romanowski: This way you can add e.g. paralelizm but with `hash(File)` you are not.

zinc/src/main/scala/sbt/internal/inc/LookupImpl.scala
```diff
@@ -64,4 +62,7 @@ class LookupImpl(compileConfiguration: CompileConfiguration, previousSetup: Opti
   override def shouldDoIncrementalCompilation(changedClasses: Set[String],
                                               analysis: CompileAnalysis): Boolean =
     externalLookup.forall(_.shouldDoIncrementalCompilation(changedClasses, analysis))
+
+  override def hashClasspath(classpath: Array[File]): Optional[Array[FileHash]] =
```
fommil, line 66: nice, this allows the user to provide a parallel implementation.

zinc/src/main/scala/sbt/internal/inc/LookupImpl.scala
```diff
@@ -39,11 +40,8 @@ class LookupImpl(compileConfiguration: CompileConfiguration, previousSetup: Opti
 
   private val entry = MixedAnalyzingCompiler.classPathLookup(compileConfiguration)
 
-  override def lookupAnalysis(binaryClassName: String): Option[CompileAnalysis] = {
-    analyses.collectFirst {
-      case a if a.relations.productClassName._2s.contains(binaryClassName) => a
-    }
-  }
+  override def lookupAnalysis(binaryClassName: String): Option[CompileAnalysis] =
```
fommil, line 43: why not make this non-optional and provide a default implementation?

romanowski: Empty result here has a meaning - that we have to treat this class as Library dependency instead of External one (so names will be invalidated on any change in *.class or *.jar file).

This change was just optimization since `find` is faster then `collectFirst` (and code looks also better)

fommil: I thought the empty meant that it falls back to the default impl of blindly using SHA-1. Couldn't that default impl just be moved to be the default?

jvican:
I'd like to do some work on this area. Can we approve and merge this? /cc @sbt/zinc-team

jvican (approved):
LGTM.

eed3si9n (approved):

internal/compiler-interface/src/main/java/xsbti/compile/ExternalHooks.java
eed3si9n: ok. Sounds good to me.

jvican:
Updating branch. If it succeeds, I'll merge.

romanowski:
@jvican  all is green

PR #335, opened 2017-07-06
[Do not merge] Adapt Zinc to use the new LibraryManagement API

Duhemm:
Depends on sbt/librarymanagement#124.

PR #428, opened 2017-10-12
Refactor the build

jvican:
The build is now structure in the following way:

* `build.sbt` keeps the project definitions.
* `BuildPlugin.scala` keeps the logic.

This way, at a glance, contributors can see which modules the build defines.

This PR does not change the logic of the build. I'll open a new PR after this
one that will improve the scripted test setup, and add `sbt-release-early` via
a new version of `sbt-houserules`.

I hope this change makes the build more readable and easier to hack in the future.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/428/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

project/Header.scala
```diff
@@ -1,27 +1,25 @@
...
-//        | * This software is released under the terms written in LICENSE.
-//        | */
-//        |
-//        |""".stripMargin
-// }
+  override def projectSettings = Seq(
+    SbtHeaderKeys.headerMappings ++= Map(
+      HeaderFileType.scala -> HeaderCommentStyle.CStyleBlockComment,
+      HeaderFileType.java -> HeaderCommentStyle.CStyleBlockComment
+    ),
```
dwijnand: This isn't necessary. It's the default definition: https://github.com/sbt/sbt-header/blob/v3.0.2/src/main/scala/de/heikoseeberger/sbtheader/HeaderPlugin.scala#L131-L134

project/Header.scala
```diff
@@ -1,27 +1,25 @@
...
 
-// import de.heikoseeberger.sbtheader.{ HeaderPlugin, HeaderPattern }
-// import HeaderPlugin.autoImport._
+import de.heikoseeberger.sbtheader.HeaderPlugin
+import de.heikoseeberger.sbtheader.HeaderPlugin.{autoImport => SbtHeaderKeys}
 
-// object CustomHeaderPlugin extends AutoPlugin {
-//   override def requires = plugins.JvmPlugin && HeaderPlugin
-//   override def trigger = allRequirements
+object CustomHeaderPlugin extends AutoPlugin {
```
dwijnand: Could you add `headerCheck` and `test:headerCheck` enforcement in CI please?

project/BuildPlugin.scala
```diff
@@ -0,0 +1,93 @@
+import sbt.{ AutoPlugin, Compile, Def, Keys, Resolver, Test, TestFrameworks, Tests, URL }
+
+object BuildPlugin extends AutoPlugin {
+  override def requires = sbt.plugins.JvmPlugin
+  override def trigger = allRequirements
+  val autoImport = BuildKeys
+
+  override def projectSettings: Seq[Def.Setting[_]] = BuildImplementation.projectSettings
+  override def buildSettings: Seq[Def.Setting[_]] = BuildImplementation.buildSettings
+  override def globalSettings: Seq[Def.Setting[_]] = BuildImplementation.globalSettings
```
dwijnand: Why the indirection?

project/BuildPlugin.scala
```diff
@@ -0,0 +1,93 @@
+import sbt.{ AutoPlugin, Compile, Def, Keys, Resolver, Test, TestFrameworks, Tests, URL }
+
+object BuildPlugin extends AutoPlugin {
+  override def requires = sbt.plugins.JvmPlugin
+  override def trigger = allRequirements
+  val autoImport = BuildKeys
```
dwijnand: Why the indirection?

project/BuildPlugin.scala
```diff
@@ -0,0 +1,93 @@
...
+}
+
+object BuildKeys {
+  val bridgeScalaVersions: List[String] =
+    List(Dependencies.scala212, Dependencies.scala211, Dependencies.scala210)
+
+  // Ids that we will use to name our projects in build.sbt
+  val CompilerInterfaceId = "compiler-interface"
+  val CompilerBridgeId = "compiler-bridge"
+  val ZincApiInfoId = "zinc-apiinfo"
```
dwijnand: These aren't ids, they're the name of the directories the projects are defined in, and will I expect will have broken the custom commands.

project/BuildPlugin.scala
```diff
@@ -228,5 +254,106 @@ object BuildImplementation {
...
+
+      val scriptedClasspath = (Keys.fullClasspath in scriptedRef in Test).value
+      val instance = (Keys.scalaInstance in scriptedRef in Test).value
+      Scripted.doScripted(scriptedClasspath, instance, source, result, logged, hook)
+    }
+
+    def zincOnlyScripted(scriptedRef: Project): Def.Initialize[InputTask[Unit]] = Def.inputTask {
+      val result = scriptedSource(dir => (s: State) => scriptedParser(dir)).parsed
+      val scriptedClasspath = (Keys.fullClasspath in scriptedRef in Test).value
+      val instance = (Keys.scalaInstance in scriptedRef in Test).value
```
dwijnand: Why is this `in Test`?

build.sbt
```diff
@@ -35,28 +42,22 @@ lazy val zinc = (project in file("zinc"))
...
     zincTesting % Test
   )
-  .configure(addTestDependencies)
-  .settings(
-    name := "zinc",
-    mimaSettings,
-  )
 
 lazy val zincTesting = (project in internalPath / "zinc-testing")
+  .configure(addSbtLmCore, addSbtLmIvy, addTestDependencies, noPublish)
```
dwijnand: why were test dependencies added here?

project/BuildPlugin.scala
```diff
@@ -0,0 +1,359 @@
...
+                   new URL("https://dl.bintray.com/sbt/ivy-snapshots/"))(Resolver.ivyStylePatterns)
+    val all: List[Resolver] =
+      List(TypesafeReleases, SonatypeSnapshots, BintrayMavenReleases, BintraySbtIvySnapshots)
+
+    import BuildKeys.{ ZincAlternativeCacheName, ZincAlternativeCacheDir }
+    val AlternativeLocalResolver: Resolver =
+      Resolver.file(ZincAlternativeCacheName, ZincAlternativeCacheDir)(Resolver.ivyStylePatterns)
+
+    // Naive way of implementing a filter to remove repeated resolvers.
+    def removeRepeatedResolvers(rs: Seq[Resolver]): Seq[Resolver] = rs.toSet.toVector
```
dwijnand: .distinct

project/BuildPlugin.scala
```diff
@@ -0,0 +1,359 @@
...
+  override def requires = sbt.plugins.JvmPlugin
+  override def trigger = allRequirements
+  val autoImport = BuildKeys
+
+  override def projectSettings: Seq[Def.Setting[_]] = BuildImplementation.projectSettings
+  override def buildSettings: Seq[Def.Setting[_]] = BuildImplementation.buildSettings
+  override def globalSettings: Seq[Def.Setting[_]] = BuildImplementation.globalSettings
+}
+
+object BuildKeys {
```
dwijnand: This name isn't good because it contains a bunch of things that aren't keys.

project/BuildPlugin.scala
```diff
@@ -0,0 +1,359 @@
...
+    def runBenchmarks(benchmarkProject: Project): Command = {
+      val dirPath = BuildKeys.benchmarksTestDir.getAbsolutePath
+      val projectId = benchmarkProject.id
+      val runPreSetup = s"$projectId/run $dirPath"
+      val runBenchmark = s"$projectId/jmh:run -p _tempDir=$dirPath -prof gc"
+      val tearDownResources = s"$projectId/tearDownBenchmarkResources"
+      Command.command("runBenchmarks")(st => runPreSetup :: runBenchmark :: tearDownResources :: st)
+    }
+
+    val all: List[Command] = List(crossTestBridges, publishBridgesAndSet, publishBridgesAndTest)
```
dwijnand: all not containing all is confusing.

build.sbt
```diff
@@ -235,71 +94,64 @@ lazy val zincCore = (project in internalPath / "zinc-core")
...
-    sourceManaged in (Compile, generateContrabands) := baseDirectory.value / "src" / "main" / "contraband-java",
     mimaSettings,
   )
-  .configure(addSbtUtilLogging, addSbtIO, addSbtUtilControl)
 
 // defines Java structures used across Scala versions, such as the API structures and relationships extracted by
 //   the analysis compiler phases and passed back to sbt.  The API structures are defined in a simple
 //   format from which Java sources are generated by the sbt-contraband plugin.
-lazy val compilerInterface = (project in internalPath / "compiler-interface")
+lazy val compilerInterface = (project in file(CompilerInterfaceId))
```
dwijnand: :+1:

project/BuildPlugin.scala
```diff
@@ -0,0 +1,359 @@
...
+  val zincPublishLocalSettings: Seq[Def.Setting[_]] = List(
+    Keys.resolvers += BuildResolvers.AlternativeLocalResolver,
+    zincPublishLocal := BuildDefaults.zincPublishLocal.value,
+  )
+
+  val tearDownBenchmarkResources: TaskKey[Unit] = taskKey[Unit]("Remove benchmark resources.")
+  val benchmarksTestDir = sbt.IO.createTemporaryDirectory
+
+  val sourcesForAllScalaVersionsSetting: Seq[Def.Setting[_]] =
+    List(Keys.unmanagedSourceDirectories ++= BuildDefaults.handleScalaSpecificSources.value)
```
dwijnand: Why the indirection? We already have a brand-new and obscure "sourcesForAllScalaVersionsSetting", why not define here what it does instead of deferring to `BuildDefaults.handleScalaSpecificSources` and then depending on build.sbt to make sure to call `inCompileAndTest`?

project/BuildPlugin.scala
```diff
@@ -0,0 +1,359 @@
...
+  import sbt.{ ScmInfo }
+  val buildSettings: Seq[Def.Setting[_]] = List(
+    Scripted.scriptedBufferLog := true,
+    GitKeys.baseVersion := BuildKeys.baseVersion,
+    GitKeys.gitUncommittedChanges := BuildDefaults.gitUncommitedChanges.value,
+    BintrayKeys.bintrayPackage := "zinc",
+    ScalafmtKeys.scalafmtOnCompile := true,
+    ScalafmtKeys.scalafmtVersion := "1.2.0",
+    ScalafmtKeys.scalafmtOnCompile in Sbt := false,
+    Keys.description := "Incremental compiler of Scala",
```
dwijnand, line 73: Keys are in a single shared namespace, which is why you can refer to them by name (e.g `TaskKey[Unit]("publishSigned") := (())`). So we should just import them instead of referring to them like this. Particularly when you end up with this repetition:

```scala
    GitKeys.gitUncommittedChanges := BuildDefaults.gitUncommitedChanges.value,
    BintrayKeys.bintrayPackage := "zinc",
    ScalafmtKeys.scalafmtOnCompile := true,
```

Could you revert back please?

dwijnand (changes requested):

project/BuildPlugin.scala
jvican: A matter of taste, I guess. I find that having `BuildKeys` being an object outside of the `AutoPlugin` definition helps readability. Less nesting and more clarity. I also want to define keys independently of the logic of the build (their implementation).

project/BuildPlugin.scala
jvican: Yup. This would have worked pre-1.0.

project/BuildPlugin.scala
jvican: Yup, we can move `unmanagedSourceDirectories` to the build.sbt.

project/BuildPlugin.scala
jvican: > Keys are in a single shared namespace,
  
Which conflicts with other things in scope because they have too many generic names, like `moduleID` et al. I don't want to accidentally be referring to a local variable named this way because it's a source of bugs.
  
The good things of having everything accessible by `Keys` are:

* It's clear that the keys I'm defining come from sbt (which is really clear for us if we don't use `Keys`, but not for people reading this build and unfamiliar with some less popular keys).
* I can do `val moduleID = Keys.moduleID.value`. Otherwise I would have to invent a similar name which doesn't clash with `moduleID`. I could choose between `id` (id of what? project? module?), `module` (which is not really true, it's an id of a module) or another variation, but this is *weird* and I don't want to spend time thinking how to name local variables that are straightforward to identify.
  
Does this make sense?

build.sbt
jvican: Good catch.

jvican:
@dwijnand Feedback addressed. I invite you to have another look. Before merge, I'll fixup the last commits together to make sure everything passes CI.

project/BuildPlugin.scala
jvican: Good catch.

project/Header.scala
jvican: Done.

CONTRIBUTING.md
```diff
@@ -31,8 +31,7 @@ Once you understand the basics of incremental compilation, start having a look
 at open tickets you can help with. All issues are labelled and will give you an
 idea about its difficulty and scope.
 
-Hacking on Zinc should not seem like a difficult task. Zinc does not implement
-a compiler, it defines the logic to analyse dependencies based on the compiler
+Hacking on Zinc should not seem like a difficult task. Zinc does not implement a compiler, it defines the logic to analyse dependencies based on the compiler
```
jvican, line 34: This change is not intentional, but does not change the guide. If it clutters up the diff, I can remove it...

CONTRIBUTING.md
```diff
@@ -55,6 +54,48 @@ bridge, you don't need to run these benchmarks.
...
+#### Project structure
+
+As of now, the current project structure is not as simple as it can be and we
+believe that it can be simpler in the future. However, as of now, no work is
+happening in this area because it's deemed to have low impact on the overall
+quality of the project.
+
+How is the Zinc build structured? Let's see it.
+
+|Project name| Project description|
```
jvican, line 68: This part should be carefully reviewed. I did my best to explain what these projects are used for, but this structure pre-dates me, so someone has to confirm. Related to https://github.com/sbt/zinc/issues/286.

project/BuildPlugin.scala
dwijnand: > I don't want to accidentally be referring to a local variable named this way because it's a source of bugs.

Could you provide an example?

> It's clear that the keys I'm defining come from sbt

I don't think that's a strong enough reason to diverge in style.

> I can do `val moduleID = Keys.moduleID.value`

You can still do that with `import Keys._`.

---

Another reason not to go with this new style is that when we have slash syntax the build code and shell notation will be unified, so we'll want to use in both cases:

```scala
Compile / javacOptions
```

not

```scala
Compile / Keys.javacOptions
```

jvican: > You can still do that with import Keys._.
  
No, you cannot. If you do `import Keys._`, `moduleID` will be in scope, so if I define `val moduleID = moduleID.value` Scala will think it's a recursive definition and will fail both at compile-time and runtime (in case you ever make it to typecheck).
  
> Could you provide an example?
  
Say I define a local method called `publish` that defines how a thing is going to be published. If I've done `import Keys._`, `publish` will resolve to different terms depending on the scope I use them, which is misleading and a source of bugs. I don't want my code to change its meaning depending on the place I use my names. Not only it's not refactor friendly, it's confusing and dangerous.
  
The problem gets worse when you don't even know that you're importing a term from `Keys.scala` that is conflicting with a term you've defined. What about `artifact`, `doc`, `mainClass`, `run` and the other bazillion of names that are really likely to appear in my build logic and conflict with sbt keys definitions?
  
> Another reason not to go with this new style is that when we have slash syntax the build code and shell notation will be unified
  
That's a matter of style. I don't mind the second version. My arguments here have to do with semantics and user readability (in terms of not having name clashes -- when users read they don't need to think if a term is defined at the end of the file, in an enclosing scope or comes from sbt -- they have a clear idea that `Keys.artifact` means the setting key from sbt).

CONTRIBUTING.md
dwijnand: Looks good enough to me.

CONTRIBUTING.md
dwijnand: It's ok.

project/BuildPlugin.scala
```diff
@@ -0,0 +1,357 @@
...
+import com.typesafe.tools.mima.plugin.MimaKeys
+
+object BuildPlugin extends AutoPlugin {
+  override def requires = sbt.plugins.JvmPlugin
+  override def trigger = allRequirements
+  val autoImport = BuildAutoImported
+
+  override def projectSettings: Seq[Def.Setting[_]] = BuildImplementation.projectSettings
+  override def buildSettings: Seq[Def.Setting[_]] = BuildImplementation.buildSettings
+  override def globalSettings: Seq[Def.Setting[_]] = Nil
```
dwijnand: redundant

project/BuildPlugin.scala
```diff
@@ -0,0 +1,357 @@
...
+    def runBenchmarks(benchmarkProject: Project): Command = {
+      val dirPath = BuildAutoImported.benchmarksTestDir.getAbsolutePath
+      val projectId = benchmarkProject.id
+      val runPreSetup = s"$projectId/run $dirPath"
+      val runBenchmark = s"$projectId/jmh:run -p _tempDir=$dirPath -prof gc"
+      val tearDownResources = s"$projectId/tearDownBenchmarkResources"
+      Command.command("runBenchmarks")(st => runPreSetup :: runBenchmark :: tearDownResources :: st)
+    }
+
+    def all(bridge: Project, interface: Project, apiInfo: Project, bench: Project): Seq[Command] = {
```
dwijnand, line 169: Could you make these `ProjectReference` please? We don't need the actual project, just a reference to it.

jvican:
For some weird reason, GitHub does not show the latest commit and its CI status.

dwijnand (changes requested):
For some reason I can't respond to the thread above, so I'll respond here:

> No, you cannot. If you do import Keys._, moduleID will be in scope, so if I define val moduleID = moduleID.value Scala will think it's a recursive definition and will fail both at compile-time and runtime (in case you ever make it to typecheck).

I said you can still do `val moduleID = Keys.moduleID.value`.  

> Say I define a local method called `publish` that defines how a thing is going to be published. If I've done `import Keys._`, `publish` will resolve to different terms depending on the scope I use them, which is misleading and a source of bugs. I don't want my code to change its meaning depending on the place I use them. Not only it's not refactor friendly, it's confusing and dangerous. 
 The problem gets worse when you don't even know that you're importing a term from `Keys.scala` that is conflicting with a term you've defined. What about `artifact`, `doc`, `mainClass`, `run` and the other bazillion of names that are really likely to appear in my build logic and conflict with my definitions?  

That's Scala for you, nothing to do with sbt per-se. You need to play with scala scopes and/or find other names.  

> That's a matter of style. I don't mind the second version. My arguments here have to do with semantics and user readability (in terms of not having name clashes -- when users read they don't need to think if a term is defined at the end of the file, in an enclosing scope or comes from sbt -- they have a clear idea that `Keys.artifact` means the setting key from sbt).

  I don't think the user cares where a key is defined, and would find it more user-friendly if the same syntax as the shell were used.

project/BuildPlugin.scala
jvican: We cannot because we need `id`, and `id` is not defined in `ProjectReference`.

jvican:
> I said you can still do val moduleID = Keys.moduleID.value.  
  
Sure, but then are you not inconsistent? You use a wildcard import and then you use the prefix to disambiguate -- it's weird.
  
Also, this assumes you know what `import sbt.Keys._` does import. *We know*, but most users don't.
  
> That's Scala for you, nothing to do with sbt per-se. You need to play with scala scopes and/or find other names.  
  
Agreed, nobody said is sbt specific. But it's an important problem. I want to write my build so that scoping problems are minimized and I don't need to be constantly thinking about variations of my names (that are stupid and can bring bugs, as said before) just to avoid name clashes produced by unnecessary wildcard imports.
  
I'm not the only one that does it this way, see this comment by Li Haoyi: https://contributors.scala-lang.org/t/proposed-syntax-for--root-/1035/37?u=jvican.

>   I don't think the user cares where a key is defined, and would find it more user-friendly if the same syntax as the shell were used.
  
Well, I don't know if they care, but I believe they should care because the place where it's defined alters semantics of their code...
  
@dwijnand PR is updated with your last feedback. Thanks!

project/BuildPlugin.scala
dwijnand:
```scala
implicit def projectToRef(p: Project): ProjectReference = LocalProject(p.id)
```

If only this returned `LocalProject` instead of `ProjectReference`..

jvican: @dwijnand Yeah... can maybe this new method be added in sbt 1.x? I think it's bincompat:
  
```scala
@ class A 
class B edefined class A
@ class B extends A 
defined class B
@ implicit def stringToA(s: String): A = new A 
defined function stringToA
@ implicit def stringToB(s: String): B = new B 
defined function stringToB
```

dwijnand: yep, I think we should consider it.

build.sbt
```diff
@@ -288,7 +288,7 @@ lazy val zincCompileCore = (project in internalPath / "zinc-compile-core")
 // defines Java structures used across Scala versions, such as the API structures and relationships extracted by
 //   the analysis compiler phases and passed back to sbt.  The API structures are defined in a simple
 //   format from which Java sources are generated by the sbt-contraband plugin.
-lazy val compilerInterface = (project in internalPath / "compiler-interface")
+lazy val compilerInterface = (project in file("compiler-interface"))
```
eed3si9n: My opinion is that some datatype might be exposed, but in general the notion of compiler interface (abstraction over Scala compiler version) should be considered internal to Zinc. In other words, unless otherwise stated, we reserve the right to change compiler interface signature from Zinc 1.0 to 1.1, as long as it's not exposed to Zinc API.

project/BuildPlugin.scala
eed3si9n: Inside of an AutoPlugin definition it is customary to `import Keys._`. I tend to import any autoImports from plugins that I works on to emulate what one would have in `build.sbt`.

eed3si9n (changes requested):

build.sbt
jvican: > My opinion is that some datatype might be exposed, but in general the notion of compiler interface (abstraction over Scala compiler version) should be considered internal to Zinc. 
  
But the compiler interface does not only expose abstractions over Scala versions, it has all the facilities that are afterwards exposed to downstream users (loggers, reporters, scala instance, etc). Lots of these things also leak to sbt's API.

> In other words, unless otherwise stated, we reserve the right to change compiler interface signature from Zinc 1.0 to 1.1, as long as it's not exposed to Zinc API.

This rule you propose is unexpected to me and doesn't satisfy the bincompat guarantees that other Scala tools do, like scalac. It's not even the same than the one for sbt, right?
  
I think we should enforce that binary compatibility is not broken in minor releases, especifically for Zinc, which is not only used by sbt but also by other builds tools that also have strong bincompat guarantees.
  
In any case, binary compatibility aside, I think this change is good. We currently have two top-level modules (one `zinc` and another one `zinc-compile`). The compiler interface is in my opinion the most important module in the project, and I think we should highlight it by placing it in the root of the project. It doesn't make sense to me that zinc-compile looks, at first glance, more important than the compiler interface.

project/BuildPlugin.scala
jvican: Yes, it is customary @eed3si9n. But I believe that, based on my technical explanations before, this particular implementation is superior than the "customary practice". I don't see any technical benefit on making the build plugin implementation be more similar to `build.sbt`. Perhaps you can elaborate on that?

jvican:
Is there anything I can do here to move this forward? I've replied to all your concerns/questions.

jvican:
Commits have been squashed. It looks like GitHub is not catching up with the parent branch.

jvican:
I'll rebase this whenever #438 is merged. The changes done by this PR start from https://github.com/sbt/zinc/pull/428/commits/ca17c4a56abb5a28bd887123901ad1bc52d06822.

build.sbt
eed3si9n: Let's continue https://github.com/sbt/zinc/pull/438

jvican:
@eed3si9n @dwijnand This PR has been updated. Can you have a look?

project/BuildPlugin.scala
jvican: You both asked me to remove this. I did, but in the other PR because rebasing otherwise was hell. https://github.com/sbt/zinc/pull/429
  
I think this is good to go.

eed3si9n: Dale's been looking at this PR more than me, so I'll defer my vote to him.
@dwijnand Is this PR good to go as is? if not could you make a list of pending items?

CONTRIBUTING.md
```diff
@@ -55,6 +54,55 @@ bridge, you don't need to run these benchmarks.
...
+|------------|--------------------|
+|zincRoot| The root of the project. Aggregates all projects except the benchmarks.|
+|zinc|The user-facing Zinc incremental compiler.|
+|zincTesting|The project that defines testing facilities.|
+|zincCompile|A thin wrapper that provides doc capabilities.|
+|zincPersist|The project that persists incremental compiler's data into a binary file.|
+|zincCore|The project that defines relations, analysis, stamps, and essential core utils.| 
+|zincBenchmarks|The project that defines the benchmarks.|
+|zincIvyIntegration|The project that defines the ivy utilities to fetch compiler bridges.|
```
Duhemm, line 77: I thought that we agreed to rename that project to `zincLmIntegration`, but it looks like it didn't make it to the PR that superseded mine. See: https://github.com/sbt/zinc/pull/335#issuecomment-315019318

Maybe now is a good time to do it?

CONTRIBUTING.md
```diff
@@ -55,6 +54,55 @@ bridge, you don't need to run these benchmarks.
...
+#### Build-specific commands/keys
+
+The sbt build defines several keys that help contributors run and test Zinc.
+Zinc's build requires the compiler bridges to be published before tests are run
+(compiler bridges are compiler-specific Scala sources that need to be fetched
+to perform incremental compilation).
+
+|Key|Use|
+|---|---|
+|crossTestBridges|Runs compiler bridge unit tests for all scala versions.|
```
Duhemm, line 102: Will it still be true after #453 gets in? I don't know why we don't test the bridge with 2.13.0-M2.

CONTRIBUTING.md
```diff
@@ -55,6 +54,55 @@ bridge, you don't need to run these benchmarks.
...
+|zincCore|The project that defines relations, analysis, stamps, and essential core utils.| 
+|zincBenchmarks|The project that defines the benchmarks.|
+|zincIvyIntegration|The project that defines the ivy utilities to fetch compiler bridges.|
+|zincCompileCore|The project that interfaces with the compiler API and provides compilation capabilities.| 
+|zincApiInfo|The project that defines name hashes and provides way to interpret api changes.|
+|zincClassfile|The project that parses class files to provide Java incremental compilation.|
+|zincClasspath|The project that provides basic utilities to load libraries with classloaders and represents Scala instances.|
+|zincScripted|The project that defines the scripted logic to run Zinc's integration test suite.|
+|compilerInterface|The public binary interface used to connect the bridges with the Zinc modules. It is written in Java and uses Contraband.|
+|compilerBridge|The module that defines the compiler plugin phases that provide incrementality for all Scala versions.|
```
Duhemm, line 84: We'll need to add `compilerBridgeTest` to the table after #453 gets in.

CONTRIBUTING.md
jvican: #453 has to be rebased on top of this, and Eugene has to do whatever changes do not hold in his PR.

CONTRIBUTING.md
jvican: Same comment as before.

CONTRIBUTING.md
jvican: I'd very much prefer not to touch this PR again, I've already done too many changes :smile:

eed3si9n:
Only bug fixes (or improvements needed for bug fixes) goes into 1.0.x branch, so this PR imo should target 1.x.

jvican:
@eed3si9n That policy has not been discussed between you and I at any point, and that does not correspond with other PRs that have been merged into the 1.0.x line.
  
This PR and the other one have been here for more than 25 days waiting for a review, and hasn't yet been reviewed since our last meeting three weeks ago... I've put so much time into rebasing the PRs and adapting it to other changes, that I'm not willing to do it again.

eed3si9n:
The 1.x vs 1.0.x convention was discussed here - https://github.com/sbt/sbt-contrib/blob/master/src/main/markdown/2017-08-16.md#general-release-cycle.
If there has been 1.0.x merges that are not for hot fixes, then likely they were mistakes.

As per the prioritization of PRs, the themes around each releases are discussed in sbt 1 meetings, and discussions in last few meetings have been wrapping things up for 1.0.4 and 1.1-RC. It's not uncommon to have PR sit on the queue for a while if they are not critical part of the milestone. For example https://github.com/sbt/sbt/pull/3621 has been sitting since Oct 8 (I think it's ready to merge, but hey).

jvican:
This is not sbt, this is Zinc, and the Scala Center has proved more than several times its great involvement in this project. To the best of my knowledge, we both have an spoken deal to drive the direction of this project, which is important because the Scala community needs our collaboration. That means that whatever it's applicable to sbt/sbt, it's not necessarily applicable here. As you said in numerous occasions, sbt != zinc and we're not Lightbend.
  
From reading your message, it looks like you're the only one that chooses the priority of the PRs and I'm sorry to tell you but that's not how it works. Your actions have consequences in my work, and that's why it's good to discuss things before taking unilateral actions.
  
Your wish is also inconsistent, lots of things that are not bug fixes have been merged into the 1.0.x line, sorry.
  
I don't think I'm asking too much when I say that I want my work to be respected. If it's not reviewed on time and keeps being delayed, it's not my fault. I've been rebasing this thing and I've already given up on lots of design decisions just to make you happy. Same with the bincompat restrictions that you want and I don't.

jvican:
And for those reading this discussion, here's where our disagreement started: https://github.com/sbt/zinc/pull/453.

jvican:
You have to understand that we're working on a team, and if you want to make a change at the last minute and override all the work that other team members have put into Zinc, it's not nice. It means I need to stop the world, come here, and update all the stuff just because you think your PR has more priority than mine. And this wouldn't have happened if you had reviewed my PRs in due time (and ironically you ping me several times in less than one week to review your PR because it's blocking a release).
  
I've done **many** concessions in the past weeks to accommodate your likes and what not (including your controversial opinions on bincompat), and I see no effort from your side to respect the work and time I am investing in this project, sometimes on my free time.
  
If you considered this was a change required in 1.x, you should have said that one month ago, and discussed with me this rule (not take the decision unilaterally in some old sbt meeting minutes I didn't attend). Not doing so is wasting my time and our time. You can at least ping me in Gitter to notify me your plan to ship something in 1.x that overrides my changes.
  
Being in a team means agreeing to make things happen together, not taking unilateral actions and then asking others to catch up with them.
  
It's also important that you don't assume that the sbt release process is the same. If we're making this thing happen, and improving the incremental compiler for the whole community, then it's only reasonable we take different decisions (because the sbt team is not the only one in charge of this project).
  
I think support for Scala 2.13.x is important. But I'm only happy to merge your fix to support Scala 2.13.x if you are the one that rebases all my changes in 1.x (keeping git history) once your PR gets merged. As an author of these two PRs, I will obviously address more feedback if I get it, but I refuse to swallow the consequences of your poor management of this issue.
  
I know you're used to take decisions in sbt/sbt, but I hope you don't realistically hope to dictate how development of Zinc should be when more than one party is involved into it.
  
And I say this in the nicest of the tones. I'm just being honest and direct.

eed3si9n:
> This is not sbt, this is Zinc, and the Scala Center has proved more than several times its great involvement in this project. To the best of my knowledge, we both have an spoken deal to drive the direction of this project, which is important because the Scala community needs our collaboration. That means that whatever it's applicable to sbt/sbt, it's not necessarily applicable here. As you said in numerous occasions, sbt != zinc and we're not Lightbend.

Agreed. And it's worth repeating that both Lightbend Tooling team and Scala community are grateful for Scala Center's contributions.

> From reading your message, it looks like you're the only one that chooses the priority of the PRs and I'm sorry to tell you but that's not how it works. Your actions have consequences in my work, and that's why it's good to discuss things before taking unilateral actions.

I've seen this claim repeated a few times, but I see it in a different way. During the development leading up to sbt 1 we've held multiple planning meetings together to discuss the todo list on what goes into RC-x. After sbt 1.0, there's been sbt 1 meetings every week since August to coordinate what goes in, when, and who wants to work on them. We discuss Zinc task planning here too. So even though Lightbend is the owner of Zinc, and I'm the lead, the governance of sbt or Zinc has been relatively open.

> Your wish is also inconsistent, lots of things that are not bug fixes have been merged into the 1.0.x line, sorry.

There's been three hot fix releases of Zinc thus far from 1.0.x branch: 1.0.1, 1.0.2, and 1.0.3.
As far as I know, all of the changes that went into them were for bug fixes that we shipped together with sbt 1.0.x hot fixes. If there has been any other PRs, then it had to be low/zero risk (like adding headers) or a mistake. Could you point to an example?

> I don't think I'm asking too much when I say that I want my work to be respected. If it's not reviewed on time and keeps being delayed, it's not my fault. I've been rebasing this thing and I've already given up on lots of design decisions just to make you happy. Same with the bincompat restrictions that you want and I don't.

I don't think it's a matter of respect. If there's an objectively verifiable bug in the code (e.g. using the result of a method that sometimes returns `null`, causing NPE), I think the review process for a fix would be straightforward since the fix can also be verified. Refactoring starts to involve matters of taste, and it could take a while to converge choices/decisions. Given the 59 comments, it's been given enough attention.

I also understand that PR taking a long time is frustrating, and we should work towards improving that.

eed3si9n:
> You have to understand that we're working on a team, and if you want to make a change at the last minute and override all the work that other team members have put into Zinc, it's not nice. It means I need to stop the world, come here, and update all the stuff just because you think your PR has more priority than mine. And this wouldn't have happened if you had reviewed my PRs in due time (and ironically you ping me several times in less than one week to review your PR because it's blocking a release).

Since there's been multiple conversations on Gitter etc around 1.0.x or sbt 1.0.x hot fixes, I was totally under the impression that you knew of the 1.x and 1.0.x distinction and were on board with the idea.

gitter1

gitter3

If the criteria or the process was somehow not clear, I think it was a miscommunication on my part. As I wrote in the previous comment too, "what goes into what release" is one of the things we discuss in the sbt 1 meetings. In other words, it's not decided unilaterally. If you can't attend it personally, you're welcome to send someone else, write in a email, or schedule chat with me afterwards.

> I've done many concessions in the past weeks to accommodate your likes and what not (including your controversial opinions on bincompat), and I see no effort from your side to respect the work and time I am investing in this project, sometimes on my free time.

See https://github.com/sbt/zinc/pull/428#issuecomment-346507814. I acknowledge that long PR process can be frustrating, but I don't think it's a matter of respect. The PR was sent in on Oct 12, and within 24h @dwijnand has given lots of feedback. It dropped from our radar for a while when we couldn't bring it to consensus in time. Other PRs during this period were reviewed and merged when it was simple (like related https://github.com/sbt/zinc/pull/438).

Also note that 2 days ago I've [asked][2] Dale to come up with a pending list if he thinks there's any work remaining on this PR when you pinged me.

> If you considered this was a change required in 1.x, you should have said that one month ago, and discussed with me this rule (not take the decision unilaterally in some old sbt meeting minutes I didn't attend). Not doing so is wasting my time and our time. You can at least ping me in Gitter to notify me your plan to ship something in 1.x that overrides my changes.

You're right that either Dale or I should've spotted that this PR targets 1.0.x branch much sooner. This was an oversight on our part.

> Being in a team means agreeing to make things happen together, not taking unilateral actions and then asking others to catch up with them.

Agreed. To be fair though, there were attendees from Scala Center in the [initial meeting][1] when we discussed the branching strategy, so from my point of view these decisions were made though discussion not by fiat.

> It's also important that you don't assume that the sbt release process is the same. If we're making this thing happen, and improving the incremental compiler for the whole community, then it's only reasonable we take different decisions (because the sbt team is not the only one in charge of this project).

Yes. My goal for Zinc is to be a useful incremental compiler for any build tool, which might take different decisions in terms of incrementality, strictness on repeatability (Bazel / Pants comes to mind here) etc. The strategy sbt uses for its releases is SemVer, and it does carry over to Zinc in many ways, but I am hopeful that we can achieve these goals.

> I think support for Scala 2.13.x is important. But I'm only happy to merge your fix to support Scala 2.13.x if you are the one that rebases all my changes in 1.x (keeping git history) once your PR gets merged. As an author of these two PRs, I will obviously address more feedback if I get it, but I refuse to swallow the consequences of your poor management of this issue.

If what's needed to ship Scala 2.13.x is hinging on time to `git`, both Dale and I are happy to take on the task. After merging 1.0.x into 1.x, hopefully it would be a straight `merge`. 

  [1]: https://github.com/sbt/sbt-contrib/blob/master/src/main/markdown/2017-08-16.md#general-release-cycle
  [2]: https://github.com/sbt/zinc/pull/428#discussion_r152179203

dwijnand:
> If what's needed to ship Scala 2.13.x is hinging on time to `git`, both Dale and I are happy to take on the task. After merging 1.0.x into 1.x, hopefully it would be a straight `merge`.

See https://github.com/sbt/zinc/pull/456

PR #429, opened 2017-10-12
Set up sbt-release-early and change build semantics

jvican:
This pull request aims at:
  
- Setting up a sbt-release-early workflow, notably using the advantages of sbt-dynver.
- Changing the build semantics to get a faster workflow with scripted.
- Removing unnecessary/dead code.

Builds upon sbt/zinc#428 and needs to be merged afterwards. When rebase, the actual diff size will be small.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/429/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

jvican:
f

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/429/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

project/plugins.sbt
```diff
@@ -1,9 +1,7 @@
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.27")
-addSbtPlugin("org.foundweekends" % "sbt-bintray" % "0.5.1")
-addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.0-M1")
-addSbtPlugin("com.lucidchart" % "sbt-scalafmt" % "1.12")
```
jvican, line 4: This is already in sbt-houserules.

jvican:
This pull request adds a cached publish local that surprisingly allows us to simplify the build by removing ad-hoc commands and workarounds (`scriptedUnpublished`). From the commit message:
```
`cachedPublishLocal` is a special task that avoids publishing if an
artifact has already been published and the publication time is greater
than the changes to the source directories.

In order to find this out, `cachedPublishLocal` resolves the artifact
intransitively and in a project-specific directory. This directory is
also the used to resolve everything from scripted. This resolution is
fast because it assumes that no other task is able to publish that jar
to that concrete resolver. As a result, this assumption allows us to
remove the global dependency lock.

Running `cachedPublishLocal` in the root directory takes from 0 to 1
second. When a new change is written into a project's source file or a
resource file, `cachedPublishLocal` publishes only the project that
received the changes. This is possible because the optimizer is not
allowed and jars are independent from each other.

In this setup, `scripted` uses `cachedPublishLocal`. `scripted`
**needs** the compiler bridges for all the Scala versions, so if you
change the bridge or this is the first time you use this setup, you must
run `+compilerBridge/cachedPublishLocal`.
```

build.sbt
```diff
@@ -148,11 +148,7 @@ lazy val compilerInterface = (project in internalPath / "compiler-interface")
   .settings(
     // javaOnlySettings,
     name := "Compiler Interface",
-    // Use the smallest Scala version in the compilerBridgeScalaVersions
-    // Technically the scalaVersion shouldn't have any effect since scala library is not included,
-    // but given that Scala 2.10 compiler cannot parse Java 8 source, it's probably good to keep this.
-    crossScalaVersions := Seq(scala210),
-    scalaVersion := scala210,
+    crossScalaVersions := bridgeScalaVersions,
```
eed3si9n: During the release, wouldn't this result to double publishing?

build.sbt
```diff
@@ -202,8 +187,6 @@ lazy val compilerBridge: Project = (project in internalPath / "compiler-bridge")
       .value,
   )
 
-// defines operations on the API of a source, including determining whether it has changed and converting it to a string
```
eed3si9n: Why are we removing these comments?

eed3si9n (changes requested):

build.sbt
jvican: Because it's documented in the contributing guide, better.

build.sbt
jvican: Good question, I think it would, I'll try it tomorrow and fix it if that's the case.

Issue #433, opened 2017-10-14
Make classpath hashing more lightweight


fommil:
It would be good to have the option to bypass the calculation of a file hash code if the size has not changed and the timestamp has not changed (just the timestamp should be ok). This has huge performance potential... avoiding doing work at all.

Possibly there could be some trigger that the file timestamp must be older than an hour or something... to avoid user interaction problems.

NIO allows a fast mechanism for collecting bulk metadata, it should be possible to traverse a directory and get all of this information very quickly before deciding what needs to be rehashed.

fommil:
See link for a proof of concept that basically removes the bottleneck entirely. Should really be implemented in the zinc layer but I have no way of doing that and testing it in a live sbt.

jvican:
> It would be good to have the option to bypass the calculation of a file hash code if the size has not changed and the timestamp has not changed (just the timestamp should be ok). This has huge performance potential... avoiding doing work at all.
  
We can use the checksums in the same jars to detect whether something has changed or not (all zip files have checksums in their headers). In fact, it's not clear why we compute a hash of the jars instead of reusing the checksum in the header.

fommil:
@jvican ha! That's a pretty good idea. Although I still think file size / timestamp is a pretty damn good low bar for everything. But having custom hashes sounds like an excellent idea, especially if the file already has a checksum!

fommil:
For completeness I raised PRs against launcher and sbt to enable monkey patching to let me play around. Until they are working and merged I'm afraid I can't help. 

https://github.com/sbt/launcher/pull/43

https://github.com/sbt/sbt/pull/3643

jrudolph:
I second the idea of using less safe but faster comparison options. Not sure about zip file signatures. Afaik zip files don't contain a hash over all the content but only crc32s for each file which would mean that you would have to read (probably significantly) into the content to collect all the hashes (but would probably still be faster than reading all the data). Getting just file system metadata should be still faster.

fommil:
Clear reproduction in https://github.com/cakesolutions/sbt-cake/tree/sbt-perf-regression

type `sbt cpl` to compile all test phases.

jvican:
> Getting just file system metadata should be still faster.
  
This is not viable. System metadata is really unreliable, and I would not like the incremental compiler to misbehave because of it.
  
I think that reading checksums from the zip file could be done in a fast way with some tweaks.

jrudolph:
[...]

PR #434, opened 2017-10-14
Timestamp hashes

fommil:
Because of  https://github.com/sbt/sbt/issues/363 I am unable to say anything about the perf characteristics of this PR.

I guess it is addressing the same problem that https://github.com/sbt/zinc/pull/371 is addressing and a hacky alternative to #433 (which is probably the correct solution).

What I thought was an I/O bottleneck in https://github.com/sbt/io/pull/79 turned out to be a CPU hashcode bottleneck when hashing *classpath files*, not source files. YourKit doesn't pick up the hash/crypto work, showing it as InputFilterStream. Given the many many bugs in the entire scala tooling chain when it comes to `.jar` files mutating on disk, see https://gitlab.com/fommil/class-monkey/blob/master/README.md, I really don't think SHA-1 is needed here. Timestamps (and file size) should be more than enough.

Also, this hashing could be done *in parallel*. Even if it's as simple as using the terrible `.par`.

### DIFF
--- a/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
@@ -182,7 +182,9 @@ object MixedAnalyzingCompiler {
       extra: List[(String, String)]
   ): CompileConfiguration = {
     val classpathHash = classpath map { x =>
-      FileHash.of(x, Stamper.forHash(x).hashCode)
+      val mod = x.lastModified
+      val length = x.length // TODO reverse
+      FileHash.of(x, (mod ^ length).##)
     }
     val compileSetup = MiniSetup.of(
       output,


lightbend-cla-validator:
Hi @fommil,

We see that you have signed the Lightbend Contributors License Agreement before, however, the CLA has changed since you last signed it.
Please review the new CLA and sign it before we proceed with reviewing this pull request:

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/434/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

fommil:
BTW, I'm not signing the Lightbend CLA... you can have my contributions under the terms of the BSD-3, the license of this project. Otherwise it is not a symmetric and fair agreement.

fommil:
I'd like to do some more work on this, but I'm basically blocked until I can try out my changes locally without having to build the entire sbt ecosystem from scratch... I just want to be able to stick a jar at the right place in the sbt classpath.

fommil:
closing in preference of #433

romanowski:
@fommil We also have problems with classpath hashing (you can probably imagine) and caching was best approach for us. 
I've open PR ( #427) that allows client o implement classpath hashing on thier side (and I think it should be possible to override this from sbt level).

fommil:
Typical @romanowski : "yeah, we already seen this and fixed it months ago... :man_shrugging: " :grin:

Issue #405, opened 2017-09-15
Provide a thread-safe version of `AnalysisCallback`


jvican:
`AnalysisCallback` is as of now not thread-safe. In @dragos's words:
>  Since we run several workers at the same time, and each worker has its own Global and analysis extraction logic, they perform updates concurrently. We used to override the compileIncrementalTask and fork some parts of Sbt to get to call our own implementation, but I think there should be a cleaner way, especially since xscbti.AnalysisCallback is an interface so it would be nice to be able to provide a different implementation.
  
This feature enables future parallel Scala compilers (like [Hydra](https://www.triplequote.com/hydra/)) to use the incremental compiler out of the box.
  
The place where the implementation should be provided is https://github.com/sbt/zinc/blob/1.x/internal/zinc-core/src/main/scala/sbt/internal/inc/Compile.scala, probably in an independent file.
  
The tests can be added in the `zincCore` project.

dragos:
I didn't check yet, but I have the impression IntelliJ would need their own implementation of `AnalysisCallback` as well. Maybe a simple way to specify it from the outside would be enough, even if Zinc itself doesn't provide a thread-safe version.

romanowski:
So far Intellij uses custom version of zinc based on 0.13.5 but AFAIK it will be migrated to zinc 1.0.
Intelliij.

Zinc is also used as scalac provider even if Intellij incremental compiler is used (and I think this was your case). [Here](https://github.com/JetBrains/intellij-scala/blob/idea172.x/jps-plugin/src/org/jetbrains/jps/incremental/scala/local/IdeaIncrementalCompiler.scala#L56-L66) you can find a piece of code that should be thread safe in current version and [here](https://github.com/JetBrains/intellij-scala/blame/idea173.x/jps-plugin/src/org/jetbrains/jps/incremental/scala/local/IdeaIncrementalCompiler.scala) code after migration to zinc 1.0.

cc @jastice

PR #438, opened 2017-10-20
Add headers and move compiler interface

jvican:
This has been split from https://github.com/sbt/zinc/pull/428/ to ease review.

It only does the following:

- Add sbt-header back.
- Add headers to missing files and checks in CI.
- Update sbt-scalafmt to the latest version.
- Move the compiler-interface to the base directory.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/438/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

build.sbt
```diff
@@ -288,7 +288,7 @@ lazy val zincCompileCore = (project in internalPath / "zinc-compile-core")
 // defines Java structures used across Scala versions, such as the API structures and relationships extracted by
 //   the analysis compiler phases and passed back to sbt.  The API structures are defined in a simple
 //   format from which Java sources are generated by the sbt-contraband plugin.
-lazy val compilerInterface = (project in internalPath / "compiler-interface")
+lazy val compilerInterface = (project in file("compiler-interface"))
```
eed3si9n: Porting my comment from https://github.com/sbt/zinc/pull/428:

My opinion is that some datatype might be exposed, but in general the notion of compiler interface (abstraction over Scala compiler version) should be considered internal to Zinc. In other words, unless otherwise stated, we reserve the right to change compiler interface signature from Zinc 1.0 to 1.1, as long as it's not exposed to Zinc API.

eed3si9n (changes requested):

build.sbt
eed3si9n: > But the compiler interface does not only expose abstractions over Scala versions, it has all the facilities that are afterwards exposed to downstream users (loggers, reporters, scala instance, etc). Lots of these things also leak to sbt's API.

This goes to show that we should clearly define the boundary of what "Zinc API" is.
Logger, as you point out probably leaks out so it's likely part of the public API, so that should be included.

This could gradually increase later, but I think we should keep the promise of Zinc 1.0 to be that it's able to accept an array of Scala and Java source files repeatedly, and produce correct `*.class` files as the result. Everything else should be internal detail.


>> In other words, unless otherwise stated, we reserve the right to change compiler interface signature from Zinc 1.0 to 1.1, as long as it's not exposed to Zinc API.
> 
> This rule you propose is unexpected to me and doesn't satisfy the bincompat guarantees that other Scala tools do, like scalac. It's not even the same than the one for sbt, right?

sbt's binary compatibility responsibilities are mainly for compiled plugins. For that we guarantee binary compatibilities for `sbt.*` except `sbt.internal.*`.

Since the intended consumer of the compiler interface is the compiler bridge (`xsbt.*` classes), we should be able to add/remove things in the interface for various reasons.

eed3si9n: > In any case, binary compatibility aside, I think this change is good. We currently have two top-level modules (one zinc and another one zinc-compile). The compiler interface is in my opinion the most important module in the project, and I think we should highlight it by placing it in the root of the project. It doesn't make sense to me that zinc-compile looks, at first glance, more important than the compiler interface.

`zinc` contains Zinc API, and `zinc-compile` should contain other related features such as ScalaDoc wrapper and Scala REPL wrapper, both intended as library someone can use.

jvican: > classes), we should be able to add/remove things in the interface for various reasons.
  
I think this is where we're not on the same page. The compiler interface contains interfaces that are used by all Zinc modules, not only the compiler bridge. You're probably right that the `xsbt.api` is not exposed and we can probably break it if we want to, but the majority of classes we cannot change in an binary incompatible way: `IncOptions`, `CompileOptions`, `Companions`, `NameHash`, etc.

> This could gradually increase later, but I think we should keep the promise of Zinc 1.0 to be that it's able to accept an array of Scala and Java source files repeatedly, and produce correct *.class files as the result. Everything else should be internal detail.

> sbt's binary compatibility responsibilities are mainly for compiled plugins. For that we guarantee binary compatibilities for sbt.* except sbt.internal.*.
  
My main fear is that breaking changes in the compiler interface will break downstream's user code (in Pants, Gradle, and other build tools that depend on us). We put a lot of effort into creating Java-friendly APIs that other tools like Maven can use, and not having strong bincompat guarantees on that code now is counterintuitive.
  
These days, I'm becoming more and more conservative about breaking APIs, and in this case I believe the consequences of breaking an API are worst than the advantages of improving it. I'm not sure we can expect our users to get compiler errors every time we make a new release of Zinc, for example.
  
It seems you want to cherry-pick concrete classes of the compiler interface and only ensure bincompat on those. I wonder two things:
  
1. Is this possible at all? Our binary interfaces are so tightly coupled, that I believe it's challenging to cherry-pick interfaces in a reasonable and general way.
2. What happens if you modify an interface that affects sbt plugins? It seems that you're saying that the compiler interface should be considered `sbt.internal`, but isn't this going to create frustration for people interfacing with this code? Examples that come to mind are [`sbt-check`](https://github.com/jeffreyolchovy/sbt-check/) and Triplequote's sbt plugin (/cc @dragos @dotta). In my opinion, a build tool is all about compiling, and if we have bincompat guarantees at all they should focus on the compilation APIs.

dotta: Thanks a lot for cc-ing us @jvican!

I'm currently not entirely clear on what are the source/binary incompatible changes that you may want to do now or in the future, but I can only agree with the general sentiment of this statement:

> These days, I'm becoming more and more conservative about breaking APIs, and in this case I believe the consequences of breaking an API are worst than the advantages of improving it.

In our sbt plugin we replace the sbt compiler-bridge with our own hydra-bridge. The main change is the implementation of `CompilerInterface`, as it needs to call the Hydra Scala compiler instead of the vanilla Scala one. In the `HydraPlugin` we do depend on things like the `ScalaInstance`, `Compiler.Inputs`, `CompileOptions` and we also had to replace some sbt internals in order to swap the `AnalysisCallback` and filter compiler options that shouldn't trigger recompilation (these hacks won't be needed once https://github.com/sbt/zinc/issues/405 and https://github.com/sbt/zinc/issues/406 are resolved).

Let me know if there is anything I should check and report that can help the discussion.

jvican:
@eed3si9n Can you review this please?

eed3si9n (approved):

eed3si9n:
LGTM

Issue #415, opened 2017-09-26
Javadoc may not log all errors


raboof:
When debugging an instance where javadoc produced the error "javadoc: error - invalid flag: -target", this error was not printed on `javacLogger.flush(exitCode)`.

It was correctly added to `JavacLogger.msgs` with level `error`, so I'd expect `parseAndDumpSemanticErrors` to fail to parse it and log it instead, not sure why that doesn't appear to be happening.

jvican:
@raboof Any ideas how we can reproduce/test this?

raboof:
Like https://github.com/raboof/zinc/commit/6253ceec9246d73b66bfdf7354275484dd194d8f ?

jvican:
That looks good. Hopefully we fix this soon. Thanks @raboof! :smile:

jvican:
Closing in favor of https://github.com/sbt/zinc/issues/520 which is richer in details. We'll add the provided reproduced case in this ticket to our test suite, thank you @raboof

PR #448, opened 2017-10-30
Ignored javadoc tool error

raboof:
Reported under #415

### DIFF
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavacProcessLogger.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavacProcessLogger.scala
@@ -13,6 +13,8 @@ package javac
 
 import xsbti._
 import java.io.File
+
+import scala.collection.mutable.ListBuffer
 import scala.sys.process.ProcessLogger
 
 /**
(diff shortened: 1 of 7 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/448/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavacProcessLogger.scala
```diff
@@ -20,43 +20,35 @@ import scala.sys.process.ProcessLogger
...
+  def flush(toolname: String, exitCode: Int): Unit = {
     // TODO - NOTES may not be displayed correctly!
-    msgs collect {
-      case (Info, msg) => msg
-    } foreach { msg =>
-      log.info(msg)
+    synchronized {
+      val parser = new JavaErrorParser(cwd)
+
+      parser.parseProblems(err, log).foreach(reporter.log(_))
```
dwijnand: Is this right? The old impl interleaved `\n`'s between error messages.

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavacProcessLogger.scala
```diff
@@ -20,43 +20,35 @@ import scala.sys.process.ProcessLogger
...
+      out.split("\n").filter(_ != "").foreach(log.info(_))
```
dwijnand: Are you purposely splitting info messages by newlines to send them independently to log.info?

Why not, instead of concat-ing info messages and then re-splitting, just collect the messages into a list buffer and then iterate that list into log.info?

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/255/console">is in progress...</a>

dwijnand (commented):
Running the cross-module validator to check if we can get rid of `JavacLogger#flush(Int)`.

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavacProcessLogger.scala
raboof: Hmm, you're right - probably makes sense to add those newlines in `err(s: => String)` (and add some more tests describing how this is actually called).

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavacProcessLogger.scala
raboof: Hmm, for some reason I got the impression that we couldn't assume `out` would be called in neatly framed chunks, but indeed looking at the previous implementation again it appears that we in fact can. In that case a `ListBuffer` indeed makes more sense, I'll update accordingly!

typesafe-tools:
[//]: # (Refs: 1.x-pull/448/head-1.x-1.x-1.x-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/255/">has checked</A> the following projects against Scala 2.12,

  sbt               | <A HREF="https://github.com/sbt/sbt/tree/1.x">1.x                 </A> | sbt/sbt@12c2e9b8e756548b4aeef743e758071acf82c3f7
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/448">pull/448/head       </A> | sbt/zinc@142f7aa12ba11baead0105638e9d285d71efa381
  io                | <A HREF="https://github.com/sbt/io/tree/1.x">1.x                 </A> | sbt/io@e8c9757799bb0b251e5c5aefd96f841428d0eb9a
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/1.x">1.x                 </A> | sbt/librarymanagement@91b56c50db93294003511fd40967c9296b07ef45
  util              | <A HREF="https://github.com/sbt/util/tree/1.x">1.x                 </A> | sbt/util@f4eadfc259f6f25b61eeb519fe29fb44220bec4a

:white_check_mark: The result is: ***SUCCESS***
<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=1.x&zinc=pull/448/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavacProcessLogger.scala
```diff
@@ -20,43 +20,35 @@ import scala.sys.process.ProcessLogger
...
-  private def parseAndDumpSemanticErrors(): Unit = {
-    val input =
-      msgs collect {
-        case (Error, msg) => msg
-      } mkString "\n"
-    val parser = new JavaErrorParser(cwd)
-    parser.parseProblems(input, log).foreach(reporter.log)
-  }
-
-  def flush(exitCode: Int): Unit = {
```
dwijnand: Looks like we can drop this method, but we need to tell MiMa to ignore it.

dwijnand (dismissed):
LGTM, thanks @raboof.

raboof:
(rebased, unfortunately it seems that triggered github to 'dismiss' the review :/)

jvican:
Hey @raboof thanks! Can you squash all the commits? The previous commits to the formatting are not passing in the CI, and it seems that this fix could be all in one commit.

raboof:
Sure, done!

dwijnand (approved):

jvican (approved):
Thanks @raboof :smile:

PR #450, opened 2017-11-01
Forward port sbt/sbt#3701, JDK9/Scala 2.{10,11} overcompilation

retronym:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/450/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -114,8 +114,9 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
             // The dependency comes from a JAR
             for {
               zip <- zipEntry.underlyingSource
-              classFile <- Option(zip.file)
-            } binaryDependency(classFile, binaryClassName)
+              jarFile <- Option(zip.file)
+              if !jarFile.isDirectory // workaround for JDK9 and Scala 2.10/2.11, see https://github.com/sbt/sbt/pull/3701
+            } binaryDependency(jarFile, binaryClassName)
```
retronym, line 119: This is the only change that fixes an observable bug. I didn't figure out why the bug in SBT 0.13 fixed by `-Dscala.ext.dirs` was not necessary here.

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/CompilerArguments.scala
```diff
@@ -13,6 +13,7 @@ import xsbti.ArtifactInfo
 import scala.util
 import java.io.File
 import CompilerArguments.{ absString, BootClasspathOption }
+import sbt.IO
```
jvican: This is causing a compilation error and can be removed.

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/CompilerArguments.scala
```diff
@@ -82,7 +83,13 @@ final class CompilerArguments(
    * @param addLibrary Flag to return the Scala library.
    */
   def createBootClasspath(addLibrary: Boolean): String = {
-    val originalBoot = System.getProperty("sun.boot.class.path", "")
+    def findBoot: String = {
+      import scala.collection.JavaConverters._
+      System.getProperties.asScala.iterator.collectFirst {
+          case (k, v) if k.endsWith(".boot.class.path") => v
```
jvican: Have you done this to avoid making it specific to `scala.boot.class.path`? Is there any other valuable system variable that users `.boot.class.path` as the suffix?

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
jvican: Can you elaborate why Scalac reports the underlying source to be a directory under JDK9? I'm curious.

jvican (changes requested):
LGTM! Just have a few questions and a suggestion.

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
retronym: To integrate the JEP-220 support into Scala 2.10/2.11, I made the `jrt://` virtual file system appear as a `ZipFile`. This was done by subclassing, but the superclass had inadvertently overriden `override val underlyingSource = Some(jarFile)`. I could not override this to return `None` without making a binary compatible change to the method I was overriding (widening to `Option`).

So I return `Some($JAVA_HOME)` instead, but that messed up Zinc.

Rocks and hard places, all around. Filtering here in Zinc seemed the least bad option for now, but we might need to revisit it to make Zinc properly invalidate after the JDK is upgraded.

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/CompilerArguments.scala
retronym: The code is from scalac. I believe the intent is to support non OpenJDK based JVMs that don't use `sun.boot.classpath`.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=188648#c6
http://mail.openjdk.java.net/pipermail/build-dev/2013-November/011103.html

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
jvican: Great to know this, thanks for the explanation.

jvican (approved):

jvican:
This is good to go on my side.

dwijnand (approved):
I'm happy with this. And given @eed3si9n merged the 0.13 PR https://github.com/sbt/sbt/pull/3701 my guess is he's cool with this too.

PR #452, opened 2017-11-09
Fix #433: Make classpath hashing more lightweight

### LINKED ISSUE #433: Make classpath hashing more lightweight

jvican:
And make it parallel!

This patch adds a cache that relies on filesystem metadata to cache
hashes for jars that have the same last modified time across different
compiler iterations. This is important because until now there was a
significant overhead when running `compile` on multi-module builds that
have gigantic classpaths. In this scenario, the previous algorithm
computed hashes for all jars transitively across all these projects.

This patch is conservative; there are several things that are wrong with
the status quo of classpath hashing. The most important one is the fact
that Zinc has been doing `hashCode` on a SHA-1 checksum, which doesn't
make sense. The second one is that we don't need a SHA-1 checksum for
the kind of checks we want to do. https://github.com/sbt/zinc/pull/371
explains why. The third limitation with this check is that file hashes
are implemented internally as `int`s, which is not enough to represent
the richness of the checksum. My previous PR also tackles this problem,
which will be solved in the long term.

Therefore, this pull request only tackles these two things:

* Caching of classpath entry hashes.
* Parallelize this IO-bound task.

Results, on my local machine:

- No parallel hashing of the first 500 jars in my ivy cache: 1330ms.
- Parallel hashing of the first 500 jars in my ivy cache: 770ms.
- Second (cached) parallel hashing of the first 500 jars in my ivy cache: 1ms.

### DIFF
--- a/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
@@ -11,6 +11,7 @@ package inc
 
 import java.io.File
 import java.lang.ref.{ Reference, SoftReference }
+import java.nio.file.Files
 import java.util.Optional
 
 import inc.javac.AnalyzingJavaCompiler
@@ -20,6 +21,7 @@ import xsbti.compile._
 import sbt.io.IO
 import sbt.util.{ InterfaceUtil, Logger }
 import sbt.internal.inc.JavaInterfaceUtil.EnrichOption
+import sbt.internal.inc.caching.ClasspathCache
 import xsbti.compile.ClassFileManager
 
 /** An instance of an analyzing compiler that can run both javac + scalac. */
@@ -181,13 +183,11 @@ object MixedAnalyzingCompiler {
       incrementalCompilerOptions: IncOptions,
       extra: List[(String, String)]
   ): CompileConfiguration = {
-    val classpathHash = classpath map { x =>
-      FileHash.of(x, Stamper.forHash(x).hashCode)
-    }
+    val classpathHash = ClasspathCache.hashClasspath(classpath)
     val compileSetup = MiniSetup.of(
       output,
       MiniOptions.of(
-        classpathHash.toArray,
+        classpathHash,
         options.toArray,
         javacOptions.toArray
       ),
--- a/zinc/src/main/scala/sbt/internal/inc/caching/ClasspathCache.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/caching/ClasspathCache.scala
@@ -1,0 +1,43 @@
+package sbt.internal.inc.caching
+
+import java.io.File
+import java.nio.file.Files
+import java.util.concurrent.ConcurrentHashMap
+import java.nio.file.attribute.{ BasicFileAttributes, FileTime }
+
+import xsbti.compile.FileHash
+import sbt.internal.inc.{ EmptyStamp, Stamper }
+
+object ClasspathCache {
+  // For more safety, store both the time and size
+  private type JarMetadata = (FileTime, Long)
+  private[this] val cacheMetadataJar = new ConcurrentHashMap[File, (JarMetadata, FileHash)]()
+  private[this] final val emptyStampCode = EmptyStamp.hashCode()
+  private def emptyFileHash(file: File) = FileHash.of(file, emptyStampCode)
+  private def genFileHash(file: File, metadata: JarMetadata): FileHash = {
+    val newHash = FileHash.of(file, Stamper.forHash(file).hashCode())
+    cacheMetadataJar.put(file, (metadata, newHash))
+    newHash
+  }
+
+  def hashClasspath(classpath: Seq[File]): Array[FileHash] = {
+    // #433: Cache jars with their metadata to avoid recomputing hashes transitively in other projects
+    def fromCacheOrHash(file: File): FileHash = {
+      if (!file.exists()) emptyFileHash(file)
+      else {
+        // `readAttributes` needs to be guarded by `file.exists()`, otherwise it fails
+        val attrs = Files.readAttributes(file.toPath, classOf[BasicFileAttributes])
+        if (attrs.isDirectory) emptyFileHash(file)
+        else {
+          val currentMetadata = (attrs.lastModifiedTime(), attrs.size())
+          Option(cacheMetadataJar.get(file)) match {
+            case Some((metadata, hashHit)) if metadata == currentMetadata => hashHit
+            case None                                                     => genFileHash(file, currentMetadata)
+          }
+        }
+      }
+    }
+
+    classpath.toParArray.map(fromCacheOrHash).toArray
+  }
+}
--- a/zinc/src/test/scala/sbt/inc/cached/CachedHashingSpec.scala
+++ b/zinc/src/test/scala/sbt/inc/cached/CachedHashingSpec.scala
@@ -1,0 +1,58 @@
+package sbt.inc.cached
+
+import java.nio.file.Paths
+
+import sbt.inc.{ BaseCompilerSpec, SourceFiles }
+import sbt.internal.inc.{ Analysis, CompileOutput, MixedAnalyzingCompiler }
+import sbt.io.IO
+
+class CachedHashingSpec extends BaseCompilerSpec {
+  def timeMs[R](block: => R): Long = {
+    val t0 = System.nanoTime()
+    block // call-by-name
+    val t1 = System.nanoTime()
+    (t1 - t0) / 1000000
+  }
+
+  "zinc" should "cache jar generation" in {
+    IO.withTemporaryDirectory { tempDir =>
+      val classes = Seq(SourceFiles.Good)
+      val sources0 = Map(Paths.get("src") -> classes.map(path => Paths.get(path)))
+      val projectSetup = ProjectSetup(tempDir.toPath(), sources0, Nil)
+      val compiler = projectSetup.createCompiler()
+
+      import compiler.in.{ setup, options, compilers, previousResult }
+      import sbt.internal.inc.JavaInterfaceUtil._
+      import sbt.io.syntax.{ file, fileToRichFile, singleFileFinder }
+
+      val javac = compilers.javaTools.javac
+      val scalac = compilers.scalac
+      val giganticClasspath = file(sys.props("user.home"))./(".ivy2").**("*.jar").get.take(500)
+
+      def genConfig = MixedAnalyzingCompiler.makeConfig(
+        scalac,
+        javac,
+        options.sources,
+        giganticClasspath,
+        CompileOutput(options.classesDirectory),
+        setup.cache,
+        setup.progress.toOption,
+        options.scalacOptions,
+        options.javacOptions,
+        Analysis.empty,
+        previousResult.setup.toOption,
+        setup.perClasspathEntryLookup,
+        setup.reporter,
+        options.order,
+        setup.skip,
+        setup.incrementalCompilerOptions,
+        setup.extra.toList.map(_.toScalaTuple)
+      )
+
+      val hashingTime = timeMs(genConfig)
+      val cachedHashingTime = timeMs(genConfig)
+      assert(cachedHashingTime < (hashingTime * 0.20),
+             s"Cache jar didn't work: $cachedHashingTime is >= than 20% of $hashingTime.")
+    }
+  }
+}


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/452/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

fommil:
can you please validate against the demo project at https://github.com/cakesolutions/sbt-cake/tree/sbt-perf-regression ?

I have no idea how to run a custom zinc, so I am unable to do so. Hence the monkey patch PRs to sbt.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
```diff
@@ -162,6 +165,13 @@ final class MixedAnalyzingCompiler(
  * of cross Java-Scala compilation.
  */
 object MixedAnalyzingCompiler {
+  private[this] val cacheMetadataJar = new ConcurrentHashMap[File, (FileTime, FileHash)]()
```
fommil: this will leak... you need to use something like a Gauva weak cache.

Also I recommend using `Path` instead of `File`, because it has a much smaller memory footprint.

In fact `File` (and `URL`) are the biggest heap hogs in sbt on large projects... mostly because the path prefix is duplicated. I'm talking GBs.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
```diff
@@ -181,9 +191,16 @@ object MixedAnalyzingCompiler {
       incrementalCompilerOptions: IncOptions,
       extra: List[(String, String)]
   ): CompileConfiguration = {
-    val classpathHash = classpath map { x =>
-      FileHash.of(x, Stamper.forHash(x).hashCode)
+    val parallelClasspathHashing = classpath.toParArray.map { file =>
+      val attrs = Files.readAttributes(file.toPath, classOf[BasicFileAttributes])
+      val currentFileTime = attrs.lastModifiedTime()
```
fommil: using size as well as modified is effectively free at his point (it's in attrs) so we could use it to be safe.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
```diff
@@ -181,9 +191,16 @@ object MixedAnalyzingCompiler {
       incrementalCompilerOptions: IncOptions,
       extra: List[(String, String)]
   ): CompileConfiguration = {
-    val classpathHash = classpath map { x =>
-      FileHash.of(x, Stamper.forHash(x).hashCode)
+    val parallelClasspathHashing = classpath.toParArray.map { file =>
```
fommil: `.toParArray`? Is there a better way? The parallel collections rarely speed things up. Certainly hitting the cache in parallel will be slower.

Perhaps parallelisation can be done in a follow up so its impact can be isolated?

jvican: That's not my experience. Parallel collections do speed things up and they're reliable. I can tell you hitting the cache is not slower, I've done measurements locally.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
jvican: Sure.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
jvican: You're right that I should use a path here. I'll use a Synchronized map wrapping a weak cache ref. Ironically that was my first approach but didn't quite like it.

jvican:
@fommil I will not validate it myself. I have already a test that proves this works. If you're interested in checking this fix out, you can publish zinc, change the zinc version in sbt/sbt, publish sbt/sbt locally and use that version in `build.properties`.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
jvican: Even for 10 jars, doing it in parallel is faster than doing it in serial.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
fommil: doesn't sbt have a synchronised weak cache already?

fommil:
I tried that process, but lost several hours, and then gave up and basically did this

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
propensive: It might be worth comparing the performance on an SSD versus a mechanical disk. It depends on a lot of factors (and I've no idea how modern technology mitigates them), but parallelizing IO operations on a mechanical disk might actually slow it down.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
jvican: Not to the best of my knowledge...
  
Finally, I've decided not to use a synchronize weak map here. We could get fancy, but this cache is only for jars and it's going to have tops 1000 entries. So I'll keep it as it is.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
```diff
@@ -162,6 +165,17 @@ final class MixedAnalyzingCompiler(
  * of cross Java-Scala compilation.
  */
 object MixedAnalyzingCompiler {
+  private type JarMetadata = (FileTime, Long)
+  // Using paths instead of files as key because they have more lightweight memory consumption
+  private[this] val cacheMetadataJar = new ConcurrentHashMap[File, (JarMetadata, FileHash)]()
```
leonardehrenfried: Doesn't the code do something different from what the comment says? The key is a `File`.

fommil: and that is why comments SUCK

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
```diff
@@ -181,13 +195,28 @@ object MixedAnalyzingCompiler {
       incrementalCompilerOptions: IncOptions,
       extra: List[(String, String)]
   ): CompileConfiguration = {
-    val classpathHash = classpath map { x =>
-      FileHash.of(x, Stamper.forHash(x).hashCode)
+    // #433: Cache jars with their metadata to avoid recomputing hashes transitively in other projects
+    val parallelClasspathHashing = classpath.toParArray.map { file =>
+      if (!file.exists()) emptyFileHash(file)
```
fommil: I'm pretty sure you can use the file attributes on a non-existent file and then use the response to detect existence... if that is the case it would be best to use it here and avoid a stack frame (which is slow on windows so best to batch / minimise all I/O calls)

jvican: I was just getting IO errors because of this, it was unintuitive to me too.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
jvican: Yup, that's a stale comment, thanks for the catch.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
jvican: In my experience, IO-bound operations yield very good results when they're parallelized. I think there are old textbooks on parallel computing that claim so too, before the pre-SSD era.
  
In any case, I benchmarked it just in case in a LAMP server. Results:
  
* Parallelized hashing: 549ms.
* Serial hashing: 1912ms.
  
I'm testing it on `/dev/vda`, which as you see is a rotational disk:
  
```
platform@scalagesrv2:~/tests/zinc$ lsblk -d -o name,rota
NAME ROTA
fd0     1
sr0     1
vda     1
```

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
jvican: @fommil I got a proof, `/caca/peluda` does not exist:
```scala
scala> val attrs = java.nio.file.Files.readAttributes((new java.io.File("/caca/peluda").toPath), classOf[BasicFileAttributes])
java.nio.file.NoSuchFileException: /caca/peluda
  at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
  at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
  at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
  at sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
  at sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:144)
  at sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
  at java.nio.file.Files.readAttributes(Files.java:1737)
  ... 28 elided
```

fommil: ok cool, that's good to know.

fommil (approved):

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
```diff
@@ -162,6 +165,17 @@ final class MixedAnalyzingCompiler(
  * of cross Java-Scala compilation.
  */
 object MixedAnalyzingCompiler {
+  // For more safety, store both the time and size
+  private type JarMetadata = (FileTime, Long)
+  private[this] val cacheMetadataJar = new ConcurrentHashMap[File, (JarMetadata, FileHash)]()
```
stuhood: Given that https://github.com/sbt/zinc/pull/427 is in, it would seem that all of this code should be located on the companion object of the "default" implementation of hashing. Put it in the top of `MixedAnalyzingCompiler` makes that class into more of a grabbag.

stuhood (approved):
Looks fine from my perspective, although aligning more cleanly with https://github.com/sbt/zinc/pull/427 would be good.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
jvican: By the companion object of the default implementation of hashing, do you mean `FileHash` or `Stamper`? It's not clear to me we want to reuse this code across all the users of `IOHash` or `FileHash`.

fommil:
is this PR against the 1.0.x line or 1.1 line?

I think you might need to backport to 1.0.x if it could be included in an sbt point release.

jvican:
It's in the 1.0.x line.

jvican:
I've just realized that https://github.com/sbt/zinc/pull/427 is in 1.x, and I need to backport it to 1.0.x before I apply my patch. I think it would be a good idea to revisit all this 1.0.x and 1.x thing...

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
stuhood: If there isn't a relevant companion class, then I guess my suggestion would be to define a new one. Not a blocker. Just that logically I would expect this to look something like:
```
val hashingStrategy = if (flag) DefaultStrategy else UserDefinedStrategy(...)
???
```

But looking at it again, ExternalHooks doesn't make it possible to do this. So only the "grabbag" part is really relevant.

fommil:
@jvican yeah that's what I thought. But you don't need to backport 427 for this. You can just implement this as the default and that won't impact the binary API for 1.0.x.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
jvican: I agree it looks like a grabbag, so it's better to move it somewhere else. thanks for the suggestion :smile:.

jvican:
Right, that's better @fommil. This will be shipped in a minor release, so it's better to avoid breaking bincompat in 1.0.x and then adapt @romanowski's approach to my PR.

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
jvican: @stuhood I've implemented your suggestion :wink:. The Zinc API has been historically such a mess that sometimes I forget it can and should be improved :smile:.

stuhood (approved):

godenji:
So I just pulled this PR; published local Zinc snapshot along with Sbt snapshot that uses it.

Not seeing any difference wrt to hashing time, `sbt.io.Hash$.apply` consumes same amount of time as with Sbt 1.0.3/Zinc 1.0.

Here's a [Visualvm screenshot](https://imgur.com/a/0qHV1) (running `compile` several times in succession).

Maybe this PR solves a different problem? `sbt -Dsbt.task.timings=true` shows all of the time is spent in `compileIncremental`; this in a 30kloc app spread across 650 scala source files with 30 subprojects.

jvican:
@godenji Are you sure you've correctly published local Zinc and sbt? I really recommend to double check that, and change the versions as much as possible to avoid problems with cached jars. Make sure you use an sbt version that hasn't been used before.
  
Since I don't have access to your project, it would really help if you can instrument this PR to report data and identify which classpath entries are being cached and which are not, as well as auxiliary useful information.

godenji:
> Are you sure you've correctly published local Zinc and sbt?

@jvican I considered that, but after blowing away `~/.ivy2/local/org.scala-sbt/` the non-change in performance persists. Starting up test project (vanilla giter8 play project) does show sbt pulling in the local artifacts (e.g. "using local for ... zinc-compiler-1.1.0-SNAPSHOT"), and published source jars do reflect the changes in this PR.

I'll dig around some more, bottleneck may be elsewhere than what this PR resolves.

jvican:
@godenji You also need to blow away the corresponding folder in `~/.sbt/boot`. I encourage you to use a completely different version, something like `37.1.1-jar-change`.

godenji:
> You also need to blow away the corresponding folder in ~/.sbt/boot

@jvican that was it, noop `compile` now takes 0ms, yes ;-)

Since this is going into Zinc 1.0.x and Sbt 1.0.4 is slated for this coming week, is there any chance that this magnificent fix can make its way into Sbt? That would be most welcome, incremental builds are horrendous ATM.

jvican:
I've benchmarked this in a personal project too and I confirm it's fixed. Thanks for volunteering to try this out @godenji.

jvican:
@eed3si9n Can you review this please :smile:?

eed3si9n (approved):
LGTM
I like that the change is contained to implementation side.

Duhemm (approved):
LGTM

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/284/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: 1.x-pull/452/head-1.x-1.x-1.x-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/284/">has checked</A> the following projects against Scala 2.12,

  sbt               | <A HREF="https://github.com/sbt/sbt/tree/1.x">1.x                 </A> | sbt/sbt@c5175179273f669edcc5ffa70f6f667f91ef0a3d
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/452">pull/452/head       </A> | sbt/zinc@a90a0e98f5be965452261388050debe5a7396dac
  io                | <A HREF="https://github.com/sbt/io/tree/1.x">1.x                 </A> | sbt/io@e8c9757799bb0b251e5c5aefd96f841428d0eb9a
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/1.x">1.x                 </A> | sbt/librarymanagement@163e8bff825750f222d8cde11574cbaf23d671b6
  util              | <A HREF="https://github.com/sbt/util/tree/1.x">1.x                 </A> | sbt/util@f4eadfc259f6f25b61eeb519fe29fb44220bec4a

:white_check_mark: The result is: ***SUCCESS***
<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=1.x&zinc=pull/452/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

PR #471, opened 2017-12-18
Add 1.0.5 to mimaPreviousArtifacts, & backfill

dwijnand:
Remove the file exclude files before (a) compiler-bridge's exclude
file is redundant since we no longer mima-check it, and (b) it interacts
badly with multiple versions.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,9 +10,11 @@ lazy val compilerBridgeTestScalaVersions = List(scala212, scala211, scala210)
 
 def mimaSettings: Seq[Setting[_]] = Seq(
   mimaPreviousArtifacts := Set(
-    organization.value % moduleName.value % "1.0.0"
+    "1.0.0", "1.0.1", "1.0.2", "1.0.3", "1.0.4", "1.0.5",
+  ) map (version =>
+    organization.value %% moduleName.value % version
       cross (if (crossPaths.value) CrossVersion.binary else CrossVersion.disabled)
-  )
+  ),
 )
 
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/471/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

eed3si9n (dismissed):

jvican (changes requested):
I think the exclusions should be handled as before. You can remove the compiler bridge exclusions and add `1.0.5` without specifying the exclusions in the sbt build file (which IMO brings us more disadvantages than advantages -- I don't want to compile my build every time I exclude a class).

dwijnand:
I tried to keep the exclusions handled as before, but the problem is when it checks for version 1.0.5 it needs all the exclusions already present in the 1.0.0 exclusions to be repeated in the 1.0.5 exclusion. So you need to duplicate the exclusion files. And copy it a bunch more times if you want to verify multiple versions like I'm doing. It seems to be a downside of the file-based approach rather than the programmatic approach.

What I propose is that we merge this as is, and discuss in the MiMa repo how to avoid the copy-and-paste issue. WDYT?

jvican:
SGTM, Dale.

jvican (dismissed):

jvican:
Can you rebase and we merge?

PR #473, opened 2017-12-20
Fix sbt/sbt#3736. Recurse into classloader of the class not the root

ravwojdyla:
sbt/sbt#3736


lightbend-cla-validator:
Hi @ravwojdyla,

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/473/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

jvican:
Thanks for the contribution @ravwojdyla :heart:. Could you check if this fixes sbt/sbt#3736? I would like to see some tests here to be 100% sure that this does the right thing.

jvican:
I’ve gone into the details of this patch and the fix looks good to me, by the way. I’m surprised it hasn’t been caught before.

dwijnand (dismissed):
Given `ClasspathFilter`'s purpose is to verify "that all classes loaded through `parent` either come from `root` or `classpath`" (from its scaladoc) I agree changing from traversing the root classloader's parents to the class's classloader's parents looks right.

And I'm equally surprised this wasn't caught before.

Ideally we'd have a test that accurately reproduces the downstream issue, but I'm happy to merge this without.

ravwojdyla:
Thanks for reviews @dwijnand, @jvican. Sorry, I have made a mistake by referencing wrong sbt issue, this solves **only** sbt/sbt#3736. I updated the commit message and title of this PR. Regarding sbt/sbt#3515, this change does not solve it.

jvican:
To me, this is good to go IMO. Further investigation can take care of sbt/sbt#3736 :smile:.

jvican (dismissed):

ravwojdyla:
@jvican thanks. This patch also fixes sbt/sbt/issues/3647, sbt/sbt/issues/3608 and sbt/sbt/issues/3733

ravwojdyla:
FYI I have signed the CLA.

eed3si9n:
@ravwojdyla Hi, thanks for your contribution.

I'm trying to verify this by adding a scripted test locally, and I am unable to confirm using https://github.com/raboof/sbt-run-classloading/blob/master/src/main/scala/Main.scala. Does it run for you?

I am still getting

```
[error] Caused by: java.lang.ClassNotFoundException: scala.Int
[error] 	at sbt.internal.inc.classpath.ClasspathFilter.loadClass(ClassLoaders.scala:74)
[error] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
[error] 	at java.lang.Class.forName0(Native Method)
[error] 	at java.lang.Class.forName(Class.java:348)
[error] 	at Main$.delayedEndpoint$Main$1(A.scala:7)
```

eed3si9n:
Never mind. I think there was some internal discrepancy in scripted due to relative path vs absolute path. Let me send a PR on top of your PR.

eed3si9n:
https://github.com/ravwojdyla/zinc/pull/1

ravwojdyla:
Thanks @eed3si9n . Merged.

eed3si9n (approved):

raboof:
Thanks @ravwojdyla! 🎉

elliotvilhelm:
Having this issue on `sbt: stable 1.2.8` running quickstart https://developer.lightbend.com/guides/akka-quickstart-scala/

System: Mac OSX Mojave 10.14.3 (18D109)

raboof:
@ElliotVilhelm hmm, I can't reproduce that (though I'm running Linux, so that might be a difference)

How exactly did you test this? Did you use the 'Create a project for me' button and then run 'sbt run'? What does the error look like, exactly, for you?

PR #475, opened 2018-01-05
Add test case for problem with constructs like A.type#B#C

romanowski:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/475/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

zinc/src/sbt-test/source-dependencies/nested-type-params/test
```diff
@@ -0,0 +1,9 @@
+> compile
+
+# change type alias
+$ copy-file changes/Definitions.scala Definitions.scala
+$ copy-file changes/Usage.scala Usage.scala
+
+
+# Both A.scala and B.scala should be recompiled, producing a compile error
```
dwijnand: Is this a copy-and-paste error? There's no A.scala and B.scala, and is the "producing a compile error" comment correct?

romanowski: Yep, this is copy paste error.

jvican:
I don't understand, the test suite is passing. If this is an issue, it should fail.

zinc/src/sbt-test/source-dependencies/nested-type-params/UsingProvier.scala
```diff
@@ -0,0 +1,3 @@
+object UsingProvier {
```
jvican: The name of the source file has a typo.

romanowski:
I've pushed correct changes to wrong repo and without checking I've opened this PR. No it represents the bug I found. Sorry for mess :)

romanowski:
@jvican @dwijnand Is this failure expected:
```
sbt.internal.inc.InvalidScalaInstance: Scala instance doesn't exist or is invalid: 
[error]     version 2.10.7, library jar: /drone/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.10.7.jar, compiler jar: /drone/.ivy2/cache/org.scala-lang/scala-compiler/jars/scala-compiler-2.10.7.jar, other jars: /drone/.ivy2/cache/org.scala-lang/jline/jars/jline-2.10.7.jar, /drone/.ivy2/cache/org.scala-lang/scala-reflect/jars/scala-reflect-2.10.7.jar, /drone/.ivy2/cache/org.fusesource.jansi/jansi/jars/jansi-1.4.jar
```

dwijnand:
@romanowski No, I've opened https://github.com/sbt/zinc/issues/479.

dwijnand:
CI issue fixed. Restarting Drone.

zinc/src/sbt-test/source-dependencies/nested-type-params/test
```diff
@@ -0,0 +1,8 @@
+> compile
+
+$ copy-file changes/Providers.scala Providers.scala
+$ copy-file changes/Bar.scala Bar.scala
+
+-> compile # This should compile
```
romanowski, line 6: @jvican this compile should succeed (clean build 2 lines below is green). I prefer this style of pointing out issues since if I leave it as a last command and change `test` to `pending` *any* failure  will make this test green. There is an issue for this and if someone fix the problem this test will fail and comment in this line will tell that person that this is expected state.

Previously tests was green since there was 3 files and changing 2 of them was over 50% and problematic compilation was in fact clean-build.

jvican: The issue with this approach is that we cannot know just by grepping through the directory which tests are failing and which are not, though.

romanowski: TBH I can chabge is to pending - its up to you.

dwijnand: I can see both sides. Let's establish a keyword to use in a comment, which we can grep for.

dwijnand: Perhaps `PENDING`

jvican: Your solution is good. Let me propose something similar that I think it's better: let's change `pending` to mean "there's one or more statements in this test case that fail but should succeed". How do we mark the statements that are "PENDING"? Well, I have two suggestions:

1. `# PENDING` at the end of the statement.
2. `!>` and `!->` instead of `>` and `->`.
  
I don't mind which one you eventually pick. The thing is that with this change:

* We know by just looking for file names which test cases are to be done.
* We know within the file which statements should be the ones that pass/fail.
  
This requires a change in the scripted implementation, but shouldn't take long. Could you do it @romanowski ? What does everyone think here?

romanowski: Maybe we can keep it even simpler:
```
PENDING > compile
PENDING $ copy-file
PENDING -> clean
# etc
````

 I can take a look tomorrow if we agree on notation.

jvican: I don't mind syntax, your take @dwijnand?

dwijnand: I like Krzysztof's `PENDING >`. Easier to intuit, easier to grep for.

jvican: Looking forward to this change and merging the pending test.

jvican:
The reason why this fails is because our API representation doesn't seem to correctly detect API changes in type aliases. Zinc recompiles the changes but doesn't detect an API change, thus it does not follow up the invalidation algorithm downstream.
  
In the branch where I'm fixing this issue, I've added a test that confirms that the dependency extraction algorithm works correctly by extracting the correct member references (`Foo` depends on `Provider` and `A`). Note that in my branch I've replaced the current `Provider` definition by the shorter `trait Provider { type Operations = A }`.

zinc/src/sbt-test/source-dependencies/nested-type-params/test
jvican: Making this change is too much of a hassle. Let's merge this and I'll follow up with a fix in another PR.

jvican (approved):

PR #484, opened 2018-01-17
Remove "Modified names" assertion

### LINKED ISSUE #292: Modified names for (class) is empty
Pants `1.3.0.dev16` is using zinc `1.0.0-X7`, and users have reported the following exception:

```
java.lang.AssertionError: assertion failed: Modified names for com.foo.bar.Baz is empty
                          at scala.Predef$.assert(Predef.scala:170)
                          at sbt.internal.inc.NamesChange.<init>(Changes.scala:31)
                          at sbt.internal.inc.IncrementalNameHashing.sameAPI(IncrementalNameHashing.scala:35)
                          at sbt.internal.inc.IncrementalCommon.sameClass(IncrementalCommon.scala:158)
                          at sbt.internal.inc.IncrementalCommon$$anonfun$7.apply(IncrementalCommon.scala:140)
                          at sbt.internal.inc.IncrementalCommon$$anonfun$7.apply(IncrementalCommon.scala:139)
                          at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:241)
                          at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:241)
                          at scala.collection.immutable.HashSet$HashSet1.foreach(HashSet.scala:316)
                          at scala.collection.immutable.HashSet$HashTrieSet.foreach(HashSet.scala:972)
                          at scala.collection.immutable.HashSet$HashTrieSet.foreach(HashSet.scala:972)
                          at scala.collection.TraversableLike$class.flatMap(TraversableLike.scala:241)
                          at scala.collection.AbstractTraversable.flatMap(Traversable.scala:104)
                          at sbt.internal.inc.IncrementalCommon.changedIncremental(IncrementalCommon.scala:139)
                          at sbt.internal.inc.IncrementalCommon.cycle(IncrementalCommon.scala:47)
                          at sbt.internal.inc.Incremental$$anonfun$1.apply(Incremental.scala:78)
                          at sbt.internal.inc.Incremental$$anonfun$1.apply(Incremental.scala:77)
                          at sbt.internal.inc.Incremental$.manageClassfiles(Incremental.scala:122)
                          at sbt.internal.inc.Incremental$.compile(Incremental.scala:77)
                          at sbt.internal.inc.IncrementalCompile$.apply(Compile.scala:60)
                          at sbt.internal.inc.IncrementalCompilerImpl.compileInternal(IncrementalCompilerImpl.scala:132)
                          at sbt.internal.inc.IncrementalCompilerImpl.incrementalCompile(IncrementalCompilerImpl.scala:87)
                          at org.pantsbuild.zinc.Compiler.compile(Compiler.scala:204)
                          at org.pantsbuild.zinc.Main$.run(Main.scala:97)
                          at org.pantsbuild.zinc.Main$.main(Main.scala:16)
                          at org.pantsbuild.zinc.Main.main(Main.scala)
                          at sun.reflect.GeneratedMethodAccessor1.invoke(Unknown Source)
                          at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
                          at java.lang.reflect.Method.invoke(Method.java:498)
                          at com.martiansoftware.nailgun.NGSession.run(NGSession.java:280)
```

...while recompiling after switching branches. The affected class was changed in the branch, it's possible others were as well.

jvican:
This assertion dates from the time where the whole Zinc 1.x was under
development. As I understand it, it was put in place to make sure that
the names invalidation was correct.

However, the assertion itself is not valid. It violates some legit
scenarios, where there are changes in APIs but there are no changes
in the names.

One of these examples is the following one, provided by @francisdb,
where the original code is:

```
trait T2
trait T1 extends T2 { def foo: String }
class C1 extends T1 { def foo = "test" }
```

And moving `foo` from `T1` to `T2` causes the assertion to fail. In this
concrete scenario, we can see how the API has changed (the hashes are
not the same than the previous scenario) but the names are identical.

There are more examples, some of them I have occassionally run into.

Therefore, this commit should fix #292 once and for all, and end the
pain of those who have kindly complained about it in the ticket. :)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?zinc=pull/484/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

eed3si9n (dismissed):
LGTM

eed3si9n:
I retargeted it to 1.1.x.

dwijnand:
Given the number of people impacted by #292 would it be possible to add that test (and/or any others tests) to ensure this use case don't regress?

jvican:
> Given the number of people impacted by #292 would it be possible to add that test (and/or any others tests) to ensure this use case don't regress?
  
Yes, good idea, I'll add that in a new commit.

dwijnand:
Thanks @jvican. If you could rebase it on 1.1.x too, so we don't merge any of the 1.x changes in.

gabro:
We just hit on this 5 minutes ago. Our reproduction case is a bit different:

```scala
trait A {
  def foo(x: String = "default"): Unit
}
trait B extends A {
  def foo(x: String): Unit
}
```

if you move the default from `A` to `B` the assertion fails:

```scala
trait A {
  def foo(x: String): Unit
}
trait B extends A {
  def foo(x: String  = "default"): Unit
}
```

I'm glad this is about to be solved, thanks @jvican!

jvican:
Will be updating this tomorrow.

kubukoz:
I just encountered this for the first time, it's great to see that it's moving so well! Thanks @jvican :)

dwijnand (approved):

PR #490, opened 2018-02-16
Fix copy/pasto in definition of IncOptions

retronym:
The wrong default value was being used, but the bug was benign because
that was also just `false`


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/490/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

dwijnand (approved):

Issue #495, opened 2018-02-21
Running unit tests fails in fresh clone


jvican:
Culprit:
  
```
[info] java.lang.NoClassDefFoundError: scala/Product$class
[info]     at xsbti.TestCallback$TestUsedName.<init>(TestCallback.scala:11)
[info]     at xsbti.TestCallback.usedName(TestCallback.scala:60)
```
  
This has happened to me and tried to fix it in my PRs improving the build, and now has been reported by @retronym. I have also experienced this issue today.

PR #496, opened 2018-02-21
Set Scala's compiler interface version to 2.12

jvican:
Fixes https://github.com/sbt/zinc/issues/495.

Trivial fix was:

```
diff --git a/build.sbt b/build.sbt
index bb93aab99..56e518d5f 100644
--- a/build.sbt
+++ b/build.sbt
@@ -310,8 +310,8 @@ lazy val compilerInterface = (project in internalPath / "compiler-interface")
    // Use the smallest Scala version in the compilerBridgeScalaVersions
    // Technically the scalaVersion shouldn't have any effect since scala library is not included,
    // but given that Scala 2.10 compiler cannot parse Java 8 source, it's probably good to keep this.
-    crossScalaVersions := Seq(scala210),
-    scalaVersion := scala210,
+    crossScalaVersions := Seq(scala212),
+    scalaVersion := scala212,
    relaxNon212,
    libraryDependencies ++= Seq(scalaLibrary.value % Test),
    exportJars := true,
```

but that would publish the bridge three times when releasing and cause
an error.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -307,11 +307,8 @@ lazy val compilerInterface = (project in internalPath / "compiler-interface")
     minimalSettings,
     // javaOnlySettings,
     name := "Compiler Interface",
-    // Use the smallest Scala version in the compilerBridgeScalaVersions
-    // Technically the scalaVersion shouldn't have any effect since scala library is not included,
-    // but given that Scala 2.10 compiler cannot parse Java 8 source, it's probably good to keep this.
-    crossScalaVersions := Seq(scala210),
-    scalaVersion := scala210,
+    scalaVersion := scala212,
+    crossScalaVersions := Seq(scala212),
     relaxNon212,
     libraryDependencies ++= Seq(scalaLibrary.value % Test),
     exportJars := true,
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/496/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican:
Getting sbt-whitesource error:
  
```
[error] Error communicating with service: Illegal arguments: Bad org token
[error] sbtwhitesource.WhiteSourceException: Error communicating with service: Illegal arguments: Bad org token
```

jvican:
Could you have a look at this @eed3si9n @dwijnand?

build.sbt
```diff
@@ -342,6 +339,17 @@ lazy val compilerInterface = (project in internalPath / "compiler-interface")
...
 
+/* Create a duplicated compiler-interface project that uses Scala 2.10 to parse Java files.
+ * Scala 2.10's parser uses Java 6 semantics, so this way we ensure that the interface can
+ * be compiled with Java 6 too. `compiler-interface` checks compilation for Java 8. */
+val compilerInterfaceJava6Compat = compilerInterface
+  .withId("compilerInterfaceJava6Compat")
+  .settings(
+    scalaVersion := scala210,
+    crossScalaVersions := Seq(scala210),
+    target := (target in compilerInterface).value / "java6-parser-compat"
```
eed3si9n: Should we add `skip in publish := true`?

eed3si9n (changes requested):
Overall LGTM

build.sbt
jvican: Indeed!

jvican (commented):
Addressed :+1:

eed3si9n (approved):

PR #497, opened 2018-02-21
Traverse original trees of constant-folded trees

jvican:
This commit adds support for `OriginalTreeAttachment`, which is an
attachment that was introduced in 2.12.4 and that will have the right
semantics in 2.12.5. See https://github.com/scala/scala/pull/6343 for
more information.

In this PR, we configure the build to cross-compile to different compat
utilities on 2.12.4 onwards because of the introduction of this
attachment.

Then, we make the used names and dependency traverser visit the original
trees that are stored in the attachments. These original trees make the
test `constants` pass.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/497/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican:
This PR is blocked until the aforementioned fix in scala/scala is merged. CI is expected to fail.

dwijnand:
Nice for this to be addressed. However I'm not sure if we should duplicate `ConsoleInterface`.. :-/

jvican:
@dwijnand Yeah, thought the same. I'm not satisfied with the solution but this seems the most reasonable way to address it I've come up with. Do you have any suggestion on how to handle this differently?

dwijnand:
not off the top of my head - I'd have to play with it

jvican:
@dwijnand @eed3si9n Thoughts on the current approach to handle the cross-compilation to 2.12.5 in the build differently? I'd like to unblock this PR.

eed3si9n:
If you knew which Scala version you're currently running on, can you do if-else?

See https://github.com/sbt/zinc/blob/1.x/internal/zinc-ivy-integration/src/main/scala/sbt/internal/inc/ZincComponentCompiler.scala#L146-L148

jvican:
@eed3si9n It's not possible because the compiler bridge implementation needs to link with an attachment that is solely introduced in 2.12.4: https://github.com/sbt/zinc/pull/497/files#diff-441a9c67d0459a1d33ccac13a83e0253R22.
  
I don't think there's a better way of solving this problem :thinking:.

eed3si9n:
The "bridge" is basically a bag of source code that's compiled on the fly and used, so one thing we could do is abstract the difference into a smallest possible Scala source, similar to Compat, and swap out just one file.

An alternative might be that we implement sbt task that does the copying as managed source, so the file is kept up to date if we ever change it.

jvican:
> The "bridge" is basically a bag of source code that's compiled on the fly and used, so one thing we could do is abstract the difference into a smallest possible Scala source, similar to Compat, and swap out just one file.
  
Even if we do this, we still need to create one different `Compat` or implementation for `2.12.4+` and `2.11-2.12.{0, 4}`, so we'd be in the same situation this PR is in: `Compat` and `ConsoleInterface` will need to be duplicated.
  
I've tried playing with implicits to get it working, but nothing I tried worked. The difficulty comes from the fact that we want one source file (from scala_2.12.4, which would only contain complementary source files to the normal `Compat`) to define an implicit that has more precedence than one in the existing scala_2.11-2.12 sources (the stub).
  
The last thing I tried was using reflection, but it's not possible to do it and stay performant (I recommend reading `Attachements.scala` in scalac). The issue is that we need to call `original` on the object via reflection, and we do that in a hot path (traversals in both `Dependency` and `API`).
  
I could try `MethodHandle`s, but I don't wanna go there given the complexity they would introduce to the codebase and the fact that there would be still a performance hit. I think it's just better to give the green light to this code, and to live with the fact that Scala 2.12 has a different source for 2.12.4+.

eed3si9n (dismissed):
I am ok with unblocking this PR, but it would be good to come back to duplicating the ConsoleInterface.scala at later time.

dwijnand:
why can't we have 2 directories for pre-2.12.4 and post-2.12.4 Compat, and another directory with 2.11-2.12 ConsoleInterface?

jvican:
> why can't we have 2 directories for pre-2.12.4 and post-2.12.4 Compat, and another directory with 2.11-2.12 ConsoleInterface?
  
We can, this is the alternative fix I proposed in Gitter. I'll implement that.

Issue #459, opened 2017-11-26
Java 9 WARNING: Illegal reflective access by com.google.protobuf.UnsafeUtil


eed3si9n:
## steps

Run on Java 9.

## problem

```
[info] Compiling 1 Scala source to /root/work/hello/project/target/scala-2.12/sbt-1.0/classes ...
[info] Non-compiled module 'compiler-bridge_2.12' for Scala 2.12.4. Compiling...
[info]   Compilation completed in 9.719s.
[info] Done compiling.
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.protobuf.UnsafeUtil (file:/root/.sbt/boot/scala-2.12.4/org.scala-sbt/sbt/1.0.4-LOCAL1/protobuf-java-3.3.1.jar) to field java.nio.Buffer.address
WARNING: Please consider reporting this to the maintainers of com.google.protobuf.UnsafeUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

## expectation

No warnings

## notes

See https://github.com/google/protobuf/issues/3781

mkurz:
I am getting something different:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by sbt.internal.inc.javac.DiagnosticsReporter$PositionImpl$ (file:/home/mkurz/.sbt/boot/scala-2.12.4/org.scala-sbt/sbt/1.1.2/zinc-compile-core_2.12-1.1.3.jar) to field com.sun.tools.javac.api.ClientCodeWrapper$DiagnosticSourceUnwrapper.d
WARNING: Please consider reporting this to the maintainers of sbt.internal.inc.javac.DiagnosticsReporter$PositionImpl$
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

I already use sbt `1.1.2` (released 3 hours ago)

eed3si9n:
https://github.com/google/protobuf/pull/4557

eximius313:
On JDK10:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.protobuf.UnsafeUtil (file:/C:/Users/X/.sbt/boot/scala-2.12.6/org.scala-sbt/sbt/1.1.5/protobuf-java-3.3.1.jar) to field java.nio.Buffer.address
[...]

PR #519, opened 2018-03-27
Cleanup DiagnosticsReporter.PositionImpl & upgrade to JDK7+

dwijnand:
I was having a look at https://github.com/sbt/zinc/issues/459.

Looks like we depend (via reflection) on private JDK APIs that haven't since been made public.. so the fix isn't as straightforward as one would hope.

### DIFF
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/DiagnosticsReporter.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/DiagnosticsReporter.scala
@@ -99,9 +99,9 @@ object DiagnosticsReporter {
 
     /**
      * Extracts PositionImpl from a Java Diagnostic.
-     * The previous implementation of PositionImpl held on to the Diagostic object as a wrapper,
+     * The previous implementation of PositionImpl held on to the Diagnostic object as a wrapper,
      * and calculated the lineContent on the fly.
-     * This caused race condition on the Diagnostic object, resulting to NullPointerException.
+     * This caused a race condition on the Diagnostic object, resulting in a NullPointerException.
      * See https://github.com/sbt/sbt/issues/3623
      */
     def apply(d: Diagnostic[_ <: JavaFileObject]): PositionImpl = {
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/519/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/DiagnosticsReporter.scala
```diff
@@ -117,67 +117,64 @@ object DiagnosticsReporter {
...
 
-      def fixSource[T <: JavaFileObject](source: T): Option[String] = {
-        try Option(source).map(_.toUri.normalize).map(new File(_)).map(_.getAbsolutePath)
-        catch {
-          case _: IllegalArgumentException =>
-            // Oracle JDK6 has a super dumb notion of what a URI is.  In fact, it's not even a legimitate URL, but a dump
-            // of the filename in a "I hope this works to toString it" kind of way.  This appears to work in practice
-            // but we may need to re-evaluate.
-            Option(source).map(_.toUri.toString)
+      def sourceUri: Option[String] =
```
eed3si9n: This should be named as `sourcePath`, not `sourceUri`, since the happy path is the absolute path, not URI.

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/DiagnosticsReporter.scala
```diff
@@ -117,67 +117,64 @@ object DiagnosticsReporter {
...
+        source map { obj =>
+          val uri = obj.toUri
+          try new File(uri.normalize).getAbsolutePath
```
eed3si9n: In our current arsenal, a better bet here is `IO.toFile(uri).getAbsolutePath` instead of the try-catch.

See https://github.com/sbt/io/blob/10dd004d3c0baad7e88016751208def76989316a/io/src/main/scala/sbt/io/IO.scala#L130-L131 and http://eed3si9n.com/encoding-file-path-as-URI-reference

eed3si9n (changes requested):
A few comments.

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/DiagnosticsReporter.scala
dwijnand: huh bizarre. it's called `sourceUri` in `PositionImpl` and copied without modification to `sourcePath`. how confusing. I'll rename it here to pass review.

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/DiagnosticsReporter.scala
dwijnand: I'll push a commit - let me know if I correctly did what you meant.

jvican (approved):

PR #521, opened 2018-03-29
Add an index that associates a name with a position

bakenezumi:
Ref https://github.com/sbt/sbt/issues/4053

I took over @eed3si9n san's branch and tried to improve the function to decide where to jump to source based on full name in `sbt lsp`.

Please pardon points that I could not squash successfully..

For specification see [ExtractUsedNamesSpecification](https://github.com/sbt/zinc/compare/1.1.x...bakenezumi:wip/names-position#diff-96c1efafc3c7d29c1c81a22b6b4b6da9) and [IncrementalCompilerSpec](https://github.com/sbt/zinc/compare/1.1.x...bakenezumi:wip/names-position#diff-0018ee6451c053d34b8a3f30ab281203).

When collecting a simple name from the tree by [ExtractUsedNames](https://github.com/sbt/zinc/compare/1.1.x...bakenezumi:wip/names-position#diff-539728e41226af62e8d439d137d0d9b8), I modified to collect the full name and its location at the same time. Also gathered similarly defined names. Put back it to AnalysisCallback, and I added it to the object to [persistent with protobuf](https://github.com/sbt/zinc/compare/1.1.x...bakenezumi:wip/names-position#diff-b53006d46fc50a5990f00e600805b85b).

Since persistence of this index can cause performance degradation, 
~~I added an option `storeNamePositions` to [IncOptions](https://github.com/sbt/zinc/compare/1.1.x...bakenezumi:wip/names-position#diff-1e57b4563e3fd14379cb2cd1be2e2348).~~
I am deciding whether to create it now using `IncOptions.storeApis`. Should  use `IncOptions.extra`?

Please check the following for incorporation of `sbt lsp`.

https://github.com/sbt/sbt/compare/1.1.x...bakenezumi:wip/names-position


lightbend-cla-validator:
Hi @bakenezumi,

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.1.x&zinc=pull/521/head&io=1.1.x&librarymanagement=1.1.x&util=1.1.x&website=1.1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
Thanks for this @bakenezumi!

> Since persistence of this index can cause performance degradation

Do you know how much degradation this is, in terms of time and space (size of Analysis file)?

eed3si9n:
I cleaned up the Git history and chery-picked it to 5 commits - https://github.com/sbt/zinc/compare/1.x...eed3si9n:wip/names-position-pk?expand=1
Do you mind if I pushed that to your branch?

internal/zinc-core/src/main/scala/sbt/internal/inc/NamePosition.scala
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright 2011 - 2017, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * This software is released under the terms written in LICENSE.
+ */
+
+package sbt.internal.inc
+
+import xsbti.compile.analysis
+
+case class NamePosition private (line: Int, column: Int, name: String, fullName: String)
```
eed3si9n: We should try to cooperate with semanticdb3 schema (https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md) and its protobuf file.

```proto
message SymbolOccurrence {
  Range range = 1;
  string symbol = 2;
  Role role = 3;
}
```

For example, it would be good to storage range position for the declared names.

eed3si9n: Also we should avoid the use of case class, and use Contraband instead.

eed3si9n (changes requested):

bakenezumi:
Thank you for the cleaned up and review @eed3si9n!

> Do you know how much degradation this is, in terms of time and space (size of Analysis file)?

I compared sbt1.1.2 with those that incorporated zinc that I modified.
The source used to measure compile time was `sbt/mainPrj`.
It compiling 79 Scala sources.
`clean`, `scalafmt`, then `compile` repeatedly, see `compile` time below.

sbt(zinc) ver|1st time|2nd time|3rd time|size of `inc_compile.bin`
-------------|--------:|---------:|--------:|---:
1.1.2(1.1.2)|20 sec | 20 sec| 19 sec | 818,766 byte|
1.1.3-SNAPSHOT(this zinc)| 21 sec | 21 sec | 21 sec| 3,201,902 byte|

After correction I will remeasure.

internal/zinc-core/src/main/scala/sbt/internal/inc/NamePosition.scala
xeno-by: @eed3si9n @bakenezumi It's great that you guys are considering SemanticDB! 

If Zinc emits SemanticDB-compatible indexes, then these indexes will be immediately usable by tools like [scalafix](https://github.com/scalacenter/scalafix), [metadoc](https://github.com/scalameta/metadoc) and others, which sounds very very exciting.

To be compatible with SemanticDB, you will need to do four things, as far as I can see:
  1) Add a dependency on `"org.scalameta" %% "semanticdb3" % "3.7.0"` that contains bindings to the SemanticDB protobuf schema.
  2) Create an instance of `scala.meta.internal.semanticdb3.TextDocuments` that contains a single `scala.meta.internal.semanticdb3.TextDocument` with `occurrences` set to a collection of SymbolOccurrence objects as highlighted by @eed3si9n.
     Note: All fields in `TextDocument` are optional, so you don't have to populate any other sections of the SemanticDB schema. Just `occurrences` will be enough.
  3) Convert fully-qualified names like `scala.Int` to SemanticDB-compatible symbols like `scala.Int#` as described [in the specification](https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#scala-symbol).
  4) Save this instance to the Zinc index using `TextDocuments.writeTo`.

Please let me know if you have questions. I'll be happy to assist you guys in any way.

eed3si9n: Another thing we should cooperate is the naming convention. Semanticdb appends `#` for type names and appends `.` after a term name.

For example, object `O1` might have full name of `_root_.p1.O1.`, and `Int` becomes `_root_.scala.Int#`.

xeno-by: @eed3si9n Actually, in February/March we changed the naming convention a little bit to drop `_root_.` and to simplify method signatures. Check out [the spec](https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#scala-symbol) for a naming algorithm and some examples.

jvican (changes requested):
I'm really reluctant to accept this patch. If someone wants an index to implement an Scala LSP server, I suggest they use Semanticdb via the Scalameta compiler plugin.
  
Zinc is not an index generator. Zinc is an incremental compiler. In a moment where we are moving towards decreasing overhead (as the overhead added by serializing the API to the analysis file), this only adds more overhead and makes it difficult to optimize for.
  
On top of that, getting an index right is *really* difficult and I don't want to add more complexity to this project. If you guys want to have a drop-in solution to get positions for sbt LSP, you can add a compiler plugin by default to every sbt module.
  
Zinc is not the solution to this problem, I'm afraid. Let's not use a bazooka to point to a small target.

bakenezumi:
@jvican thanks for opinion.
Zinc gathers dependencies using the presentation compiler, but I think that it is not bad to leave the information acquired when traversing the AST for LSP and other tools.
However, I agree that zinc should not be too complicated.

jvican:
@bakenezumi I highly respect your work, but I'm afraid I don't see how we can get down this rabbit hole while still keeping complexity and performance to a sane level.
  
https://github.com/scalameta/scalameta/ is an excellent compiler plugin that has gone through many work iterations and nails indexing for Scala programs. If you want indices, it has a drop-in installation. I would highly suggest you direct your efforts towards working with the Scalameta team and enable a cross-tool language server for Scala. There's some WIP in https://github.com/scalameta/metals which works with any editor that supports LSP (though atom and vscode are the most tested ones).

eed3si9n:
@jvican 

> On top of that, getting an index right is really difficult and I don't want to add more complexity to this project. If you guys want to have a drop-in solution to get positions for sbt LSP, you can add a compiler plugin by default to every sbt module.

Getting the correct type information for all position is difficult (not everything has names), but collecting all the full names is fairly simple additions to the existing tree walking. At least no more complicated than the stuff that's already going on to collect the names for incremental compilation.

The perf overhead is a legitimate concern, but @bakenezumi's initial data seem to show only 1s added overhead. I'd be happy to put this behind a flag, so people can opt-in or -out.

I think this is a low hanging fruit that can be useful to any tooling that uses Zinc. Zinc already has mapping from sources to classes, and from classes to API info. Position to type information is just a small missing link, which this approximates (with all due credit to SemanticDB of course). I'd be happy to get more feedback from @xeno-by et al, if accuracy is a concern.

Similarly, I'd love to get rangepos enabled by default for Zinc.

jvican:
> Getting the correct type information for all position is difficult (not everything has names), but collecting all the full names is fairly simple additions to the existing tree walking. At least no more complicated than the stuff that's already going on to collect the names for incremental compilation.

> I think this is a low hanging fruit that can be useful to any tooling that uses Zinc. Zinc already has mapping from sources to classes, and from classes to API info
  
These comments arise from the essential complexity of compiling incrementally, the main purpose of Zinc. It has nothing to do with indexing, which is a difficult problem that I suggest it's solved elsewhere.
  
> The perf overhead is a legitimate concern, but @bakenezumi's initial data seem to show only 1s added overhead. I'd be happy to put this behind a flag, so people can opt-in or -out.
  
1 or 2s out of 21 is a lot, even for a small project like Zinc. It translates to 5-10% of the whole compile. The best case (1s out of 21s) is the equivalent to the performance difference between 2.12.3 and 2.12.4 (!), and 2.12.4 was the release that brought the fastest compile speed relative to previous releeases.
  
I think it's fair to say that the performance overhead will be way higher for bigger projects. Just see that the size of the analysis file has been multiplied by four.
  
> I'd be happy to put this behind a flag, so people can opt-in or -out.
  
Adding this code to Zinc adds accidental complexity, a complexity we don't want to deal with. We will start with this, and then people will start to add more and more features, and it will be chaos in the blink of an eye, hurting both performance and maintenance, and forcing us to fix any bugs in this area.
  
The way to go is to continue the promising work of metals. Scalameta has been heavily tested in many codebases, and duplicating work in such a fundamental problem is not only concerning for us as maintainers, but also a waste of resources. Motivated people like @bakenezumi can make a huge impact contributing to the ongoing tooling efforts to bring a solid Scala LSP server to reality, and sbt can use that to offer a nice LSP server.

olafurpg:
I think it's exciting to see exploration in this area! We've been working on
mapping positions to symbols in SemanticDB for over a year now with promising
results. Here are some thoughts:

* -Yrangepos helps, but it has problems with crashing the compiler when
  positions overlap. I've made decently successful experiments running the
  scanner to get `(start, end)` pairs and use those to map offset positions to
  range positions. This approach avoids the -Yrangepos requirement at the cost
  of a single scanner pass, which should be cheap. We don't use this approach in
  SemanticDB but I think it's worth exploring.
* A good testing infrastructure is essential. Positions from the compiler are
  often wrong and require minor adjustments. A good testing infrastructure is
  critical to fix issues and catch regressions.
* There's a long tail of constructs that need special handling, anonymous
  classes, package objects, for comprehensions, val destructuring, constant
  folding, `private[within]`, `List()` is transformed into `Nil`, to name a few.
  Be sure to test these.
* There is compilation overhead to persisting this data. In XCode
  Index-While-Building the overhead is ~3-5% (see talk
  https://www.youtube.com/watch?v=jGJhnIT-D2M&) but in SemanticDB it's ~20-30%
  depending on the settings (we persist more than symbol occurrences and haven't
  spent much time optimizing). I would love learn how to improve in this space!

I think it's good to decide ahead of time what level of accuracy you are aiming
for. How many false positives/negatives are acceptable? If the data is only used
for goto definition it may be fine to have noisy data but if the data is used
for refactorings like rename you might have lower tolerance for false positives
or off-by-ones.

jvican:
> in SemanticDB it's ~20-30% depending on the settings (we persist more than symbol occurrences and haven't spent much time optimizing).

A note: from 2.12.5 on, the overhead is about 16% if dynamic plugin classloading is disabled. I presume there is still lots of work to do to make it more performant on the compiler plugin side.

bakenezumi:
Thanks for advice and opinion.
I am pleased that I can participate in the discussion.
I empathized with SBT which has a lightweight LSP as a standard service and created this pull request.

I wanted to confirm performance first, so I revised according to review from @eed3si9n and advice from @xeno-by.

The change point is as follows.
 - Change the permanent format is made to conform to semanticdb3 schemasemanticdb3 schema.
 - Ceased the use of case class, and use Contraband instead.
 - Follow to the semanticdb3 naming convention.
 - Add `IncOptions.storeSymbolIndex` that determines whether incremental compiler stores symbol index alongside analysis.

For naming conventions I received advice from @olafurpg.
Based on [scala.meta.internal.semanticdb.scalac.SymbolOps](https://github.com/scalameta/scalameta/blob/00ed4fd65996d4a57b37454d3a88702f0d3f8b42/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/SymbolOps.scala#L14-L47), I created [NoMetaSemanticdbOps](https://github.com/sbt/zinc/pull/521/files#diff-6daff2c33a13a84cbf2cf958d63b0f2d).

However, the following is omitted currently.
 - Local symbol.
 - Distinguish overloaded methods that a same name but a different package.

The test code for this is [here](https://github.com/sbt/zinc/pull/521/files#diff-09bdb9e89b375f363201552072de92fe).

Performance is below.
The target is the same as before.

 sbt(zinc) ver|`storeSymbolIndex`|1st time|2nd time|3rd time|size of `inc_compile.bin`
-------------|:-----:|--------:|---------:|--------:|---:
1.1.3-SNAPSHOT(this zinc)|true|21 sec | 20 sec| 20 sec | 3,777,989 byte|
1.1.3-SNAPSHOT(this zinc)|false| 19 sec | 19 sec | 19 sec| 819,640 byte|

To opt-out `storeSymbolIndex` is as follows.

```
set incOptions := incOptions.value.withStoreSymbolIndex(false)
```

The current index can be used for the following purposes.
 - Jump to definition
 - Find usage

Trying to do more than this will increase complexity.


sbt changes as below.
https://github.com/sbt/sbt/compare/1.x...bakenezumi:wip/symbol-index

internal/compiler-interface/src/main/contraband-java/xsbti/compile/IncOptions.java
```diff
@@ -52,6 +52,9 @@ public static boolean defaultUseCustomizedFileManager() {
     public static boolean defaultStoreApis() {
         return true;
     }
+    public static boolean defaultStoreSymbolIndex() {
+        return true;
```
olafurpg, line 56: Is `true` a good default? This behavior is a regression for all users that don't use sbt language server. I think it makes more sense that users opt-into this new functionality.

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
```diff
@@ -177,6 +240,30 @@ class ExtractUsedNames[GlobalType <: CallbackGlobal](val global: GlobalType)
...
+          val ignoreReference = {
+            val tpe = symbol.tpe
+            role == Role.REFERENCE && (definitions.isByNameParamType(tpe) || definitions
+              .isRepeatedParamType(tpe))
+          }
+          if (!registeredReference && !ignoreReference && !symbol.hasPackageFlag && !ignoredSymbol(
+                symbol) && !isEmptyName(symbol.name)) {
+            val semanticName = symbol.toSemanticName
+            // Local names are not collected.
+            if (!semanticName.isEmpty) {
```
olafurpg, line 257: Does this mean goto definition won't work for local variables?

internal/compiler-bridge-test/src/test/scala/xsbt/ExtractUsedNamesSpecification.scala
```diff
@@ -38,9 +38,24 @@ class ExtractUsedNamesSpecification extends UnitSpec {
                   |   }
                   |}""".stripMargin
     val compilerForTesting = new ScalaCompilerForUnitTesting
-    val usedNames = compilerForTesting.extractUsedNamesFromSrc(srcA, srcB)
+    val (usedNames, referencedSymbols, definedSymbols) =
+      compilerForTesting.extractUsedNamesFromSrc(srcA, srcB)
     val expectedNames = standardNames ++ Set("a", "c", "A", "B", "C", "D", "b", "X", "BB")
     assert(usedNames("b.X") === expectedNames)
+    val expectedSymbols = Set(
+      (2, 16, "a.A#"),
```
olafurpg, line 46: I would recommend to come up with a more ergonomic testing solution. In scalameta we use "expect tests" which make it nice to review and easy to create new tests https://blog.janestreet.com/testing-with-expectations/

Here is an example expect file in scalameta repo 
- expect file https://github.com/scalameta/scalameta/blob/master/tests/jvm/src/test/resources/lowlevel.expect#L160-L184
- actual test suite https://github.com/scalameta/scalameta/blob/master/tests/jvm/src/test/scala/scala/meta/tests/semanticdb/ExpectSuite.scala#L44-L47

olafurpg (commented):
Nice job @bakenezumi !

I think you should reconsider what default value makes most sense for the majority of zinc users.

internal/compiler-bridge-test/src/test/scala/xsbt/ExtractUsedNamesNamingConventionSpecification.scala
```diff
@@ -0,0 +1,306 @@
...
+                |    type E = xe.type forSome { val xe: AnyRef }
+                |  }
+                |}""".stripMargin
+    val compilerForTesting = new ScalaCompilerForUnitTesting
+    val (_, referencedSymbols, definedSymbols) =
+      compilerForTesting.extractUsedNamesFromSrc(src)
+    val expectedReferencedSymbols = Set(
+      (0, 25, "scala.Int#"),
+      (1, 10, "scala.Int#"),
+      (1, 16, "scala.Predef#$qmark$qmark$qmark()."),
```
olafurpg, line 26: Symbolic identifiers should be in decoded form instead of encoded as here. This particularity is not documented in the spec, I will update it https://github.com/scalameta/scalameta/issues/1510

internal/compiler-interface/src/main/contraband-java/xsbti/compile/IncOptions.java
bakenezumi: Thank you for review @olafurpg !
Given the main purpose of zinc,  I think your opinion is right.

internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala
bakenezumi: No, I think the language server that uses this index should analysis the local variables separately and return it to the client.
Since it is enough to analyze only one file, it should be able to execute at low cost.
Considering the enlargement of index size, and performance, I omitted it.

internal/compiler-bridge-test/src/test/scala/xsbt/ExtractUsedNamesSpecification.scala
bakenezumi: Thanks for the wonderful advice.
I learned a lot.
I will try it soon.

internal/compiler-bridge-test/src/test/scala/xsbt/ExtractUsedNamesNamingConventionSpecification.scala
bakenezumi: OK. Thank you.

eed3si9n:
Closing this for now.

PR #532, opened 2018-04-22
Use read/write mappers for Relations in protobuf format

pauldraper:
@romanowski (author of mappers) @jvican (author of protobuf format)

`RelationsTextFormat` already uses these mappers.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/532/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

dwijnand:
I can think up reasons, but I might be wrong: what's the motivation of this change?

dwijnand:
btw, looks like this affected 2 tests

PR #533, opened 2018-04-23
Disable recompile all fraction for all scripted tests

jvican:
The default for recompile all fraction is 0.5, which means that if the changes
done in an scripted test affect more than 50% of all the tests, a test could
pass because all files are recompiled even though it should fail.

To protect us from this potential issue, this patch tweaks the value of the
recompile all fraction for scripted so that full recompilation is disabled for
our scripted tests.

### DIFF
--- a/internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
+++ b/internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
@@ -473,8 +473,14 @@ case class ProjectStructure(
       val scalacOptions =
         Option(map.get("scalac.options")).map(_.toString.split(" +")).getOrElse(Array.empty)
 
-      (IncOptionsUtil.fromStringMap(map, scriptedLog), scalacOptions)
-    } else (IncOptions.of(), Array.empty)
+      val incOptions = {
+        val opts = IncOptionsUtil.fromStringMap(map, scriptedLog)
+        if (opts.recompileAllFraction() != IncOptions.defaultRecompileAllFraction()) opts
+        else opts.withRecompileAllFraction(1.0)
+      }
+
+      (incOptions, scalacOptions)
+    } else (IncOptions.of().withRecompileAllFraction(1.0), Array.empty)
   }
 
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/533/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

dwijnand (commented):
sgtm! but `transitive-class` is failing in CI..

jvican:
Yeah, looking into it. We should at some point also increase the factor for normal applications. I have a feeling the current number is too conservative, and  the limit on 50% seems to be counter-productive to the higher accuracy of the class dependency algorithm.

jvican:
The error in `transitive-class` is not serious :tada:.

jvican:
Ready for review. @dwijnand Mind having another look?

dwijnand (approved):
LGTM. Could you (briefly) explain why transitive-class had to change, please?

jvican:
Sure thing. `checkRecompilations` only checks that the last triggered compile compiled the given class names. In our case, the last triggered compile doesn't have `A` because `A` was the modified source and the first one to be compiled. After `A` was compiled, it triggered the last incremental compilation of `B`, `C` and `D`, which are the transitive dependencies, and this is the compile that `checkRecompilations` is checking against.

dwijnand:
Thanks.

So it sounds to me like we're being over specific in the test: we care to assert that all `A`, `B`, `C` `D`, and `Hello` were compiled, without necessarily caring if it was in the last triggered compile or not, right?

jvican:
> So it sounds to me like we're being over specific in the test: we care to assert that all A, B, C D, and Hello were compiled, without necessarily caring if it was in the last triggered compile or not, right?
  
Yeah. I don't know why these are the semantics. If I find some time soon, maybe I can fix this one.

PR #534, opened 2018-04-23
Add Bloop scripted integration

jvican:
I'd like to add upstream this tiny bloop scripted integration that extends the
zinc scripted infrastructure to be able to run it with bloop.

These are the benefits of running it with bloop:

1. Faster turnaround time and startup.
2. No need to publish jars before running.
3. Small maintenance cost.

This script still needs to be tweaked so that we can test different compiler
bridges at the same time. I think this is possible and I'd like to make this
change in a future PR.

Since @Duhemm and I hack on Zinc, we'd like to merge this upstream because
we believe it gives value to future contributors. We take ownership of it and
are happy to maintain it as bloop evolves.

### DIFF
--- a/bin/run-ci.sh
+++ b/bin/run-ci.sh
@@ -12,6 +12,7 @@ sbt -Dfile.encoding=UTF-8 \
   whitesourceCheckPolicies \
   compilerInterfaceJava6Compat/compile \
   zincRoot/test:compile \
+  bloopScripted/compile \
   crossTestBridges \
   "publishBridgesAndSet $SCALA_VERSION" \
   zincRoot/test \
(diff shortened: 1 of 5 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/534/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

Duhemm (approved):

PR #542, opened 2018-05-11
Detect changes in private trait definitions

jvican:
This pull request builds upon @Duhemm's work on
https://github.com/sbt/sbt/pull/2441, but changes slightly the way it's
solved. It instead adds a specific api change that only invalidates
changes via inheritance (both global and local) and doesn't invalidate
member references, which gives us more correctness and precision.

As explained in https://github.com/sbt/sbt/issues/2490, the previous fix
violated some of the invariants of the incremental algorithm. As such,
we introduce the notion of an extra hash which is independent of the
normal one. This extra hash, in the case of traits, is computed from the
private definitions found, and otherwise its value is the empty hash (-1).

As a result, we can tell apart the cases where there are changes
in the normal public API hash and the extra one. I've called it extra
instead of `trait private hash` or something similar because it can
accommodate other fixes in the future and we cannot afford adding a new
hash per `AnalyzedClass` every time we need to have `HashAPI` be special
cased for a particular encoding. Right now, when the extra hash is
different and the classes involved in the hash computations are traits
we treat it as a breakage in the trait encoding and recompile
appropriately. If the classes involved are not hash computations, we
fail because no more cases have been added yet.

Fixes:

1. `source-dependencies/trait-private-val`
1. `source-dependencies/trait-private-var`
1. `source-dependencies/trait-private-object`
1. `source-dependencies/trait-super`

And it also adds another two tests to complement
`source-dependencies/trait-member-modified` in checking that member
references to private trait definitions do not trigger unnecessary
compilations and that local inheritance does trigger an extra compile.

Note that the code in `SameAPI` is rotten and will need to go at some
point in the future. It's outdated and the only one-liners used in the
incremental compiler should be inlined, and the rest of the code
ditched. Right now, it tries to emulate the semantics of `HashAPI` but
it hasn't been up-to-date ever since we released Zinc 1.0. I leave this
task for a future PR.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/542/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican:
P.S. I wanted to tackle this problem before I merge both `API` and `Dependency` and implement hashing on the fly so that I can have a better understanding of how I will handle this extra API hash then.

smarter:
Shouldn't you only include private val/var, not private def? The latter do not need fields in the subclasses. (It seems that including private def is what allow `trait-super` to pass but that's a special case: super-accessors are created as private defs but become public later in the compiler pipeline)

jvican:
From the current logic: ‘case d: Def => d.modifiers.isSuperAccessor’

jvican:
Integration tests are passing. Lacking the junit tests, @eed3si9n @smarter would you mind reviewing the current design? I'd like to have early suggestions so that I can batch the implementation of junit tests and the potential improvements to the fix.

jvican:
Could you have a look at this too? @dwijnand Would love some feedback.

dwijnand:
I'm not going to have time right now, I'm personally taking care of hitting some Lightbend-internal deadlines atm. Sorry 😕

internal/compiler-interface/src/main/contraband/other.json
```diff
@@ -10,6 +10,7 @@
         { "name": "name",        "type": "String"          },
         { "name": "api",         "type": "lazy Companions" },
         { "name": "apiHash",     "type": "int"             },
+        { "name": "apiHashExtra","type": "int"             },
```
eed3si9n: `apiHash` means hash of API info (sum total of public signatures).
`apiHashExtra` doesn't make sense. I would suggest something like `extraHash`, and also let's add `"docs"` to explain subtle differences of those like in https://github.com/sbt/zinc/blob/1.x/internal/compiler-interface/src/main/contraband/incremental.json

eed3si9n: Also I assume we want to keep binary compatibility of compiler interface, so it might be good to use `"since"` feature so we can generate bincompat overloads.

eed3si9n (changes requested):
overall looks good
minor nitpick on the naming

jvican:
Feedback addressed and unit tests added. @eed3si9n Can you review this? I'd love to merge this some time soon so that I can continue working on this area 😸

jvican:
After struggling a little bit with the CI, now it's green.

jvican:
@eed3si9n Can you have a look at this again? This is ready to be merged.

eed3si9n (approved):

PR #545, opened 2018-05-27
Avoid null for analysis callback

som-snytt:
Prefer a dummy value, with `enabled = false`.

Fixes https://github.com/sbt/sbt/issues/4057

This is WIP for me to get up to speed on the sbt builds. There is another issue to enable showing scalac options, so the broader goal is to enable showing info from the compiler, including phases, option help.

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/CallbackGlobal.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/CallbackGlobal.scala
@@ -139,18 +139,18 @@ sealed class ZincCompiler(settings: Settings, dreporter: DelegatingReporter, out
     getOutputClass(fqn).map(f => (f, true)).orElse(findOnClassPath(fqn).map(f => (f, false)))
   }
 
-  private[this] var callback0: AnalysisCallback = null
+  private[this] var callback0: AnalysisCallback = DummyAnalysisCallback
 
   /** Returns the active analysis callback, set by [[set]] and cleared by [[clear]]. */
   def callback: AnalysisCallback = callback0
 
   final def set(callback: AnalysisCallback, dreporter: DelegatingReporter): Unit = {
-    this.callback0 = callback
+    if (callback == null) this.callback0 = DummyAnalysisCallback else this.callback0 = callback
     reporter = dreporter
   }
 
   final def clear(): Unit = {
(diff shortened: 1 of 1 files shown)


jvican:
What is the motivation of this change?

dwijnand (approved):
LGTM

som-snytt:
@dwijnand I was trying something easy to learn about the sbt build ecosystem, but didn't have time to make progress. There may also be a better solution to init to ensure `-Xshow-phases` reports correctly, for instance.

jvican (changes requested):
I really don't get the motivation of this change. Can you further elaborate?

som-snytt:
@jvican One way to avoid NPE is not to set values to null.

Closing because I can't really elaborate.

jvican:
I really don’t mind merging this! So consider reopening @som_snytt I wanted to know because it looked like you got into trouble with this and couldn’t imagine how.

eed3si9n:
@jvican This fixes sbt/sbt#4057, which I now transferred over as sbt/zinc#621.
By passing `"-Xshow-phases"` the OP reports `NullPointerException`.

```
sbt:sample-name> compile
[info] Compiling 1 Scala source to /home/enolan/junk/npeTestcase/target/scala-2.12/classes ...
[error] ## Exception when compiling 1 sources to /home/enolan/junk/npeTestcase/target/scala-2.12/classes
[error] null
[error] xsbt.ZincCompiler.phaseDescriptors$lzycompute(CallbackGlobal.scala:121)
[error] xsbt.ZincCompiler.phaseDescriptors(CallbackGlobal.scala:119)
[error] scala.tools.nsc.Global$Run.<init>(Global.scala:1180)
[error] scala.tools.nsc.Global.phaseNames$lzycompute(Global.scala:698)
```

In general we should probably avoid using `null` since it inevitably leaks out like this.

eed3si9n:
Making `DummyAnalysisCallback` I don't think would do the right thing either. I added comment in #621 on what I think is happening.

jvican:
> In general we should probably avoid using null since it inevitably leaks out like this.
  
Agreed, using a dummy is better than the old strategy of using `null` here.

som-snytt:
"crash test dummy."

PR #548, opened 2018-06-04
Add option to Ignore scalac options change

lukaszwawrzyk:
This PR adds `externalCompileSetupEquiv` field to IncOptions. Ths is a function that tells if two `MiniSetup` s are equivalent. 
This allows to do an override of this as in IncHandler.scala:499 to e.g. ignore the change of scalac options for recompilation decision. Without that all the files would be recompiled, with the change, only those that actually changed.

I also resolved the problem with hardcoded implicits in `MiniSetupUtil.scala`. That was caused by a missing import.

The PR seems to also solve https://github.com/sbt/zinc/issues/406


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/548/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/780/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: 1.x-pull/548/head-1.x-1.x-1.x-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/780/">has checked</A> the following projects against Scala 2.12,

  sbt               | <A HREF="https://github.com/sbt/sbt/tree/1.x">1.x                 </A> | sbt/sbt@2848770f851febe34a1205de1408dcbd8090e663
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/548">pull/548/head       </A> | sbt/zinc@adf63dcf379b6e34b9c29d092d5241a8d1c6de35
  io                | <A HREF="https://github.com/sbt/io/tree/1.x">1.x                 </A> | sbt/io@0e4ee5a2ae958c26cef2ed5e61431708c09959b1
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/1.x">1.x                 </A> | sbt/librarymanagement@6e4ad6ebd4042d3b9f1631b4a7a20d27aaa87db8
  util              | <A HREF="https://github.com/sbt/util/tree/1.x">1.x                 </A> | sbt/util@b412c9c92991e27f48ad90153e74a10198246551

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=1.x&zinc=pull/548/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

romanowski:
This option is need for compiler plugins development. If your plugin accepts parameter every change causing everything to be recompiled. In many cases all you need it increase e.g. logging verbosity.

Moreover when debugging compiler (as a part of compiler plugin development) it is really important to have option to compile single file (regardless how much files should be recompiled). The only scenario that cannot be implemented with current hooks is compiler option change - and we aim to fix that in this PR.

romanowski:
@jvican, @dwijnand @eed3si9n is PR verification known to be flaky? Failures in this [build](https://platform-ci.scala-lang.org/sbt/zinc/711) does not seem to be related to this PR.

dwijnand:
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGBUS (0x7) at pc=0x00007f605e9e2a32, pid=34, tid=0x00007f5cc0dad700
#
# JRE version: Java(TM) SE Runtime Environment (8.0_102-b14) (build 1.8.0_102-b14)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.102-b14 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libzip.so+0x11a32]  newEntry+0x62
#
# Core dump written. Default location: /drone/src/github.com/sbt/zinc/core or core.34
#
# An error report file with more information is saved as:
# /drone/src/github.com/sbt/zinc/hs_err_pid34.log
Compiled method (nm)  615154  139     n 0       java.util.zip.ZipFile::getEntry (native)
 total in heap  [0x00007f6050379290,0x00007f6050379600] = 880
 relocation     [0x00007f60503793b8,0x00007f6050379400] = 72
 main code      [0x00007f6050379400,0x00007f6050379600] = 512
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
/usr/bin/sbt: line 34:    34 Aborted                 (core dumped) /usr/lib/bin/sbt "$@" -Dsbt.boot.properties=/sbt.boot
```

Wow

jvican:
I would propose another approach to this problem: instead of outsourcing the function that computes the equivalence between two mini setups, let's create a wishlist of flags that should not be detected to invalidate sources/products.
  
We can have a default whitelist with all the flags you deem appropriate, plus a configurable user-defined whitelist that can come from a build tool.

romanowski:
>We can have a default whitelist with all the flags you deem appropriate, plus a configurable user-defined whitelist that can come from a build tool.

This is not as powerful as  proposed approach and can be implemented with it (it may be good idea to include such implementation in this PR). The situation I've got in mind (mentioned in original issue as well #408) is not recompiling everything when order of options change. It happens sometimes especially when incremental compilation artifacts (a'la Hoarder) are reused between IDE and build tool.

jvican:
> This is not as powerful as proposed approach and can be implemented with it (it may be good idea to include such implementation in this PR). The situation I've got in mind (mentioned in original issue as well #408) is not recompiling everything when order of options change. It happens sometimes especially when incremental compilation artifacts (a'la Hoarder) are reused between IDE and build tool.
  
If this is the case, why don't we just re-implement the equivalence function to ignore order of compilation flags? We can have this + the whitelist. I'm aware it's not as powerful as the solution proposed here, but I think it's more elegant and helps all the clients of Zinc reuse all of this information (instead of requiring every client to repeat the same task over and over again).

dragos:
We're using a similar technique in our plugin, and we needed to test for flags that start with a certain prefix, not just equality (think about adding or removing `-Ylog:parser` or other phase names). Just something to consider if you go for the whitelist approach.

lukaszwawrzyk:
So how about keeping things as they are in the current PR but adding some means to implement more common use cases conveniently. For example to provide method that takes a function that compares two arrays of scalac opts, or takes a whitelist.
To show one example of how it could be used:
```
incOptions := incOptions.value.withExternalCompileSetupEquiv(ignoreScalacOptionsChange("-Xprint:typer", "-Xfatal-warnings")))
```
If that is feasible, what would be a good place to put such helper functions?

jvican:
> So how about keeping things as they are in the current PR but adding some means to implement more common use cases conveniently. For example to provide method that takes a function that compares two arrays of scalac opts, or takes a whitelist.
  
What other cases do you need aside from detecting changed order in compile options and filtering out some flags from the equivalence to merit the inclusion of this hook?
  
The way I see it, we solve all the problems by manually ignoring order in the default equivalence function and introducing the concept of whitelist. The whitelist can support regex and we can provide a default for all our clients.

dragos:
> The way I see it, we solve all the problems by manually ignoring order in the default equivalence function and introducing the concept of whitelist.

It's a can of worms. Parameters are not atomic, think about `-classpath foo:bar` and the like. You will need to know *exactly* which ones take a second argument and keep them together. It will probably break once someone adds a new argument in Scala that takes a second "part" separated by space.

jvican:
The whitelist can be conservative. That’s fine, it’s something it will need to be done anyway if there’s a hook for it.

lukaszwawrzyk:
I am a bit confused on what is the conclusion. Do we want to go with that hook equiv function or only with the whitelist?
As I already had it almost done, now I replaced hook with the whitelist/regex mechanism. 
There is one non obvious thing I implemented. The taken assumption is that all options start with `-` and option can have 0 or 1 parameters that does not start with `-`. Base on that, if option with a parameter is passed, I combine those with a single space for the analysis. This means that one can e.g. specify to ignore `-cores .*` and change `-cores 1` to `-cores 2` without full recompilation. See the tests I added for more details. The order is also ignored. I guess this covers everything that was needed.

internal/zinc-core/src/main/scala/sbt/internal/inc/MiniSetupUtil.scala
```diff
@@ -127,4 +126,26 @@ object MiniSetupUtil {
       def equiv(a: CompileOrder, b: CompileOrder) = a == b
     }
   }
+
+  def equivScalacOptions(ignoredRegexes: Array[String]): Equiv[Array[String]] = {
+    def groupWithParams(opts: Array[String]): Set[String] = {
+      def isParam(s: String) = !s.startsWith("-")
+      def recur(opts: List[String], res: Set[String]): Set[String] = opts match {
+        case opt :: param :: rest if isParam(param) => recur(rest, res + (opt + " " + param))
```
romanowski: Maybe string interpolation? It will be more readable (especially that we are using `+` to add elements to Set

internal/zinc-core/src/main/scala/sbt/internal/inc/MiniSetupUtil.scala
```diff
@@ -127,4 +126,26 @@ object MiniSetupUtil {
...
+      recur(opts.toList, Set.empty)
+    }
+    val ignoreRegex = ignoredRegexes.mkString("|").r
+    def dropIgnored(opts: Set[String]): Set[String] =
+      opts.filterNot(ignoreRegex.pattern.matcher(_).matches)
+
+    new Equiv[Array[String]] {
+      override def equiv(opts1: Array[String], opts2: Array[String]): Boolean = {
+        dropIgnored(groupWithParams(opts1)) == dropIgnored(groupWithParams(opts2))
+
```
romanowski: Empty linie

internal/zinc-core/src/test/scala/sbt/internal/inc/MiniSetupUtilSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  it should "detect change for options with parameters" in {
+    val equiv = equivScalacOptions(ignoredRegexes = Array("-opt a.*"))
+
+    val before = Array("-opt", "abc")
+    val after = Array("-opt", "def")
+
+    equiv.equiv(before, after) shouldBe false
+  }
+
```
romanowski, line 51: We are missing test case for option ordering

zinc/src/sbt-test/source-dependencies/scalac-options/incOptions.properties
```diff
@@ -0,0 +1 @@
+scalac.options = -Xfatal-warnings
```
romanowski: `+ignoredScalacOptions = -Xfatal-warnings` IMHO should be also added there
I would also add another options and change order - to have better coverage

romanowski (changes requested):
Looks good I've added few minor style comments and asked to improve tests a bit

jvican (approved):
Excellent contribution, this is what I meant by the whitelist. I think this will be useful to downstream users and it's in a shape in which it can be immediately used. Thank you @lukaszwawrzyk.
  
/cc @dwijnand @eed3si9n for another review + merge.

dwijnand (approved):
LGTM. Thanks @lukaszwawrzyk.

Issue #125, opened 2016-05-15
Need some replacement for custom ClassfileManager


sjrd:
In sbt 0.13, `IncOptions` has a field `newClassfileManager` of type `ClassfileManager`, which is configurable. In 1.0.0-M4, we only have a very limited `classfileManagerType`.

Scala.js uses the highly configurable `ClassfileManager` here: https://github.com/scala-js/scala-js/blob/v0.6.9/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L110-L137
This is necessary so `.sjsir` files are deleted when their "companion" `.class` files are deleted. Which of course is necessary for incremental compilation to work at all with Scala.js.

We absolutely need a way to implement the same functionality, i.e., this is a **blocker** for Scala.js. Whether by reintroducing `newClassfileManager` or another way to do it.

stuhood:
Correspondingly, when pants uses zinc to build with scala.js, we have the same concern.

stuhood:
@sjrd : One alternative here might be to lean further in on the virtual-file-system approach that @peiyuwang began with #181 . In particular, using a similar vfs implementation for scalac (assuming `scala.js` uses the scalac filesystem APIs) would capture these naturally.

jvican:
@sjrd It seems we can already do this, but we need to dive deeper into the new Zinc 1.0 API. AFAIK, Grzgorz introduced a new interface called `ExternalHooks` that allows you to inject your own `LookUp` and `ClassFileManager` instances for further customisation.
  
This is how I would replicate that functionality with the new API:
```scala
import xsbti.compile.{IncOptions, ExternalHooks, ClassFileManager}

/** Patches the IncOptions so that .sjsir files are pruned as needed.
 *
 *  This complicated logic patches the ClassfileManager factory of the given
 *  IncOptions with one that is aware of .sjsir files emitted by the Scala.js
 *  compiler. This makes sure that, when a .class file must be deleted, the
 *  corresponding .sjsir file are also deleted.
 */
def scalaJSPatchIncOptions(incOptions: IncOptions): IncOptions = {
  // We can be more fine-grained and check which ClassFileManagerType we expect
  val inheritedNewClassfileManager =
    sbt.internal.inc.ClassfileManager.getClassfileManager(incOptions)
  val newClassfileManager = () => new ClassFileManager {
    private[this] val inherited = inheritedNewClassfileManager()

[...]

PR #562, opened 2018-07-31
Fix logic for adding external class file manager

allanrenucci:
If the current external class file manager was empty, the new one was
ignored. Now if the current external is empty, the new one becomes the
current one, otherwise the two are mixed (as specified bu the
documentation in `ExternalHooks.java`).

### DIFF
--- a/internal/compiler-interface/src/main/java/xsbti/compile/DefaultExternalHooks.java
+++ b/internal/compiler-interface/src/main/java/xsbti/compile/DefaultExternalHooks.java
@@ -30,17 +30,15 @@ public class DefaultExternalHooks implements ExternalHooks {
 
     @Override
     public ExternalHooks withExternalClassFileManager(ClassFileManager externalClassFileManager) {
-        Optional<ClassFileManager> currentManager = this.getExternalClassFileManager();
-        Optional<ClassFileManager> mixedManager = currentManager;
-        if (currentManager.isPresent()) {
-            Optional<ClassFileManager> external = Optional.of(externalClassFileManager);
-            mixedManager = Optional.of(WrappedClassFileManager.of(currentManager.get(), external));
-        }
-        return new DefaultExternalHooks(this.getExternalLookup(), mixedManager);
+        Optional<ClassFileManager> external = Optional.of(externalClassFileManager);
+        Optional<ClassFileManager> mixedManager = classFileManager.isPresent()
+            ? Optional.of(WrappedClassFileManager.of(classFileManager.get(), external))
+            : external;
+        return new DefaultExternalHooks(lookup, mixedManager);
     }
 
     @Override
     public ExternalHooks withExternalLookup(ExternalHooks.Lookup externalLookup) {
-        return new DefaultExternalHooks(Optional.of(externalLookup), this.getExternalClassFileManager());
+        return new DefaultExternalHooks(Optional.of(externalLookup), classFileManager);
     }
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.2.x&zinc=pull/562/head&io=1.2.x&librarymanagement=1.2.x&util=1.2.x&website=1.2.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

allanrenucci:
Probably related to scala-js/scala-js#3419

eed3si9n:
Thanks for the contribution @allanrenucci.
Do you think you can add a scripted test under https://github.com/sbt/zinc/tree/1.x/zinc/src/sbt-test?

allanrenucci:
How can I modify the `incOptions`? Zinc scripted tests ignore `build.sbt` right?

I would need to do something like:
```scala
incOptions := {
  val myCustomFileManager = new ClassFileManager {
    def delete(classes: Array[File]): Unit = ...
    def generated(classes: Array[File]): Unit = ...
    def complete(success: Boolean): Unit = ...
  }
  incOptions.value.externalHooks.withExternalClassFileManager(myCustomFileManager)
}
```

eed3si9n:
You're right. So you might need to make it something like https://github.com/sbt/zinc/blob/1.x/zinc/src/test/scala/sbt/inc/ClasspathHashingHookSpec.scala

allanrenucci:
Test added

eed3si9n (approved):
LGTM

jvican:
The only thing that worries me about this PR is that it a breaking change that may affect current uses of these APIs. People right now are doing this aggregation manually (see how Scala.js uses it https://github.com/sbt/zinc/issues/125 via class inheritance). If somebody is not careful, depending on the definition of the class file manager, this can cause problems.
  
The change itself looks good to me, but I would ask that we double check first with @stuhood and other users of these APIs before we merge. The odds of changing behaviors mainstream is slim, but still existent.

allanrenucci:
If people are doing the aggregation manually, then it shouldn't break their code. If people are using `withExternalClassFileManager` (e.g. [Scala.js](https://github.com/scala-js/scala-js/blob/b3f6e7347f4dee06b8f1adc2ab506f8e0758cbdc/sbt-plugin/src/main/scala-sbt-1.0/org/scalajs/sbtplugin/SBTCompat.scala#L55-L56)), then they have a bug

jvican (approved):
> If people are using withExternalClassFileManager (e.g. Scala.js), then they have a bug
  
I don't think there's a bug there, they are relying on an invariant that is always true for them because of sbt: there is always a previous external classfile manager, and therefore the change is applied.
  
However, I don't know why but I had the impression this change would be more harmless than it actually is. Re-reading the code gives me the confidence to merge this.

allanrenucci:
> there is always a previous external classfile manager

That is not true for Dotty, we have to do [manual aggregation](https://github.com/lampepfl/dotty/blob/f39943db4d75d81917dc055d819ef6b41292fe67/sbt-dotty/src/dotty/tools/sbtplugin/DottyPlugin.scala#L117-L126) to workaround this bug. I don't know about Scala.js, but I expect they have the same issue

PR #572, opened 2018-08-13
Traverse synthetic top-level members in API

jvican:
When a case class is defined with no companion, the companion is
synthesized by the compiler and every object creation of that case class
is proxied to the synthesized apply method of the newly synthesized
companion. From this perspective, if we have a `case class A(x: Int)`
and a use site `A(1)`, `ExtractUsedNames` will extract the name to the
apply method in `A(1)` and mark it as used.

However, this is wrong. When the user changes the signature of the case
class, only the signature of the case class constructor changes and this
change is not propagated to the apply signature (`ExtractAPI` traverses
trees, and the synthesized module has no trees as it is added in namer).
Therefore, when we compare changed names in the old and new API, Zinc
concludes that only references to `A;<init>;` must be recompiled and
since the use site that contained `A(1)` had no such reference, then
it's ignored.

<s>To fix this problem, we protect ourselves from this point of indirection
and extract the proper name of the case class constructor iff the
companion case class is indeed synthesized by the compiler. Note that
when the user defines `object A` alongside the definition of `A`, Zinc
does the right thing.</s>

After some more investigation, the problem at hand is that we don't traverse
synthetic top-level trees in `API` and therefore we don't extract the
used names from them.

This was causing us to ignore any names in the synthetic companion and,
on top of that, also synthesized top-level members like package object
definitions! Now, thanks to removing the filtering out that was
happening in `isTopLevel`, our previous issue is fixed and with it also
a pending test checking that package objects are recognized when
introduced in a change `packageobject-and-traits`.

This commit also adds a `checkRecompilations 2` in one of the tests
related to package objects to make sure that test is behaving correctly.

Note that this fixes https://github.com/sbt/sbt/issues/4316 and also
fixes issues with default parameters in case classes.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/572/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican:
I propose @smarter as a reviewer. On an unrelated topic, @eed3si9n Can we give Guillaume merge rights to this repository so that he can review?

eed3si9n:
Created https://github.com/orgs/sbt/teams/epfl-zinc/

jvican:
Thank you Eugene :smile: Can you also have a look at the PR?

eed3si9n (dismissed):
The explanation makes sense, and I like the fact that the change itself is small change to bridge code.
LGTM

eed3si9n:
Why not target 1.2.x branch?

jvican:
Because then somebody has to back port it to 1.x, which is the main master branch, and PRs get spread out across different branches that I need to manage myself downstream (in scalacenter/zinc). I would prefer if we just kept everything to 1.x and then branch out before the 1.x.n release (and cherry-picking the changes that should make it or not).

eed3si9n:
Sounds good.

smarter:
> ExtractAPI traverses
trees, and the synthesized module has no trees as it is added in namer

I don't understand this, namer is definitely going to create trees for the synthesized module, and this will be done before the extractapi phase happens, something is fishy here.

jvican:
> I don't understand this, namer is definitely going to create trees for the synthesized module, and this will be done before the extractapi phase happens, something is fishy here.

You're right, I was misled by the fact that `showCode` wouldn't show me the synthesized code so I assumed namer was setting up completers working on synthesized trees in new compilation units.
  
The issue at hand here is that we're filtering out synthetic top level members in `API.scala`, and therefore the synthetic companion is ignored. It looks safer to remove the synthetic check there as it's more general than this fix.
  
```scala
    def isTopLevel(sym: Symbol): Boolean = {
      !ignoredSymbol(sym) &&
      sym.isStatic &&
      !sym.isImplClass &&
      !sym.hasFlag(Flags.SYNTHETIC) &&
      !sym.hasFlag(Flags.JAVA) &&
      !sym.isNestedClass
    }
```
  
It looks like the filtering out of synthetic top level members comes from the dawn of time, before the Zinc 1.0 algorithm, around nine years ago: https://github.com/sbt/sbt/commit/2977fd41318838525c50f8cffe374dc450c7ba71.
  
On top of this, it looks like removing that seems to make `packageobjects-and-traits` passing.

jvican:
I've updated the code with this fix, all tests are passing locally, module the pending one that is now passing. I've added a new commit instead of squashing it with the previous one because I believe this discussion will be useful to people in the future and it's better if we let git history reflect that.
  
@smarter I was not able to find any example where extracting used names from synthetic top-level members would be problematic. Let me know if you find something.
  
@eed3si9n @smarter Could you have a look again?

smarter:
I don't think dotty filters synthetic stuff and that hasn't caused any problem as far as I know, but I haven't checked if your testcase passes currently in Dotty.

jvican:
I'm coming back to this soon. In the CI, `packageobject-and-traits` does not pass but in my computer it does. So I'm investigating what's happening (and it doesn't look related to cached compiler bridges).

eed3si9n (dismissed):

smarter (approved):

PR #576, opened 2018-08-14
Implement project matrix by hand

jvican:
This change to the build is verbose but it changes the way we work in the
repo dramatically (for good). Before, we were relying on
`publishBridges` all the time and we would get errors in `scripted` if the
bridges for all the versions were not published for every change we make in
the compiler bridge.

Despite from the previous clear disadvantages, that slowed us down since we
cannot quickly iterate in the compiler bridge. The big plus of doing this
change is that we don't need to do `±±` every time we want to publish bridges,
which removes a big memory overhead and speeds up development (`±±` is one
of the causes of metaspace errors that our users have experienced).

I've tried to use https://github.com/sbt/sbt-projectmatrix but it seemed to
fail in several ways (including some reported tickets), and since we were
already using this technique with `compilerInterfaceJava6Compat`, I found
it appropriate to extend it to the rest of the modules. In the future, we
can migrate to sbt-projectmatrix.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.2.x&zinc=pull/576/head&io=1.2.x&librarymanagement=1.2.x&util=1.2.x&website=1.2.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican:
Either Dale or Eugene, could you have a look at this change?

eed3si9n (approved):
LGTM pending CI

PR #581, opened 2018-08-22
Fix the extraHash used when reading a V1 file

smarter:
Setting extraHash to apiHash is not safe because in
IncrementalNameHashing#sameAPI an error is emitted if the extra hashes
are different and the class is not a trait. Setting it to 0 should be
safe since that's the value used for things which are not traits.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/581/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican (approved):
LGTM, thanks for the catch

PR #582, opened 2018-08-22
Notify analysis callback of generated non-local classes before genbcode

jvican:
This is my first step towards making incremental compilation more efficient. I plan to do this task in several pull requests. There will be two kind of changes: the first kind will improve the implementation of our current infrastructure (e.g. merging `Dependency` and `API` together, reducing the memory overhead of names by creating a name table, etc); the second will make structural changes to Zinc to improve performance globally. This pull request is the second kind.

The idea of this pull request is to run the incremental algorithm as soon as possible in the pipeline. We extract all compiler information right after pickler so that:
  
1. Zinc can play nicely with pipelining both at the build and target level.
2. We can run the incremental algorithm in parallel with the main compilation pipeline. (This is left as future work and requires further investigation, however the idea is to remove shared state with the main compilation pipeline (no info transformers, no symbol dependencies) so that both pipelines are independent of each other and build tools can run them independently.)
  
In this pull request, I focus on the first item: making Zinc work together with pipelining. The current problem with pipelining and incremental compilations is that they don't work well together: incremental compilation requires the main compilation pipeline to finish (before this PR it depended on information created by the last compiler phase, genbcode) and pipelining yields no benefit if we are blocked on the full compilation pipeline. The fix in this PR reconciles this situation by removing the dependency on genbcode and instead generating the names for non-local classes before flatten and genbcode are run.
  
zinc current pipelining diagram
  
The benefit is that with this change, we can run the incremental compiler right after pickler. If Zinc decides to recompile more sources, we don't spawn compilation in downstream projects; otherwise, we do. 
  
This workflow has the great advantage that we can pipeline the incremental compiler runs themselves! If we know that we need to compile again the project with more sources, we can start that compilation before the first compilation is over. In general, this allows us to aggregate a ~40% faster incremental compilation per run. An implementor of this technique must carefully make sure that the parallel compiler runs write class files/IR files to the file system in order.
  
## Implementation details
  
I added a thorough test suite to check the proper behavior of class names *before* replacing the implementation, so that we can check that it works before and after the migration of the logic from `Analyzer` to `API`. The pull request contains complete docs describing what the implementation does, but the most important thing to keep in mind is that to run the Zinc incremental algorithm we don't need to know what the local products (the class files associated with local classes) are.
  
Local classes never show up in name hashes of `AnalyzedClass`es, **meaning that they don't have any effect on the correctness of the algorithm**. Before and after this PR, a change in a local class will always be picked up. Local class files are registered in their owners, whose class file change whenever the local classes and their names change. As they don't have any effect on the correctness, we don't fully remove the logic in `API` and instead notify of their creation there. Conversely, the names and paths of non-local classes are notified in `API`, right after pickler, so that we can execute the incremental algorithm with the data collected in the analysis callback right after `pickler` has run.
  
We keep notifying about local class files not to break tools consuming the class files specified in the analysis. This approach is the best as we cannot reliably guess what the class file names of local classes will be without running the info transformer in flatten. In the case of non-local classes, guessing the class file names is much simpler and therefore allows us to remove the dependency on `genbcode`. Ideally, the compiler would provide us the information required to know what the class file path of a class symbol will be without the need of running the backend.
  
I elect @retronym and @smarter to review this pull request. I'd like to gauge some feedback on this approach.

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/API.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/API.scala
@@ -15,8 +15,11 @@ object API {
   val name = "xsbt-api"
 }
 
-final class API(val global: CallbackGlobal) extends Compat with GlobalHelpers {
+final class API(val global: CallbackGlobal) extends Compat with GlobalHelpers with ClassName {
   import global._
+
+  import scala.collection.mutable
+  private val nonLocalClassSymbolsInCurrentUnits = new mutable.HashSet[Symbol]()
 
   def newPhase(prev: Phase) = new ApiPhase(prev)
   class ApiPhase(prev: Phase) extends GlobalPhase(prev) {
@@ -25,15 +28,18 @@ final class API(val global: CallbackGlobal) extends Compat with GlobalHelpers {
     override def run(): Unit = {
       val start = System.currentTimeMillis
       super.run()
+
+      // After processing all units, register generated classes
+      registerGeneratedClasses(nonLocalClassSymbolsInCurrentUnits.iterator)
+      nonLocalClassSymbolsInCurrentUnits.clear()
+
       callback.apiPhaseCompleted()
       val stop = System.currentTimeMillis
       debuglog("API phase took : " + ((stop - start) / 1000.0) + " s")
     }
 
     def apply(unit: global.CompilationUnit): Unit = processUnit(unit)
-
     private def processUnit(unit: CompilationUnit) = if (!unit.isJava) processScalaUnit(unit)
-
     private def processScalaUnit(unit: CompilationUnit): Unit = {
       val sourceFile = unit.source.file.file
       debuglog("Traversing " + sourceFile)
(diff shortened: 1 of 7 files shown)


smarter:
> This workflow has the great advantage that we can pipeline the incremental compiler runs themselves!

That's brilliant. I never thought of that :)

smarter:
Regarding local classes, there's some context in https://github.com/sbt/sbt/issues/1104#issuecomment-169146039 by @gkossakowski  and subsequent comments, notably:

> The idea is that we want to recompile everything that will be recompiled anyway in one step. However, if we attribute dependency on A2 to B we'll invalidate all classes inheriting from B when A2 changes. That's not the behaviour we want. We cannot record that dependency as a member reference because inheritance dependencies are not filtered by name hashing (for correctness). To solve that, we need a new kind of dependency local inheritance.

jvican:
> Regarding local classes, there's some context in sbt/sbt#1104 (comment) by @gkossakowski and subsequent comments, notably
  
I was also looking at that a few days ago, and my conclusion is that the approach taken in this PR does not affect local dependencies as we still register the used names and relations appropriately, so both can coexist. The only thing that we don't do anymore after this PR is to add the generated local class files to the local products and make that accessible to the algorithm. I need to double check this tomorrow, but CI is passing and my previous investigation led me to believe this :smile:

jvican:
To elaborate more on my previous answer: this change does not conflict with the registration of local dependencies because those dependencies link the owner of the local class and the classes inherited by the local class, so no reference to the local class is ever made and therefore the local products never used in the incremental algorithm.

jvican:
I haven't mentioned it before, but note as well that still keeping the analyzer logic for local generated classes is important so that the class file manager knows all the class files that need to be persisted/removed if compilation fails.

eed3si9n (dismissed):

jvican:
@smarter Anything to add here? Otherwise I'll merge.

internal/compiler-bridge/src/main/scala/xsbt/API.scala
```diff
@@ -59,6 +65,133 @@ final class API(val global: CallbackGlobal) extends Compat with GlobalHelpers {
...
+   *
+   * 1. Offload the incremental compiler logic out of the primary pipeline and
+   *    run the incremental phases concurrently.
+   * 2. Know before the compilation is completed whether another compilation will or
+   *    will not be required. This is important to make incremental compilation work
+   *    with pipelining and enables further optimizations; for example, we can start
+   *    subsequent incremental compilations before (!) the initial compilation is done.
+   *    This can buy us ~30-40% faster incremental compiler iterations.
+   *
+   * This method only takes care of non-local classes because local clsases have no
```
smarter: typo: clsases

jvican:
Alrighty then, this is good to go.

PR #584, opened 2018-08-23
Make scripted runnable and remove bloop scripted

jvican:
Scripted can now be run like any other kind of application. As there was some
problem in the community build with depending on sbt-complete, that dependency
has been sacrificed and instead we handle the input manually.  /cc @eed3si9n

The input of the new runnable scripted is a list of scripted filters followed
by an optional 'true/false' argument to enable/disable log buffering.

The most notorious improvement caused by this change is that now we don't need
to resolve and compile the bridge, but instead we just get the output of
classes directory of every bridge, put it in a jar and classload it. This
speeds up test execution and will make it easier to run dotty's bridge. Down
the road, I want to test the dotty bridge on sbt/zinc's CI.

After this change, the bloop scripted module is no longer necessary as bloop
can run the application in the `zincScripted-test` target. All scripted tests
run in 97s after this change in my laptop.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/584/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican:
This pull request also puts a little bit of order in the scripted test directory by moving the macro tests to its own folder and by fixing some tests that used a concrete scala version (which we cannot resolve anymore in the runnable version of scripted), as well as removing some stale scripted tests.
  
Note that the previous version of scripted is left untouched and can still be run as before.

jvican:
After this change, the CI is 1 minute and a half faster (out of 10, so 10%) than running scripted via the scripted sbt integration. I suspect some of this speedup does not only come from not resolving the bridges, but also from the fact that we don't class load the zinc artifacts in the sbt universe as we did when using `zincRoot/scripted`.

eed3si9n (approved):

PR #585, opened 2018-08-26
Document internals of invalidation and hook profiler

jvican:
While I was trying to implement some improvements here, I realize that
the current code can hardly be read and reused. I struggled as well to
understand the nitty-gritty of the algorithm as the control flow was not
clear and the order of functions was confusing.

This PR aims to fix this by reorganising the code to allow for more
reuse (functions in `IncrementalCommon`'s companion), reorganizes the
outline of `IncrementalCommon` to have clearer names and structure and
documents the most important parts of the code with implementation and
contract details. Aside from that, it introduces a protobuf data to
track all the invalidation logic. This protobuf is internal to Zinc
and should only be used by users that really know what they are doing --
as the format can change in the future (protobuf makes it easy to evolve
this profiling schema). This is related to the idea I pitched in in #550.

With this idea, this data will not only be useful for debugging but for
providing an automatic way of reporting bugs in Zinc. The infrastructure
is far from finished but it's already in a usable state for libraries
that depend on Zinc directly and have direct access to `Incremental`.
For performance reasons, a string table is kept per zinc profiler. More
details in the scaladoc.

**Note that by default no profiler is used**. Only people that change the
profiler argument for `Incremental.compile` will be able to get the run
profiles. How these profiles are persisted is out of scope of this PR.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=pull/585/head&io=1.x&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican:
I'll fix the MiMa errors later on, but this PR is ready for review.

jvican:
MiMa errors are fixed, CI passing.

internal/zinc-core/src/main/protobuf/zprof.proto
```diff
@@ -0,0 +1,68 @@
+syntax = "proto3";
```
muuki88, line 1: It would be nice to add more documentation on the messages and the `zprof.proto` proto file :smiley: 
Especially answering questions like

- why is this message necessary?
- who/what might actually use this message to solve which problem
- a remake to this pull request which marks the initial implementation

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
```diff
@@ -0,0 +1,267 @@
...
+ * A profiler interface should only be used by project, and not globally, as
+ * this class is not thread safe.
+ */
+abstract class InvalidationProfiler {
+  def profileRun: RunProfiler
+  def registerRun(run: zprof.ZincRun): Unit
+}
+
+object InvalidationProfiler {
+  final val empty: InvalidationProfiler = new InvalidationProfiler {
```
muuki88, line 25: From the implementation I guess this is being used when no profiling should be done?

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
```diff
@@ -0,0 +1,267 @@
...
+class ZincInvalidationProfiler extends InvalidationProfiler {
+  private final var lastKnownIndex: Long = -1L
+  private final val stringTable1: ArrayBuffer[String] = new ArrayBuffer[String](1000)
+  private final val stringTable2: ArrayBuffer[String] = new ArrayBuffer[String](10)
+  private final val stringTableIndices: mutable.HashMap[String, Long] =
+    new mutable.HashMap[String, Long]
+
+  def profileRun: RunProfiler = new ZincProfilerImplementation
+
+  private final var runs: List[zprof.ZincRun] = Nil
```
muuki88: Can you add some information why we need to persist every run here?

Also this list grows with every compile run if I understand this correctly?
AFAIK it's only used for serialization to disc. Should we empty the runs after
serialization? Or is the memory footprint  small enough that we should never care

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
```diff
@@ -0,0 +1,267 @@
...
+    ()
+  }
+
+  /**
+   * Returns an immutable zprof profile that can be serialized.
+   *
+   * It is recommended to only perform this operation when we are
+   * going to persist the profiled protobuf data to disk. Do not
+   * call this function after every compiler iteration as the aggregation
+   * of the symbol tables may be expensive, it's recommended to
```
muuki88: Why is this recommended? :smile:

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
```diff
@@ -0,0 +1,267 @@
...
+    override def profileRun: RunProfiler = RunProfiler.empty
+    override def registerRun(run: zprof.ZincRun): Unit = ()
+  }
+}
+
+class ZincInvalidationProfiler extends InvalidationProfiler {
+  private final var lastKnownIndex: Long = -1L
+  private final val stringTable1: ArrayBuffer[String] = new ArrayBuffer[String](1000)
+  private final val stringTable2: ArrayBuffer[String] = new ArrayBuffer[String](10)
+  private final val stringTableIndices: mutable.HashMap[String, Long] =
```
muuki88: Can you add some information what these indices are for? `String` / `Long` are very generic types and people like me with very few compiler experience would find one sentence explaining the variable quite helpful :smile:

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
```diff
@@ -0,0 +1,267 @@
...
+        shouldCompileIncrementally = shouldCompileIncrementally
+      )
+
+      cycles = newCycle :: cycles
+      ()
+    }
+  }
+}
+
+abstract class RunProfiler {
```
muuki88: Same here :wink: I find some scaladocs on a fresh API usually very helpful to get
the intention and purpose of the API. The question _why is this interface necessary_
should be answered in the scaladocs :smiley:

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
```diff
@@ -0,0 +1,267 @@
...
+      startNanos: Long,
+      durationNanos: Long
+  ): Unit
+
+  def registerInitial(
+      changes: InitialChanges
+  ): Unit
+
+  def registerEvent(
+      kind: String,
```
muuki88, line 218: What `kind`s are allowed here? Is there any other implementation I can look at to see what
is applicable here?

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
```diff
@@ -0,0 +1,267 @@
...
+        zprof.InitialChanges(
+          changes = Some(profChanges),
+          removedProducts = toStringTableIndices(toPathStrings(changes.removedProducts)).toList,
+          binaryDependencies = toStringTableIndices(toPathStrings(changes.binaryDeps)).toList,
+          externalChanges = toApiChanges(changes.external).toList
+        )
+      )
+    }
+
+    private final var currentEvents: List[zprof.InvalidationEvent] = Nil
```
muuki88, line 153: Same as with `runs`. This grows unbounded, right?

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
```diff
@@ -0,0 +1,267 @@
...
+        kind = kind,
+        inputs = toStringTableIndices(inputs).toList,
+        outputs = toStringTableIndices(outputs).toList,
+        reason = reason
+      )
+
+      currentEvents = event :: currentEvents
+    }
+
+    private final var cycles: List[zprof.CycleInvalidation] = Nil
```
muuki88, line 170: Same as with `runs`. This grows unbounded, right?

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
```diff
@@ -0,0 +1,267 @@
...
+   *
+   * It is recommended to only perform this operation when we are
+   * going to persist the profiled protobuf data to disk. Do not
+   * call this function after every compiler iteration as the aggregation
+   * of the symbol tables may be expensive, it's recommended to
+   * persist this file periodically.
+   *
+   * @return An immutable zprof profile that can be persisted via protobuf.
+   */
+  def toProfile: zprof.Profile = zprof.Profile(
```
muuki88: This is *the* most important method in the whole implementation, right?
It creates a serializable format for the profile data  gathered so far. Why isn't this part
of the public API?

muuki88 (commented):
Thanks for all the comments, making things more understandable for a zinc newcomer :+1: :heart_eyes: 
But I want more :joy: 

The general implementation seems reasonable from the understanding I have so far (which can be described as almost nothing :stuck_out_tongue: ). The only question I have is why the `toProfile` method is not part of the public API. It seems to me that this is *the* method to get the profile data in a serializable format

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
jvican: Yes

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
jvican: The `ZincProfiler` is managed outside of Zinc, by another tool. Whomever uses this profiler is responsible for managing the lifetime of the profilers used. That being said, the memory profile of the runs is very small given that we use a string table

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
jvican: Because the string table for very big projects could be reasonable big, so you want to make the most out of it for as many runs as you can

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
jvican: What would you write? I didn't write anything because i thought the variable names were self-descriptive, i.e. `stringTableIndices` means a map from string to indices of the string table.

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
jvican: What do you mean by this grows unbounded? In every run, we create a new `RunProfiler` so that new profile wont' see events from the past

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
jvican: It's in generic on purpose so that we can easily involve it in the future, even though we only have one implementation. The kinds are defined in https://github.com/sbt/zinc/pull/585/files#diff-de86f02542b47ffddbb8687a94316daeR259

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
jvican: Because this exposes internal details of the algorithm, and I only want people that really know what they're doing using it :wink:

jvican:
@muuki88 Thanks for the review, I answered most of your comments. Regarding the additions of more docs (for example in the profiling schema), I'm afraid I won't be able to do so until the beginning of October. My schedule for the next 4 weeks looks pretty busy.

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
muuki88: Ah, thanks :+1: Misunderstood that.

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
muuki88: okay :smile:

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
muuki88: what I don't understand (yet) is what these strings are. Class names, variable names, file paths or all of this :grin: And `long` is the index of what. 

If you say it's self explaining once one is deeper into how things work, I'm fine with not putting any docs on it :wink:

muuki88:
@jvican thanks for the quick feedback :hugs:

> I'm afraid I won't be able to do so until the beginning of October. My schedule for the next 4 weeks looks pretty busy.

no worries :smile:

internal/zinc-core/src/main/scala/sbt/internal/inc/InvalidationProfiler.scala
jvican: No, I think you make a good point. It's likely that I'm too much into the implementation that I see it obvious, but it isn't. Thanks for pointing that out, I'll add a quick comment.

jvican:
@muuki88 OK, motivated by your comments I quickly went into the GitHub interface and submitted two changes: more docs and an improvement to the internals of the profiler. Feel free to have a look at them and tell me if they are useful.

eed3si9n:
> Aside from that, it introduces a protobuf data to track all the invalidation logic.

Given that Google has not move protobuf towards JDK 9/10/11 yet, I am concerned about bringing Protobuf in as a critical piece. As far as I know, the current usage is somewhat limited to Analysis serialization only right? What do you think about using Contraband instead here?

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -11,321 +11,348 @@ package inc
...
-  }
-  private[this] def expand(invalidated: Set[File], all: Set[File]): Set[File] = {
-    val recompileAllFraction = options.recompileAllFraction
-    if (invalidated.size > all.size * recompileAllFraction) {
-      log.debug(
-        "Recompiling all " + all.size + " sources: invalidated sources (" + invalidated.size + ") exceeded " + (recompileAllFraction * 100.0) + "% of all sources")
-      all ++ invalidated // need the union because all doesn't contain removed sources
-    } else invalidated
-  }
+  def recompileClasses(
```
eed3si9n, line 160: Documentation, reorganization, and some changes are intermixed in this commit, so it's difficult for me to preview, but have you made any intentional change in the observable behavior, either in the difference in output or in performance characteristics? For example, I noticed that the method named `recompileClasses` are no longer recompiling classes, but it's recompiling sources.

jvican: There’s no change in semantics, I double checked this several times manually and the tests confirm it.

jvican:
> Given that Google has not move protobuf towards JDK 9/10/11 yet, I am concerned about bringing Protobuf in as a critical piece. As far as I know, the current usage is somewhat limited to Analysis serialization only right? What do you think about using Contraband instead here?

It can be ported in the future if need so, but the elephant in the the room is the analysis file implementation. Conversely, this use of protobuf can be easily disabled without affecting the correctness of the language (and in fact by default it won’t even initialize the protobuf classes — the warning won’t happen), so I wouldn’t worry about it. The protobuf issue will be fixed down the road (I have high confidence in that), and if it doesn’t we can find solutions by then.

muuki88:
Awesome @jvican ! That clarifies a lot of things 😍

eed3si9n:
> It can be ported in the future if need so, but the elephant in the the room is the analysis file implementation.

Analysis file implementation can be switched back to using text file like sbt 0.13.

> Conversely, this use of protobuf can be easily disabled without affecting the correctness of the language

If it's easy to not use protobuf, let's switch to using Contraband like the rest of the datatypes.

jvican:
> If it's easy to not use protobuf, let's switch to using Contraband like the rest of the datatypes.
  
I think you misread me 😛 I didn't say it's easy to switch to Contraband (it would actually take me quite a lot of time that I cannot afford now), I said it's easy not to use the protobuf-based profiler in the future. By default, the profiler is empty and in the worst-case scenario if protobuf doesn't remove the warning in JDK 11 you can (1) no use any profiler at all or (2) implement your own (and that is if you use it in the first place).

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
jvican: Wrt the change you mention @eed3si9n, the reason why you see that we recompile sources instead of classes in a clearer way is because `recompileClasses` was doing the mapping between invalidated classes and sources to recompile, and I pulled that logic out of it to keep things clear.

eed3si9n (approved):

PR #591, opened 2018-08-31
Fix spurious problems with unset `associatedFile`

jvican:
A thorough analysis and description of the issue is in the documentation
of the fix.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/591/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican:
I'd like to add some performance improvements to the implementation in case it silently regresses, so let's not merge this yet.

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -126,18 +126,42 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
...
+        if (targetSymbol.hasFlag(scala.tools.nsc.symtab.Flags.PACKAGE)) None
+        // Ignore `Any` which by default has no `associatedFile`
+        else if (targetSymbol == definitions.AnyClass) ()
+        else {
+          classFile(targetSymbol) match {
+            case Some((at, binaryClassName)) =>
+              // Associated file is set, so we know which classpath entry it came from
+              processExternalDependency(binaryClassName, at)
+            case None =>
+              /* If there is no associated file, it's likely the compiler didn't set it correctly.
```
eed3si9n, line 149: Are we covering up for compiler? Should we report this upstream?

jvican: In the general case, we're covering some error, but we don't know which error :smile: We've been lucky and we've found one of these errors by surfacing the bug with default arguments. I've not opened a bug report because it's not clear how to reproduce, but I've given steps to Lukas and pinged him on the bug ticket that this PR fixes so that he can clarify the situation. We'll move from there.
  
The issue here is that fixing this instance of errors is not trivial and have to be fixed bug per bug because of the way the class file parser works.

OlegYch:
any idea when this might be released?

eed3si9n:
@OlegYch it's currently in develop branch, so it will come out when sbt 1.3.0 comes out in a few months. If the patch is considered relatively stable, someone can backport it to 1.2.x branch and then it will be part of next sbt 1.2.x, which likely will come out sooner.

PR #592, opened 2018-08-31
Upgrade to 2.13.0-M5

jvican:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/592/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican:
Looping in @eed3si9n To correctly upgrade to 2.13.0-M5 I need published versions of:
  
- [ ] scalatest
- [ ] sbt-io
- [ ] sbt-util-logging
- [ ] scalacheck

SethTisue:
ScalaTest and ScalaCheck have both published JVM artifacts for M5.

can sbt-io and sbt-util-logging be published...?

eed3si9n:
See https://github.com/sbt/zinc/pull/453. The compiler bridge itself depends only on compiler|util-interface, which are both Java library. What requires IO and Util Logging is the unit tests for compiler bridge for making a harness for testing called ScalaCompilerForUnitTesting - https://github.com/sbt/zinc/blob/v1.2.3/internal/compiler-bridge-test/src/test/scala/xsbt/ScalaCompilerForUnitTesting.scala
As it stands, we should be able to proceed without the unit test.

We can still do testing using the scripted test, which accepts scalaVersion. I put them in, but it seems like they were removed in https://github.com/sbt/zinc/pull/584/commits/d5b8a53b14e28c07044065d1c9aecea115fdde2c. The bridge check scripted tests go through the motion of actually compiling something, so I think we should bring them back as a quick check that the bridge functions.

We should then port the compiler bridge unit tests to actual Zinc harness that does not require IO and Util cross built to 2.13.0-M5. Getting Util Logging to compile is going to take a while since it depends on series of (somewhat old) JSON libraries.

eed3si9n:
> We should then port the compiler bridge unit tests to actual Zinc harness that does not require IO and Util cross built to 2.13.0-M5.

https://github.com/sbt/zinc/pull/607

eed3si9n:
This was superseded by https://github.com/sbt/zinc/pull/607

Issue #594, opened 2018-09-04
Removing default value of case class constructor doesn't invalidate usages


jrudolph:
I'm seeing an issue when removing a default value for a top-level case class field. This doesn't invalidate and rebuild any files (in the same project) that depended on the default value being present.

The project uses sbt 1.2.1 / scala 2.12.6.

It can be reproduced by compiling akka-http at https://github.com/akka/akka-http/pull/2185/commits/ce622a16a8fa22a72dfc4121695805ae9a7436f7 and then removing the default parameter value from `EntityStreamSizeException.actualSize` which doesn't trigger any downstream recompilations at the usage sites.

When `akka-http-tests/testOnly akka.http.scaladsl.server.SizeLimitSpec` is run `java.lang.NoSuchMethodError: akka.http.scaladsl.model.EntityStreamSizeException$.apply$default$2()Lscala/Option;
` is thrown.

I'll try to come up with a reproducer.

(I first thought it might be related to #559 but from @jvican's analysis so far, this one seems to be a broader issue.)

jvican:
Thanks @jrudolph I’ll have a look at this on Thursday.

jrudolph:
Thanks, @jvican, I'll try to minify until then.

jrudolph:
It's as simple as that:

Main.scala:

```scala
object Main extends App {
  println(A())
}
```

A.scala:

```scala
case class A(value: Int = 0)
```

Compile first with the default parameter. Then remove it and only A.scala will be recompiled.

jrudolph:
It's interesting that the api diff doesn't even seem to know about the case class apply methods:

```
[debug] [diff] object A extends  {
[debug] [diff] 
[debug] [diff] 
[debug] [diff] 
[debug] [diff] }
```

jvican:
At first sight it looks quite familiar to the bug I fixed one week ago, but in that one the api change in the synthetic default methods was correctly reported by zinc.

PR #596, opened 2018-09-07
Add passing test case for default parameters in classes

jvican:
This pull request adds a test case for @jrudolph's issue https://github.com/sbt/zinc/issues/594. However, it is passing, so it seems that the issue he diagnosed is not reproducible in master and may have already been fixed.
  
As I'm not sure if the minimized example provided in the ticket is for sure a good reproduction of the original issue in akka, @jrudolph would you mind publishing locally zinc from master and checking if the error persists in the akka build?


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/596/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jrudolph:
Thanks a lot, @jvican.

> publishing locally zinc from master and checking if the error persists in the akka build?

Sure. What would be the fastest way to make sure that the local zinc version would be picked up by sbt?

jvican:
> Sure. What would be the fastest way to make sure that the local zinc version would be picked up by sbt?
  
I think the fastest way would be to publish a local sbt version of the sbt version you use in the akka build with the aforementioned change in zinc.

jrudolph:
Ok, I'll have a try.

jrudolph:
I just tried your test on a few older commits of zinc and it seems 746d474 has fixed it (which was pretty likely based on its commit message).

jrudolph (approved):
LGTM

jvican:
Ha, can you believe I didn't remember fixing this before? I even had a lot of tests to cover this scenario. Well, I'm happy this is now fixed. I'm closing since we're already testing this. Thank you @jrudolph for following this through and the initial bug report -- next time maybe we get to find an issue with latest master :smile:

jrudolph:
Yep, having master already fixed it is really the second best option (after no problem at all ;) )

PR #603, opened 2018-10-01
Update IO.classfileLocation usage

eatkins:
I was having trouble getting sbt tests to pass with the new version of
io. I determined that the big issue is that value returned by
IO.classfileLocation was different from what was previously expected.
Zinc specifically relied on the IO.classfileLocation returning a full
path name. In the case of a jar, this meant a jar url with a
"!/$CLASSNAME" at the end. In the case of a file in a directory, it was
a file url that ended with $CLASSNAME. After io/185, the
IO.classfileLocation would generally return the class file directory for
a particular class, but not the full absolute path to the class.

In the io project, I updated IO.classfileLocation to always return a url
to just the jar file or the base directory. It is then up to the caller
to append the classname. If I made IO.classfileLocation return the
absolute path of the class path, that broke other tests in sbt that
assumed it returned the base directory for the class.

Linked to 
https://github.com/sbt/io/pull/188

### DIFF
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -204,7 +204,7 @@ object ClassToAPI {
 
   /** TODO: over time, ClassToAPI should switch the majority of access to the classfile parser */
   private[this] def classFileForClass(c: Class[_]): ClassFile =
-    classfile.Parser.apply(IO.classfileLocation(c))
+    classfile.Parser.apply(c)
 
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/603/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
The tests do pass when run locally against an io snapshot from 
https://github.com/sbt/io/pull/188.

PR #606, opened 2018-10-09
[1.2.x] Fix the extraHash used when reading a V1 file

smarter:
This backports #581 which fixes an issue introduced in #542, the fix
wasn't backported when #542 itself was backported in #599.

Setting extraHash to apiHash is not safe because in
IncrementalNameHashing#sameAPI an error is emitted if the extra hashes
are different and the class is not a trait. Setting it to 0 should be
safe since that's the value used for things which are not traits.

### DIFF
--- a/internal/zinc-persist/src/main/scala/sbt/internal/inc/binary/converters/ProtobufReaders.scala
+++ b/internal/zinc-persist/src/main/scala/sbt/internal/inc/binary/converters/ProtobufReaders.scala
@@ -542,8 +542,8 @@ final class ProtobufReaders(mapper: ReadMapper, currentVersion: schema.Version) 
     val name = analyzedClass.name
     val api = mkLazy(analyzedClass.api.read(fromCompanions, ExpectedCompanionsInAnalyzedClass))
     val apiHash = analyzedClass.apiHash
-    // Default on api hash to avoid issues when comparing hashes from two different analysis formats
-    val extraHash = if (currentVersion == schema.Version.V1) apiHash else analyzedClass.extraHash
+    // Default to 0 to avoid issues when comparing hashes from two different analysis formats
+    val extraHash = if (currentVersion == schema.Version.V1) 0 else analyzedClass.extraHash
     val nameHashes = analyzedClass.nameHashes.toZincArray(fromNameHash)
     val hasMacro = analyzedClass.hasMacro
     AnalyzedClass.of(compilationTimestamp, name, api, apiHash, nameHashes, hasMacro, extraHash)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.2.x&zinc=pull/606/head&io=1.2.x&librarymanagement=1.2.x&util=1.2.x&website=1.2.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

smarter:
@eed3si9n I suggest releasing an sbt 1.2.5 asap with that fix in to prevent more confusion like https://gitter.im/sbt/sbt?at=5bbbbcea6e5a401c2d39580f

eed3si9n (approved):

eed3si9n:
Sounds like a good idea to get it out soon.

SethTisue:
to help searchers, the error message is "different extra api hashes for no traits" (at least, that's the one I've seen)

jvican (approved):

smarter:
also to help searchers: you can work around this issue by running "sbt clean".

patriknw:
`clean` doesn't help for me. I see this on current akka master (bf4a6c37b4) now

patriknw:
updating to sbt 1.2.6 solves the issue (didn't know it was released at first)

avdv:
Hi.

`clean` also did not help for me, when upgrading to 1.2.4.

I now also see this when using 1.2.6:
```
[error] one error found
[error] java.lang.RuntimeException: A fatal error happened in `SameAPI`: different extra api hashes for no traits!
[error]   `eve.publish.codegen.package`: 0
[error]   `eve.publish.codegen.package`: 1126415563
[error]              
[error] 	at scala.sys.package$.error(package.scala:26)
[error] 	at sbt.internal.inc.IncrementalNameHashing.sameAPI(IncrementalNameHashing.scala:52)
[error] 	at sbt.internal.inc.IncrementalCommon.sameClass(IncrementalCommon.scala:204)
[error] 	at sbt.internal.inc.IncrementalCommon.$anonfun$changedIncremental$1(IncrementalCommon.scala:186)
```
Could this be due to me upgrading to 1.2.4 first, which already writes a v2 file with the extra hash information?

jvican:
That would be my guess, yes :+1:

mariszin:
Hi!

I have the same issue as @avdv - upgrading to 1.2.6 helped 1 project, for other I still see the error:
```
[error] java.lang.RuntimeException: A fatal error happened in `SameAPI`: different extra api hashes for no traits!
[error]   `org.recmap.MapRecursiveExtensions`: 0
[error]   `org.recmap.MapRecursiveExtensions`: 1182952996
```
I deleted ~/.ivy2/cache contents, all target directories, and still the same error.

jvican:
Are you sure you have removed all target directories in your project? The fact that sbt is able to deserialize the previous compilation analysis means you probably haven’t cleaned all the target directories, as that’s the only place where the analysis files are written.

mariszin:
Is it possible, the issue might stem from a dependency, that is a git repository? 
It has sbt=1.2.4 in build.properties. If I change checkout it locally and change to 1.2.6,, it compiles without error. Going back to 1.2.4 does not produce error any more.

jvican:
> Is it possible, the issue might stem from a dependency, that is a git repository?
  
Yes, absolutely :smile: If you're using a git repository as a dependency and you're not yet on 1.2.4, sbt will use the zinc compiler in 1.2.4 to compile.

scalavision:
I got this error when upgrading from 1.2.3 to 1.2.6 in scalaz-zio build. 
```bash
[error] java.lang.RuntimeException: A fatal error happened in `SameAPI`: different extra api hashes for no traits!
[error]   `scalaz.zio.package`: 0
[error]   `scalaz.zio.package`: -1375558950
```
It might be possible to reproduce it from this branch ..???

https://github.com/scalavision/scalaz-zio/tree/hashing-failure-with-1.2.6-sbt.

So, might this have been introduced before 1.2.4?

eed3si9n:
I just ran into this using sbt 1.2.6, and for some reason running `clean` did not solve the issue.
But running `git clean -f` worked.

jvican:
Probably because of aggregation not covering all projects in the build?

eed3si9n:
That's what I thought too, but it was lm, and lmRoot covers all the subprojects:

```
lazy val lmRoot = (project in file("."))
  .aggregate(lmCore, lmIvy, lmCoursier)
```

including the ones I had issue with, which I think was core. I wonder if there's some cache stickiness caused by the metabuild, which `clean` won't get to.

jvican:
> I wonder if there's some cache stickiness caused by the metabuild, which clean won't get to.
  
Agreed, sounds like a plausible hypothesis.

trobert:
If it can help, I just had the issue when migrating a project from 1.2.3 to 1.2.6. Issuing `sbt clean` in the **project** directory solved it , which corroborates with the metabuild hypothesis.

jlncrnt:
> If it can help, I just had the issue when migrating a project from 1.2.3 to 1.2.6. Issuing `sbt clean` in the **project** directory solved it , which corroborates with the metabuild hypothesis.

Same. I ran `clean` while in sbt console on root project and it didn't worked but it worked when I ran `sbt clean` from the terminal inside the project dir.

grindvoll:
I just encounteded the same issue with SBT 1.2.6. In my case, an `sbt clean` inside the project dir didn't resolve the issue. From my experience, it seems that the frequency of the error has been reduced, but there is still a chance that it will turn up.

PR #618, opened 2018-11-29
Remove false error message (compile-to-jar)

lukaszwawrzyk:
The error here is not an error actually. If the classFile is not relative to output it just most likely means it is an external jar or something. It should be returned as is and will be handled elsewhere to be recognized as external dependency. Notice that for non compile-to-jar flow there is no such check.

It turned out that this innocent error message can cause some serious problems. As we report error here, when building from intellij using this version of zinc it will be recognized as compilation error and the building will fail.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/618/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican (approved):

PR #619, opened 2018-11-29
Optimize finding output dir for source

lukaszwawrzyk:
The original `lazy val outputDir` was dead code. Looking at it I noticed that we have here some indirection, that is not needed (?).
The method is processing compilation unit which maps 1:1 with source file, and we have it already at the beginning of the method. But then for each class produced from this source we get the source file from the symbol and resolve the output dir each time.
After my change output dir is resolved once per source file.

Also, the `unit.source.file` seems to be more reliable. I faced some uncommon case when sym.sourceFile would return null, causing NPE when looking for output dir. Though it is most likely something with scalac itself. The change in this PR also solves that problem.

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/Analyzer.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/Analyzer.scala
@@ -45,15 +45,15 @@ final class Analyzer(val global: CallbackGlobal) extends LocateClassFile {
 
     def apply(unit: CompilationUnit): Unit = {
       if (!unit.isJava) {
-        val sourceFile = unit.source.file.file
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/619/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican (approved):
LGTM :+1:

PR #627, opened 2018-12-16
Remove obsolete API usage

som-snytt:
Remove it here so it can be removed on scala-2.13.

https://github.com/scala/scala/pull/7544

### DIFF
--- a/internal/compiler-bridge/src/main/scala_2.13/xsbt/ConsoleInterface.scala
+++ b/internal/compiler-bridge/src/main/scala_2.13/xsbt/ConsoleInterface.scala
@@ -47,7 +47,6 @@ class ConsoleInterface {
               if (loader eq null) super.parentClassLoader
               else loader
           }
-          intp.setContextClassLoader()
         } else
           super.createInterpreter(interpreterSettings)
 


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/627/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/compiler-bridge/src/main/scala_2.13/xsbt/ConsoleInterface.scala
```diff
@@ -47,7 +47,6 @@ class ConsoleInterface {
               if (loader eq null) super.parentClassLoader
               else loader
           }
-          intp.setContextClassLoader()
```
eed3si9n, line 50: Would this mean using some range of Scala 2.13.x would require some range of future version of sbt?

som-snytt:
Don't want to be that guy with the ancient PR. Scala 2 is almost over anyway.

PR #628, opened 2018-12-18
Do not swallow exception when reading analysis

lukaszwawrzyk:
`FileAnalysisStore` swallows exceptions when reading analysis. It is expected to return `Optional.empty` when file does not exists, this seems to be a check that zinc relies on, hence I had to add the check in the method. But any other case probably indicates some kind of error, leading to unexpected full recompilations and leaving no information to debug those.
In fact most weird problems faced with zinc were hidden exactly in this place.
To name one: sometimes analysis could not be read. It turned out to be because of protobuf max message length (it was throwing, but exception was swallowed). That was because there were 2 protobufs on the classpath, one was the old version with lower limit. With many moving pieces related, debugging that it was related to this place took a while, an exception would make it more obvious. And there was a number of other issues that were simply hidden there.

I can understand that it changes the behavior and other code may depend on that. Maybe this should be a part of the "strict mode" as discussed in https://github.com/sbt/zinc/pull/617 ?

### DIFF
--- a/internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
+++ b/internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
@@ -50,21 +50,21 @@ object FileAnalysisStore {
      * Get `CompileAnalysis` and `MiniSetup` instances for current `Analysis`.
      */
     override def get: Optional[AnalysisContents] = {
-      import JavaInterfaceUtil.EnrichOption
-      val nestedRead: Option[Option[AnalysisContents]] = allCatch.opt {
+      if (file.exists()) {
         Using.zipInputStream(new FileInputStream(file)) { inputStream =>
           lookupEntry(inputStream, analysisFileName)
           val reader = CodedInputStream.newInstance(inputStream)
           val (analysis, miniSetup) = format.read(reader)
-          val analysisWithAPIs = allCatch.opt {
+          val analysisWithAPIs = {
             lookupEntry(inputStream, companionsFileName)
             format.readAPIs(reader, analysis, miniSetup.storeApis)
           }
-
-          analysisWithAPIs.map(analysis => AnalysisContents.create(analysis, miniSetup))
+          val contents = AnalysisContents.create(analysisWithAPIs, miniSetup)
+          Optional.of(contents)
         }
+      } else {
+        Optional.empty()
       }
-      nestedRead.flatten.toOptional
     }
 
     /**


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/628/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
```diff
@@ -50,21 +50,21 @@ object FileAnalysisStore {
...
     override def get: Optional[AnalysisContents] = {
-      import JavaInterfaceUtil.EnrichOption
-      val nestedRead: Option[Option[AnalysisContents]] = allCatch.opt {
+      if (file.exists()) {
         Using.zipInputStream(new FileInputStream(file)) { inputStream =>
           lookupEntry(inputStream, analysisFileName)
           val reader = CodedInputStream.newInstance(inputStream)
           val (analysis, miniSetup) = format.read(reader)
-          val analysisWithAPIs = allCatch.opt {
+          val analysisWithAPIs = {
```
eed3si9n, line 58: Instead of changing the existing semantics, why don't we add another API that returns `Either[Option[String], AnalysisContents]` or something like that, so we can surface the error as a warning message.

eed3si9n (changes requested):

internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
eatkins: Could we do better with `Either[AnalysisException, AnalysisContents]` or `Try[AnalysisContents]`? Using Either preserves a bit more type information than Try. I made up AnalysisException. It would be whatever exception type the api throws.

eed3si9n: I included `Option[String]` in there because analysis file not existing is a case that needs to be handled. Maybe it's better to have `Try[Option[AnalysisContents]]` or `Either[something, Option[AnalysisContents]]`.

eatkins: Oh, I see. I just saw the title was "Do not swallow exception" and assumed there was some exception that was being caught and dropped so I was just advocating preserving that information since `Option[String]` implied a loss of information to me.

lukaszwawrzyk: @eatkins You assumed correctly. `allCatch.opt` does catch and drop exceptions. These are not expected exceptions (like for missing file) but some rare cases like e.g. corrupt analysis file or ClassDefNotFoundError or any other random issues. Currently they will just be ignored and treated equivalently to missing file.
@eed3si9n It would still be an improvement to add separate method (any suggested name), though I believe this new method should be called by zinc internally, because otherwise the error will be lost as before. I would be in favour of throwing the exception, but I guess logging could be ok, yet harder to spot.

eed3si9n: Once there's `Try` or `Either` the consumer of Zinc can take appropriate action, including failing the build etc..

lukaszwawrzyk: I tried doing that. It seems that adding the new method would require altering the api of `xsbti.compile.AnalysisStore` and I think scala types like `Try` or `Either` shouldn't be used there.
I was thinking about adding `getUnsafe` variant that would throw, and keep `get` as `try getUnsafe() catch { case e: Exception => Optional.empty() }`. Though it feels a bit odd that (in my opinion) the preferred method would be the unsafe variant. I still think that the allCatch there was not the best decision and should just be replaced as it currently is in this PR.

SethTisue:
closing for inactivity. happen to reopen if activity resumes.

PR #629, opened 2018-12-18
Allow disabling compression when saving analysis

lukaszwawrzyk:
This is a thing that was on my compile-to-jar PR but we decided to leave for later, so here it is.
This adds an option to disable zip compression while storing analysis. It allows to save some time when storing the analysis if the space is not big of a concern. 
It is implemented through system property. Makes it easy to enable globally and does not require changing the API, but I am open for suggestions how to do it better.

### DIFF
--- a/internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
+++ b/internal/zinc-persist/src/main/scala/sbt/internal/inc/FileAnalysisStore.scala
@@ -11,7 +11,7 @@ package inc
 
 import java.io._
 import java.util.Optional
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/629/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

SethTisue:
closing for inactivity. happy to reopen if activity resumes

PR #632, opened 2019-01-18
Replace WriterOutputStream with internal implementation

lukaszwawrzyk:
I am working on integrating the compile-to-jar functionality with scala plugin for intellij.
This small change https://github.com/sbt/zinc/pull/610 that introduced a call to `scala.tools.nsc.interpreter.WriterOutputStream` causes `ClassNotFoundException` in the compile server, as scalac jars are not on the classpath. They are passed individually for each compilation.
Converting `Writer` to `OutputStream` is sadly not implemented in the JDK. 
It is e.g. in `org.apache.commons.io.input.WriterInputStream`, but it is not on our classpath either.
As the class is small I think it is ok to have our own implementation like that instead of pulling it from internals of scalac.

### DIFF
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/LocalJava.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/LocalJava.scala
@@ -12,8 +12,9 @@ package inc
 package javac
 
 import java.io.{ File, InputStream, OutputStream, PrintWriter, Writer }
-import java.nio.charset.Charset
 import java.util.Locale
+import java.nio.{ ByteBuffer, CharBuffer }
+import java.nio.charset.{ Charset, CodingErrorAction }
 
 import javax.tools.JavaFileManager.Location
 import javax.tools.JavaFileObject.Kind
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/632/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

lukaszwawrzyk:
@jvican Could you take a look? I still have some work to do on the scala plugin, I am working on a local version right now, but it would be nice to have it merged/released some time soon.

jvican (approved):
LGTM, I'm surprised you use such small byte and char buffers, but I guess it's OK.

lukaszwawrzyk:
Thanks! 
I was a bit surprised by buffers as well, though it is exactly the same as it was before this PR, as the class is almost a copy of `scala.tools.nsc.interpreter.WriterOutputStream`.

lukaszwawrzyk:
@eed3si9n any chance of releasing zinc with this patch?

lukaszwawrzyk:
Actually would be best to release it after https://github.com/sbt/zinc/pull/633

PR #633, opened 2019-01-25
Handle compile to jar switch

lukaszwawrzyk:
I added a small check. If previously we were compiling to class files, the analysis with previous products will contain plain files. If current compilation is to jar (compile-to-jar was just enabled), it will try to do incremental compilation based on plain class files, which does not make sense. Most of the code in zinc assumes that if compile to jar is enabled for current compilation it was enabled for the previous one, which is convenient. So I added this simple check in the beginning - if a product of previous compilation was not a class in jar, compilation starts with empty analysis.

### DIFF
--- a/internal/zinc-classfile/src/main/scala/sbt/internal/inc/JarUtils.scala
+++ b/internal/zinc-classfile/src/main/scala/sbt/internal/inc/JarUtils.scala
@@ -102,7 +102,7 @@ object JarUtils {
    * that changes between compilation runs (incremental compilation cycles).
    * Caching may hide those changes and lead into incorrect results.
    */
-  val scalacOptions = Set("-YdisableFlatCpCaching")
+  val scalacOptions: Set[ClassFilePath] = Set("-YdisableFlatCpCaching")
 
   /**
    * Options that have to be specified when running javac in order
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/633/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

jvican (approved):
LGTM

eed3si9n (approved):

niktrop:
@lukaszwawrzyk Does it handle change in other direction?

lukaszwawrzyk:
@niktrop Yes, the other direction was working before. Files that refer to jars just won't be found.

lukaszwawrzyk:
@jvican @eed3si9n Could you make a M2 release with this PR? It would help with testing and allow merging of https://github.com/JetBrains/intellij-scala/pull/453
Thanks!

eed3si9n:
Sounds good.

eed3si9n:
Zinc 1.3.0-M2 is on its way to Maven Central.

PR #652, opened 2019-04-09
Make Zinc work with Metals

jvican:
1. Disables project generation for bridge templates.
</s>2. Fixes a formatting change that happens when using latest scalafmt.</s>

A side-effect of disabling project generation for bridge templates is that the
bloop plugin must be added to the build.
  
EDIT: The formatting change was due to a wrong rebase on top of latest develop branch.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -472,6 +472,21 @@ def noSourcesForTemplate: Seq[Setting[_]] = inBoth(
   },
 )
 
+val disableBloop: Seq[Setting[_]] = List(
+  bloopGenerate in Compile := {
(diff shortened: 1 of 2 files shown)


eed3si9n:
I am not a huge fan of committing dev environment specific things into the repo.
Can Metals somehow express the same information as `project/LocalPlugin.scala` or something like that we can gitignore, or ideally a configuration file that doesn't affect the build?

jvican:
Let me try tomorrow to have another workaround that doesn't require bloop. I've tried several approaches with any luck in any of them so far, so even though I'm not fan of this either I'm not sure I'll succeed in finding another solution.

Issue #598, opened 2018-09-18
Changing abstract class to trait causes "invalid superClass"


pauldraper:
Using sbt 1.2.3, Scala 2.12.6

```sh
git clone git@github.com:typelevel/cats.git
cd cats
sbt compile
sed -i 's/abstract class StateTFunctions/trait StateTFunctions/' core/src/main/scala/cats/data/IndexedStateT.scala
sbt compile
```

```
[info] Compiling 1 Scala source to /home/paul/dev/typelevel/cats/core/.jvm/target/scala-2.12/classes ...
java.lang.AssertionError: assertion failed: Invalid superClass in Lcats/data/package$StateT$;: Some(Lcats/data/StateTFunctions;)
	at scala.tools.nsc.backend.jvm.BTypes$ClassBType.scala$tools$nsc$backend$jvm$BTypes$ClassBType$$checkInfoConsistency(BTypes.scala:649)
	at scala.tools.nsc.backend.jvm.BTypes$ClassBType$.apply(BTypes.scala:817)
	at scala.tools.nsc.backend.jvm.BTypesFromSymbols.classBTypeFromSymbol(BTypesFromSymbols.scala:104)
	at scala.tools.nsc.backend.jvm.BTypesFromSymbols.$anonfun$typeToBType$1(BTypesFromSymbols.scala:158)
	at scala.collection.MapLike.getOrElse(MapLike.scala:128)
	at scala.collection.MapLike.getOrElse$(MapLike.scala:126)
	at scala.collection.AbstractMap.getOrElse(Map.scala:59)
	at scala.tools.nsc.backend.jvm.BTypesFromSymbols.primitiveOrClassToBType$1(BTypesFromSymbols.scala:158)
	at scala.tools.nsc.backend.jvm.BTypesFromSymbols.typeToBType(BTypesFromSymbols.scala:173)
	at scala.tools.nsc.backend.jvm.BCodeSkelBuilder$PlainSkelBuilder.tpeTK(BCodeSkelBuilder.scala:80)
	at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genLoad(BCodeBodyBuilder.scala:249)
	at scala.tools.nsc.backend.jvm.BCodeBodyBuilder$PlainBodyBuilder.genLoadQualifier(BCodeBodyBuilder.scala:900)
[...]

PR #654, opened 2019-04-11
Workaround for 'symbol is missing from the classpath' error

lukaszwawrzyk:
We faced an odd problem while renaming a type. In case of the test it is Before -> After. The problem is in `Problem` class. It is not selected during initial invalidation, but without that happening the 3 other classes won't compile at all which will prevent zinc from building the Problem class in the next cycle. Without ProblemUser class it would work properly.
It seems like there is no reasonable way to detect that `Problem` needs to be recompiled without running the compilation (or usually including too many sources to recompile).

The workaround is to just detect this specific compile error:
```
[error] /tmp/sbt_dbac1806/ProblemUser.scala:5:13: Symbol 'type <empty>.Before' is missing from the classpath.
[error] This symbol is required by 'method Problem.x'.
[error] Make sure that type Before is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
[error] A full rebuild may help if 'Problem.class' was compiled against an incompatible version of <empty>.
[error]     println(p.x)
[error]             ^
```
and if it happens invalidate all sources. I can see how hacky it is, so if you have any suggestions to improve that... 
I also know that we don't need to invalidate all sources, but we could just use the analysis and parse the error to figure out what is actually relevant but I rather went with a simple solution and as the error is quite rare it shouldn't be a distaster to recompile the whole module when this happens.

### DIFF
--- a/zinc/src/main/scala/sbt/internal/inc/IncrementalCompilerImpl.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/IncrementalCompilerImpl.scala
@@ -331,18 +331,31 @@ class IncrementalCompilerImpl extends IncrementalCompiler {
         Incremental.prune(srcsSet, previousAnalysis, output, outputJarContent)
     }
 
-    // Run the incremental compilation
-    val compile = IncrementalCompile(
-      srcsSet,
-      lookup,
-      mixedCompiler.compile,
-      analysis,
-      output,
-      log,
-      incOptions,
-      outputJarContent
-    )
-    compile.swap
+    def runIncrementalCompilation(analysis: CompileAnalysis) = {
+      val compile = IncrementalCompile(
+        srcsSet,
+        lookup,
+        mixedCompiler.compile,
+        analysis,
+        output,
+        log,
+        incOptions,
+        outputJarContent
+      )
+      compile.swap
+    }
+
+    def recoverableFailure(e: CompileFailed): Boolean = {
+      e.problems.exists(_.message.contains("A full rebuild may help"))
+    }
+
+    try {
+      runIncrementalCompilation(analysis)
+    } catch {
+      case e: CompileFailed if recoverableFailure(e) =>
+        mixedCompiler.config.reporter.reset()
+        runIncrementalCompilation(Analysis.empty)
+    }
   }
 
   private def compileToJarSwitchedOn(config: CompileConfiguration): Boolean = {
--- a/zinc/src/sbt-test/source-dependencies/no-type-annotation/Before.scala
+++ b/zinc/src/sbt-test/source-dependencies/no-type-annotation/Before.scala
@@ -1,0 +1,1 @@
+class Before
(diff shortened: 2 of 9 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/654/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

zinc/src/main/scala/sbt/internal/inc/IncrementalCompilerImpl.scala
```diff
@@ -331,18 +331,31 @@ class IncrementalCompilerImpl extends IncrementalCompiler {
...
+    def recoverableFailure(e: CompileFailed): Boolean = {
+      e.problems.exists(_.message.contains("A full rebuild may help"))
+    }
+
+    try {
+      runIncrementalCompilation(analysis)
+    } catch {
+      case e: CompileFailed if recoverableFailure(e) =>
+        mixedCompiler.config.reporter.reset()
+        runIncrementalCompilation(Analysis.empty)
```
eed3si9n, line 357: I am not sure if this is a good idea. If we have false positive on the "A full rebuild may help" for whatever reason, we'd run into an infinite loop. Instead we should try to fix the invalidation for your use case.

lukaszwawrzyk: There shoun't be an infinite loop. This is before we are `cycling`, so if compilation fails once with this error, it will be retried only once.
I'd prefer to fix the invalidation, but to be honest I don't know how. I was thinking about the actual problem for a while and it sounds like not possible with how zinc is designed. We can't detect that prior to initial compilation, which will fail. Extending the heuristics to detect that prior to the compilation would usually  cause too many recompilations. I don't see any other way to fix this.

jvican (changes requested):
I feel pretty strongly against this change. This is a hack and if someone wants to do it it should not be done by Zinc but by the build tool.

eed3si9n:
For reference, the "A full rebuild may help" error message is coming from scalac.

https://github.com/scala/scala/blob/v2.12.8/src/reflect/scala/reflect/internal/pickling/UnPickler.scala#L265-L269

lukaszwawrzyk:
Yeah I know this is from scalac. Nevertheless, given that you have more experience in zinc, do you see a way to make the test pass without dirty hacks?

smarter:
At the very least, this should be done by changing scalac to report the error in a way that the compiler bridge can unambiguously interpret without parsing strings, though I can't think of a way to do it given that scalac doesn't give unique IDs to errors like dotty does.

smarter:
I think maybe the heuristic to use here is "if among the files selected for invalidation there exists two files A and B and among the files _not_ selected for invalidation there exists a file X such that A depends on X and X depends on B, then X should be added to the set of files selected for invalidation".

smarter:
In fact, if we generalize this heuristic to allow A and B to be the same file, I think this would also solve https://github.com/sbt/zinc/issues/598

eed3si9n:
I am closing this for now.

PR #588, opened 2018-08-28
Follow-up to the fields added in #580, and add Problem#rendered

smarter:
This PR requires a release of sbt util with https://github.com/sbt/util/pull/175 merged.

After having a look at https://discuss.lightbend.com/t/patching-strategy/1924 I have no idea whether I'm supposed to open things against 1.2.x or 1.x, so I'm opening against 1.2.x since that's what I care about right now.

PR #699, opened 2019-09-22
Fix sbt/sbt#5058: error messages coming from Dotty are incorrectly reported

smarter:
One year ago in https://github.com/sbt/zinc/pull/588, I wrote:
> After having a look at https://discuss.lightbend.com/t/patching-strategy/1924 I have no idea whether I'm supposed to open things against 1.2.x or 1.x, so I'm opening against 1.2.x since that's what I care about right now.

Afterwards this PR was merged and then... it was never forward-ported to the develop branch. This reveals a deep flaw in the development model of sbt, long-living branches are only an acceptable practice if there's some process in place to make sure that they stay synchronized in some way, this is usually achieved by periodically merging one branch into another (example: https://github.com/scala/scala/pull/8405). Alternatively, just give up on long-lived branches and have shorter, frequent releases from master.

Anyway, this PR forward-ports https://github.com/sbt/zinc/pull/588 to 1.3.x. Once this is merged I will also forward-port it to develop since I apparently can't trust any process to do that for me.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -303,9 +303,15 @@ lazy val zincPersist = (project in internalPath / "zinc-persist")
         exclude[DirectMissingMethodProblem]("sbt.internal.inc.schema.Problem#ProblemLens.rendered"),
         exclude[MissingClassProblem]("sbt.internal.inc.text.Java678Encoder"),
         // Added {start,end}{Offset,Line,Column}
+        // Added Position#{start,end}{Offset,Line,Column}
         exclude[DirectMissingMethodProblem]("sbt.internal.inc.schema.Position.apply"),
         exclude[DirectMissingMethodProblem]("sbt.internal.inc.schema.Position.copy"),
         exclude[DirectMissingMethodProblem]("sbt.internal.inc.schema.Position.this"),
+
+        // Added Problem#reported
+        exclude[DirectMissingMethodProblem]("sbt.internal.inc.schema.Problem.apply"),
+        exclude[DirectMissingMethodProblem]("sbt.internal.inc.schema.Problem.copy"),
+        exclude[DirectMissingMethodProblem]("sbt.internal.inc.schema.Problem.this"),
       )
     }
   )
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/ProblemStringFormats.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/ProblemStringFormats.scala
@@ -24,20 +24,21 @@ import sbt.util.InterfaceUtil.jo2o
 trait ProblemStringFormats {
   implicit lazy val ProblemStringFormat: ShowLines[Problem] = new ShowLines[Problem] {
     def showLines(p: Problem): Seq[String] =
-      p match {
-        case p if !p.position.sourcePath.isPresent && !p.position.line.isPresent =>
-          Vector(p.message)
-        case _ =>
-          val pos = p.position
-          val sourcePrefix = jo2o(pos.sourcePath).getOrElse("")
-          val columnNumber = jo2o(pos.pointer).fold(1)(_.toInt + 1)
-          val lineNumberString = jo2o(pos.line).fold(":")(":" + _ + ":" + columnNumber + ":") + " "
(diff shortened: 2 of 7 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.3.x&zinc=pull/699/head&io=1.3.x&librarymanagement=1.3.x&util=1.3.x&website=1.3.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
It is a bummer that process wasn’t clear a year ago. For 1.3.x, we have been making all prs against the main development branch and then backporting them to 1.3.x if appropriate. Merging doesn’t really make sense from develop -> 1.3.x, but periodically merging 1.3.x back into develop is also a good practice.

The default branch is also origin/develop and I’m not sure if that was the case a year ago. Hopefully this kind of thing will be less likely to slip through the cracks with the current development flow.

eed3si9n:
The discussion thread https://discuss.lightbend.com/t/patching-strategy/1924 concluded with

> we first merge everything to current 1.x branch, and then manually cherry pick to 1.2.x.

Soon afterwards we renamed `1.x` to `develop`, and the process has not changed from 1 year ago. I think the mistake was that we first merged into 1.2.x branch even though the process says otherwise. Let's _not_ repeat the same mistake. We should target this PR to `develop` branch, and then cherry pick to 1.3.x afterwards.

smarter:
> We should target this PR to develop branch, and then cherry pick to 1.3.x afterwards.

OK, I've rebased the branch, but I find that unfortunate since there's no relationship between the commits, whereas if we pushed to 1.3.x and then merged into develop, the same commits would be in both.

eatkins:
> I find that unfortunate since there's no relationship between the commits, whereas if we pushed to 1.3.x and then merged into develop, the same commits would be in both.

There's definitely merit to doing it that way. One question I'd have with that process is when do you merge 1.3.x into develop? After every merge to 1.3.x? My guess is that no one is really dogfooding the 1.3.x branch on a regular basis so I'd be somewhat concerned that if we didn't merge back after every commit that we wouldn't notice regressions that aren't covered by scripted. Granted, few people are dogfooding the develop branch so I'm not sure this is a huge issue.

I think both approaches will also confuse some contributors trying to figure out against which branch they should open their PR. Overall, I think preserving the status quo is the easiest way forward. Perhaps we should add a PR template similar to issues where it explicitly explains under which circumstances each branch should be used.

smarter:
> One question I'd have with that process is when do you merge 1.3.x into develop? After every merge to 1.3.x?

I'd say every month or so, I think that's how scalac proceeds: https://github.com/scala/scala/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+merge+2.12.x+to+2.13.x

> Granted, few people are dogfooding the develop branch so I'm not sure this is a huge issue.

Dogfooding of sbt in general is a big problem, nightlies are often broken, and the official repositories (sbt, zinc, ...) are rarely using the latest versions, when I think they should be using nightlies to catch regressions early.

eed3si9n:
> OK, I've rebased the branch, but I find that unfortunate since there's no relationship between the commits, whereas if we pushed to 1.3.x and then merged into develop, the same commits would be in both.

That was the whole debate on https://discuss.lightbend.com/t/patching-strategy/1924 basically.
I agree that "Merge to Stable branch (for example 1.2.x) first" is in some way more correct approach. The downside is that it needs prior discussion on where the PR should target and also it lets the stable branch live longer.

If we merged everything in `develop` and released regularly, there's less need for backporting, so it's a bit chicken-egg problem.

eed3si9n (approved):

SethTisue:
> they should be using nightlies to catch regressions early

(perhaps the Scala community build could/should play a role here? it caught a number of regressions in 1.3.x before release)

eed3si9n:
@SethTisue Yes. We've discussed it in the past, but I'll make a GitHub issue on that to make that more concrete.

jvican (approved):
LGTM :+1:

PR #733, opened 2020-01-28
 Improve error message for non-absolute paths

x3ro:
I just ran into this while I made a boo-boo in my SBT configuration, so I've tried to improve the error message to make it more clear. Feedback welcome :)

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
@@ -720,11 +720,17 @@ object IncrementalCommon {
       addedSources.filterNot(_.isAbsolute).toList match {
         case first :: more =>
           val fileStrings = more match {
-            case Nil      => first.toString
-            case x :: Nil => s"$first and $x"
-            case _        => s"$first and ${more.size} others"
+            case Nil      => s"${first.toString} was"
+            case x :: Nil => s"$first and $x were"
+            case _        => s"$first and ${more.size} others were"
           }
-          sys.error(s"Expected absolute source files instead of ${fileStrings}.")
+
+          val errorMsg =
+            s"""File paths fed to the compiler are required to be absolute, but
+              |${fileStrings} not. Please check whether you've configured your
+              |build system correctly.""".stripMargin
+
+          sys.error(errorMsg)
         case Nil => ()
       }
     }


lightbend-cla-validator:
Hi @x3ro,

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/733/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
@x3ro Could you show some sample error message before and after?

x3ro:
Sure! Before:

> Expected absolute source files instead of rest/src/main/scala/Foo.scala and rest/src/main/Bar.scala

After:

> File paths fed to the compiler are required to be absolute, but
> rest/src/main/scala/Foo.scala and rest/src/main/Bar.scala were not.
> Please check whether you've configured your build system correctly.

eed3si9n:
I am open to error message improvements but I am not sure about the details of the proposal.

> File paths fed to the compiler are required to be absolute, but

If we make distinction between Zinc and Scala Compiler, the compiler actually can support non-File inputs. Future version of Zinc will support that.

> Please check whether you've configured your build system correctly.

I'd avoid "please" in error messages for the sake of consistency in tone. Also it's not clear what is meant by checking the configuration of build system.

Maybe a simpler improvement of wording might be:

```
Absolute paths were expected but found relative paths instead: rest/src/main/scala/Foo.scala and rest/src/main/Bar.scala
```

What do you think?

x3ro:
Mmh I see your point. The specific reason I ended up making this PR is because I had trouble understanding where the problem in the build was, and I assumed it was some problem with the files themselves, which was made harder by the fact that I had no idea what zinc is at the time (as is the case with most people using SBT, I'd assume). That is why I ended up mentioning the build system.

Long story short, it would've helped me to somehow explicitly indicate that this problem is caused before actual compilation of those source files ever started, but from your comment I'm guessing that e.g.

> This can, for example, be caused by a misconfiguration in your build system.

is also not an option? 🤔

eed3si9n:
btw which setting caused the problem? Maybe we can put the check in sbt itself?

dwijnand:
"Please check whether you've configured your build system correctly" does provide a good hint that the problem isn't in the files themselves, but it's so vague and un-actionable that I feel it's only slightly shifting the problem and not making it easier to deal with.

Can you think of a better message to use?  Otherwise the status quo looks good.

PR #758, opened 2020-04-21
Avoid slow path for associated file more often

### LINKED ISSUE #757: Further issues with associatedFile
- Scalac does not set `associatedFile` until  a class symbol's info is forced. It is possible to get a reference to type after typechecking that didn't actually need to force it. I've see this in a tree type of `x.getClass.getName`. The `Select(Ident(x), TermName(getClass))` tree had a type `Class[_ <: SomeClass]` where the class symbol for `SomeClass` was unitialized. I'm working on a reproducer, there is some special logic in type for `getClass` so maybe thats irregular?
   - I could set up scalac to return the correct `associatedFile` even before the info is forced.
 - When Zinc traverses the `original` of ` TypeTree` we an have an `Ident` with a `tpe` assigned but _no_ `.symbol` assigned. Again, I'm working on a reproducer, this might be related to case class accessor synthesis. This searches the classpath for the FQN of `<none>`.
 - .. which it incorrectly computes as `<none>xxxxxxxx`, where the `xxx` is whatever characters are adjacent to `<none>` in the name table. This is due to a bug in `CallbackGlobal.fullName` which computes string lengths of `symName.encoded` but then discards the encoded name and takes data from the name table starting at `symName.start`.
 - Other special symbols like `<byname>` / `<repeated>` and `scala.Nothing` are also looked for on the classpath. This could be fixed by a) first checking for `definitions.isPhantomClass | defintions.syntheticCoreClasses`; b) caching negative lookups or c) both.

retronym:
Also fix a bug in fullName with encoded names.

Fixes #757

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/CallbackGlobal.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/CallbackGlobal.scala
@@ -185,6 +185,7 @@ sealed class ZincCompiler(settings: Settings, dreporter: DelegatingReporter, out
       val newResult = findOnPreviousCompilationProducts(fqn)
         .map(f => (f, true))
         .orElse(findOnClassPath(fqn).map(f => (f, false)))
+
       newResult.foreach(res => fqnsToAssociatedFiles.put(fqn, res))
       newResult
     }
@@ -220,18 +221,18 @@ sealed class ZincCompiler(settings: Settings, dreporter: DelegatingReporter, out
     def loop(size: Int, sym: Symbol): Unit = {
       val symName = sym.name
       // Use of encoded to produce correct paths for names that have symbols
-      val encodedName = symName.encoded
+      val encodedName = symName.encode
       val nSize = encodedName.length - (if (symName.endsWith(nme.LOCAL_SUFFIX_STRING)) 1 else 0)
       if (sym.isRoot || sym.isRootPackage || sym == NoSymbol || sym.owner.isEffectiveRoot) {
         val capacity = size + nSize
         b = new java.lang.StringBuffer(capacity)
-        b.append(chrs, symName.start, nSize)
+        b.append(chrs, encodedName.start, nSize)
       } else {
         val next = if (sym.owner.isPackageObjectClass) sym.owner else sym.effectiveOwner.enclClass
         loop(size + nSize + 1, next)
         // Addition to normal `fullName` to produce correct names for nested non-local classes
         if (sym.isNestedClass) b.append(nme.MODULE_SUFFIX_STRING) else b.append(separator)
-        b.append(chrs, symName.start, nSize)
+        b.append(chrs, encodedName.start, nSize)
       }
       ()
     }
--- a/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
@@ -148,12 +148,16 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with 
       val targetSymbol = dep.to
       val onSource = targetSymbol.sourceFile
       if (onSource == null) {
-        // Ignore packages right away as they don't map to a class file/jar
-        if (targetSymbol.hasFlag(scala.tools.nsc.symtab.Flags.PACKAGE)) ()
-        // Ignore `Any` which by default has no `associatedFile`
-        else if (targetSymbol == definitions.AnyClass) ()
-        else {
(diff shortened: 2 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/758/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -149,11 +149,13 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
       val onSource = targetSymbol.sourceFile
       if (onSource == null) {
         // Ignore packages right away as they don't map to a class file/jar
-        if (targetSymbol.hasFlag(scala.tools.nsc.symtab.Flags.PACKAGE)) ()
-        // Ignore `Any` which by default has no `associatedFile`
-        else if (targetSymbol == definitions.AnyClass) ()
+        if (targetSymbol.hasFlag(scala.tools.nsc.symtab.Flags.PACKAGE) ()
```
dwijnand:
```suggestion
        if (targetSymbol.hasFlag(scala.tools.nsc.symtab.Flags.PACKAGE)) ()
```

```
[error] /home/travis/build/sbt/zinc/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala:154:9: ')' expected but 'else' found.
[error]         else if (targetSymbol == NoSymbol) ()
[error]         ^
[error] one error found
```

jvican (approved):
LGTM 👍 good catch!

internal/compiler-bridge/src/main/scala/xsbt/CallbackGlobal.scala
```diff
@@ -225,13 +226,13 @@ sealed class ZincCompiler(settings: Settings, dreporter: DelegatingReporter, out
       if (sym.isRoot || sym.isRootPackage || sym == NoSymbol || sym.owner.isEffectiveRoot) {
         val capacity = size + nSize
         b = new java.lang.StringBuffer(capacity)
-        b.append(chrs, symName.start, nSize)
+        b.append(chrs, encodedName.start, nSize)
```
dwijnand, line 229:
```[error] /home/travis/build/sbt/zinc/internal/compiler-bridge/src/main/scala/xsbt/CallbackGlobal.scala:229:24: type mismatch;
[error]  found   : encodedName.type (with underlying type String)
[error]  required: ?{def start: ?}
[error] Note that implicit conversions are not applicable because they are ambiguous:
[error]  both method stringToTermName in trait Names of type (s: String)ZincCompiler.this.TermName
[error]  and method stringToTypeName in trait Names of type (s: String)ZincCompiler.this.TypeName
[error]  are possible conversion functions from encodedName.type to ?{def start: ?}
[error]         b.append(chrs, encodedName.start, nSize)
[error]                        ^
```

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -150,10 +150,12 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
       if (onSource == null) {
         // Ignore packages right away as they don't map to a class file/jar
         if (targetSymbol.hasFlag(scala.tools.nsc.symtab.Flags.PACKAGE)) ()
-        // Ignore `Any` which by default has no `associatedFile`
-        else if (targetSymbol == definitions.AnyClass) ()
+        // Seen in the wild: an Ident as the original of a TypeTree from a synthetic case accessor was symbol-less
+        else if (targetSymbol == NoSymbol) ()
+        // Also ignore magic symbols that don't have bytecode like Any/Nothing/Singleton/<byname>/<repeated>/...
+        else if (definitions.syntheticCoreClasses(targetSymbol) || definitions.isPhantomClass(targetSymbol)) ()
```
dwijnand:
```
[error] /home/travis/build/sbt/zinc/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala:156:51: type mismatch;
[error]  found   : Dependency.this.global.Symbol
[error]  required: Int
[error]         else if (definitions.syntheticCoreClasses(targetSymbol) || definitions.isPhantomClass(targetSymbol)) ()
[error]                                                   ^
```

PR #764, opened 2020-04-28
Memory efficiency for used names collection and other Relations building

retronym:
- Build the maps in the used name relation in a mutable map,
    then use a HashMap builder in a second pass to build the
    resulting maps. In Scala 2.12.12+ or 2.13.0+, use of the
    builder is fastest as the builder mutates the structure
    internally.
  - Intern the keys and values, so we don't have a separate
    instance of `UsedName("toString", EnumSet.of(UseScope.Default))`
    for each call site.
 - Factor this into `RelationBuilder,` and use this in
   `ProtobufReaders` where I recently made similar opts.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/764/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/RelationBuilder.scala
```diff
@@ -0,0 +1,51 @@
...
+        update(a, b)
+      }
+    }
+  }
+
+  def result(): Relation[A, B] = {
+    def toImmutable[K, V](
+        map: java.util.HashMap[K, (K, mutable.Builder[V, immutable.HashSet[V]])]
+    ): immutable.HashMap[K, immutable.HashSet[V]] = {
+      val builder = immutable.HashMap.newBuilder[K, immutable.HashSet[V]]
```
retronym, line 45: https://github.com/scala/scala/pull/8726 is the 2.12.x backport of efficient `immutable.HashMap$Builder` (thanks, @mkeskells!). With that, or with 2.13, this step is step is quite efficient.

I considered an alternative of just doing an `unsafeToImmutableMap` wrapper, but that is a somewhat more fragile if the users of the built `Relation` start to call map operations on it. 

It would be great if we had a richer `immutable.HashMap.Builder` that supported key `getOrElseUpdate` on the map being built, as well as a find `transformValues` step. Then, we could build up the desired structure from the begining and do a second pass that just mutates the data in the leaves.

internal/zinc-core/src/main/scala/sbt/internal/inc/Relations.scala
```diff
@@ -636,17 +636,18 @@ private class MRelationsNameHashing(
...
+  private[inc] def addUsedNames(data: Relation[String, UsedName]): Relations = {
     new MRelationsNameHashing(
       srcProd,
       libraryDep,
       libraryClassName,
       internalDependencies = internalDependencies,
       externalDependencies = externalDependencies,
       classes,
-      names = names + (className, name),
+      names = if (names.forwardMap.isEmpty) data else names ++ data,
```
retronym, line 574: This optimization could be pushed down into `sbt-util`.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCompile.scala
```diff
@@ -367,8 +367,11 @@ private final class AnalysisCallback(
     ()
   }
 
-  def usedName(className: String, name: String, useScopes: util.EnumSet[UseScope]) =
-    add(usedNames, className, UsedName(name, useScopes))
+  def usedName(className: String, name: String, useScopes: util.EnumSet[UseScope]) = {
+    usedNames.synchronized {
```
retronym: synchronized for hydra.

retronym:
@jvican @eed3si9n I tried to reinstate the benchmark subproject (which was commented out in the change to include 2.13.x in cross building). It compiles and runs but I get errors when running JMH. Any ideas?

jvican:
Did you upgrade jmh and update the scala version used in the benchmark source?

retronym:
No, I just did https://github.com/sbt/zinc/pull/764/commits/e47b500ec646815f49500046530b90c9e6052664

eed3si9n:
I'm working on getting the benchmark back online - https://github.com/sbt/zinc/pull/767

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCompile.scala
```diff
@@ -381,12 +384,9 @@ private final class AnalysisCallback(
...
   def addCompilation(base: Analysis): Analysis =
     base.copy(compilations = base.compilations.add(compilation))
-  def addUsedNames(base: Analysis): Analysis = usedNames.foldLeft(base) {
-    case (a, (className, names)) =>
-      import scala.collection.JavaConverters._
-      names.asScala.foldLeft(a) {
-        case (a, name) => a.copy(relations = a.relations.addUsedName(className, name))
-      }
+  def addUsedNames(base: Analysis): Analysis = {
+    assert(base.relations.names.size == 0)
```
dwijnand: Why?  And should we keep it?

dwijnand (approved):
A light-touch/non-paranoid LGTM.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCompile.scala
retronym: I added a fast path to `Relations.addUsedNames` when `this.names` is empty, as is always the case in the way Zinc is implemented. I want to test that assumption here. Maybe its worth keeping to defend the assumption.

internal/zinc-core/src/main/scala/sbt/internal/inc/Relations.scala
retronym: ... although that repo is now archived so I guess `Zinc` will end up with a copy of a subset of its data structues to avoid a circular dependency. @eed3si9n what's the plan there?

retronym:
Rebased on #767 so I can do before/after benchmarking, and marking as a draft until that's done.

retronym:
Note to self: the benchmarks are unlikely to show much benefit on 2.12.11. Run on 2.12.x-SNAPSHOT instead.

In that build:
  - ambient allocations are reduced by sharing the buffer for classfile reading, so we can better focus on the changes in this PR.
  - `immutable.HashMap.newBuilder` is way more efficient.

dotta:
@retronym Hey! :)

I'm concerned that with this change we may experience concurrency hazards when compiling with Hydra (such as visibility issues), as the underlying `AnalysisCallback` instance is shared between the different compilation threads. If you agree, could we find a way to make this change compatible with Hydra?

\cc @dragos @eed3si9n

retronym:
Yep, that's totally fine with me. Looking at this again, I see I did synchronizes writes to the `usedName` `RelationBuilder`, but I don't have a synchronized in the final read in `def addUsedNames` so updates from the threads might not be visible.

retronym:
You might want a different performance tradeoff (preferring less potential for contention and worrying less about allocation pressure). In that case you could create a second implementation of `RelationBuilder` that is more like the original code (`TrieMap`-based) and enable that in Hydra.

dotta:
Yep, that's it. Problem is that swapping the `AnalysisCallback` from the outside (i.e., from a sbt plugin) isn't really possible at the moment (and we should do it in a way that works within the IntelliJ scala-plugin, as it leverages zinc and we have a Hydra integration in IntelliJ too). This doesn't look simple to achieve.

As the changes in this PR are motivated by (allocation) performance, what's the order of magnitude of the savings obtained using the current implementation? (I suspect you have been seeing a meaningful difference in large code bases, haven't you? If so, could you share some figures) By the way, would you have a benchmark that we can use to see if we can tweak the implementation in a way that would make it work with Hydra too, and at no meaningful penalty for vanilla Scala?

dotta:
@retronym ping :)

retronym:
@dotta We've now got a way to avoid building relations (bi-directional maps) for used names. So we can go back to the lock-free data structures: https://github.com/sbt/zinc/pull/995/files#r677157060

PR #785, opened 2020-05-18
Push lookupAnalyzedClass into Lookup

dwijnand:
... allowing users to optimize it - in both directions: in finding the
class or in not finding the class (i.e. without having to
lookupAnalysis).

### DIFF
--- a/zinc/src/main/scala/sbt/internal/inc/LookupImpl.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/LookupImpl.scala
@@ -13,6 +13,7 @@ package sbt.internal.inc
 
 import java.util.Optional
 
+import xsbti.api.AnalyzedClass
 import xsbti.compile.{ Changes, CompileAnalysis, FileHash, MiniSetup }
 import xsbti.{ VirtualFile, VirtualFileRef }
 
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/785/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

retronym (approved):

zinc/src/main/scala/sbt/internal/inc/LookupImpl.scala
```diff
@@ -55,6 +56,13 @@ class LookupImpl(compileConfiguration: CompileConfiguration, previousSetup: Opti
     .flatMap(ext => ext.getExternalLookup().toOption)
     .collect { case externalLookup: ExternalLookup => externalLookup }
 
+  override def lookupAnalyzedClass(binaryClassName: String): Option[AnalyzedClass] = {
+    externalLookup match { // not flatMap so that external lookup can fast-track returning None
```
eed3si9n, line 60: I'm trying to integrate this change into sbt, and I'm a bit confused. What does it mean for `externalLookup.lookupAnalyzedClass(binaryClassName)` to return `None` here?
sbt already implements `externalLookup`, so I would like a way to say I don't want to participate in AnalyzedClass lookup, but if I return `None` it would be interpreted as "I couldn't find the AnalyzedClass"?

dwijnand: Yeah, the current setup has no way for `externalLookup.lookupAnalyzedClass` to say "I don't know, you try".  You'd have to modify this if you need that.

I'm not sure why `Some(APIs.emptyAnalyzedClass)` wasn't used for fast-track's "I got you, here's the API for that class name".

PR #786, opened 2020-05-19
Introduce "provenance" to AnalyzedClass

dwijnand:
Adding a "provenance" field to AnalyzedClass allowing "lookupAnalyzedClass" to be fast-tracked.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/786/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/compiler-interface/src/main/contraband/other.json
```diff
@@ -17,6 +17,11 @@
           "type": "int",
           "since": "1.2.0",
           "default": "apiHash"
+        },
+        { "name": "provenance",
```
eed3si9n: Could we add documentation on what this is, and how it's intended to be used from the build tools? It might be good to consider more descriptive name for long-term maintenance, like `remoteCacheId`, which is the term I'm using for https://github.com/sbt/sbt/pull/5534.

eed3si9n (approved):

PR #790, opened 2020-06-02
Convert VirtualFile to PlainFile in the compiler bridge

### LINKED ISSUE #770: [1.4.x] Presenting source files as VirtualFileWrap prevents memory efficient code path in scalac's SourceReader
https://github.com/sbt/zinc/blob/db028c338af528876e24fc8d6214e24f7d19d707/internal/compiler-bridge/src/main/scala/xsbt/CompilerInterface.scala#L156-L160

Seems to stop use getting to the efficient first line in:

https://github.com/scala/scala/blob/2826c6cb96f948cbe1c90ca73e43c8aa399fa1a9/src/compiler/scala/tools/nsc/io/SourceReader.scala#L58-L63

Which leads to allocation profiles like:

image

### LINKED ISSUE #784: DelegationReporter creates stale File objects
The `DelegatingReporter` converts `scala.reflect.api.Position` to `xsbti.Position` in the `makePosition` method.

When compiling inside sbt, the `scala.reflect.api.Position` contains a path string that derives from a `VirtualFileRef`. This path cannot be used as is to create the actual File in `xsbti.Position`.

It might be possible to use a `FileConverter` to convert this path to a `VirtualFile` and then to a valid system Path:
```scala
converter.toVirtualFile(new BasicVirtualFileRef(pos.path) {})).path
```
Or the `xsbti.Position` could return a `VirtualFileRef` instead of a `File` like the `CompileAnalysis` does.

### LINKED ISSUE #2053: The semanticdb plugin is incompatible with future Zinc 1.4.0
The introduction of `VirtualFile` in Zinc [#712](https://github.com/sbt/zinc/pull/712) impacts the semanticdb plugin.

Zinc is now using an `xsbti.VirtualFile` interface to abstract out all absolute paths. When invoking the compiler it wraps the instances of `xsbti.VirtualFile` inside an implementation of  `scala.reflect.io.AbstractFile` called `VirtualFileWrap`.

This prevents scalameta to convert `SourceFile` to scalameta `Input` in `InputOps`:

https://github.com/scalameta/scalameta/blob/27e5795f7a51bbf2ac68251ce32e5148033efaf8/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/InputOps.scala#L31-L38

https://github.com/scalameta/scalameta/blob/27e5795f7a51bbf2ac68251ce32e5148033efaf8/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/InputOps.scala#L65-L87

This may be easier to fix when the [scala/compiler-interface](https://github.com/scala/compiler-interface) will be released. See [zinc/#661](https://github.com/sbt/zinc/pull/661) for more info about this.

adpi2:
Fixes #770 
Fixes #784
Fixes [#2053](https://github.com/scalameta/scalameta/issues/2053) in [scalameta](https://github.com/scalameta/scalameta)

Wrap a `xsbti.VirtualFile` inside a `scala.reflect.io.PlainFile` or a `scala.reflect.io.VirtualFile` depending on the PathBased nature of the file. I am not sure if it is a good idea to extend those `scala.reflect.io` classes, but it has several benefits:
- The compiler will see `PlainFile`s and it can optimize on that fact.
- The downstream tooling will see `PlainFile` as well and it can perform on that fact:
  - scalameta can publish the semanticdb files
  - the sbt reporter can print the compiler diagnostics
  - ...
- Zinc will be able to retrieve the underlying `VirtualFile` without having to maintain a `Map[AbstractFile, VirtualFile]`

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/CompilerInterface.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/CompilerInterface.scala
@@ -154,7 +154,7 @@ private final class CachedCompiler0(args: Array[String], output: Output, initial
       compiler.set(callback, underlyingReporter)
       val run = new compiler.ZincRun(compileProgress)
 
-      val wrappedFiles = sources.map(new VirtualFileWrap(_))
+      val wrappedFiles = sources.map(VirtualFileWrap(_))
       val sortedSourceFiles: List[AbstractFile] =
         wrappedFiles.sortWith(_.underlying.id < _.underlying.id)
       run.compileFiles(sortedSourceFiles)
--- a/internal/compiler-bridge/src/main/scala/xsbt/DelegatingReporter.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/DelegatingReporter.scala
@@ -89,7 +89,9 @@ private object DelegatingReporter {
 
     def makePosition(pos: Position): xsbti.Position = {
       val src = pos.source
-      val sourcePath = src.file.path
+      val sourcePath = src.file match {
+        case VirtualFileWrap(virtualFile) => virtualFile.id
+      }
       val sourceFile = new File(src.file.path)
       val line = pos.line
       val lineContent = pos.lineContent.stripLineEnd
--- a/internal/compiler-bridge/src/main/scala/xsbt/ScaladocInterface.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/ScaladocInterface.scala
@@ -64,7 +64,7 @@ private class Runner(
       def document(ignore: Seq[String]): Unit = {
         import compiler._
         val run = new Run
-        val wrappedFiles = sources.toList.map(new VirtualFileWrap(_))
+        val wrappedFiles = sources.toList.map(VirtualFileWrap(_))
         val sortedSourceFiles: List[AbstractFile] =
           wrappedFiles.sortWith(_.underlying.id < _.underlying.id)
         run.compileFiles(sortedSourceFiles)
--- a/internal/compiler-bridge/src/main/scala/xsbt/VirtualFileWrap.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/VirtualFileWrap.scala
@@ -11,81 +11,44 @@
 
 package xsbt
 
-import xsbti.VirtualFile
-import scala.reflect.io.AbstractFile
-import java.io.{ File, InputStream, OutputStream }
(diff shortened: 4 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/790/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
I have no idea if it would be the case here, but there are a number of classes in scala reflect that are extremely slow to classload. If zinc was not touching those classes before and they do after this change, it could introduce a startup overhead of around a second for many tools. Based on the description, it sounds like the compiler is already using scala.reflect.io so it’s probably not a problem, but I figured it was worth mentioning.

dwijnand:
Yeah, scala-reflect is a facade for the compiler, and contains bits and bobs of things that are used by the compiler.  That aspect is unlikely to be adding any overhead.

dwijnand (approved):
> I am not sure if it is a good idea to extend those `scala.reflect.io` classes

I'm not sure either, but this looks like a good change to pursue.

eed3si9n:
Looks like there's a test failure:

```
[info] VirtualFileIncrementalSpec:
[info] incremental compiler
[info] - should allow virtual file *** FAILED ***
[info]   java.nio.file.Files.exists(expectedOut) was false /tmp/sbt_f38e1258/sub1/target/test/pkg/Depender$.class does not exist (VirtualFileIncrementalSpec.scala:121)
```

adpi2:
There is another test failure now in the AppVeyor check but I was not able to reproduce. It might be an non-deterministic error that is not linked to the changes.

```
[error] x source-dependencies / sealed
[error] sbt.internal.inc.BatchScriptRunner$PreciseScriptedError: Command failed: 'compile'
[error] Caused by: java.nio.file.FileSystemException: C:\Users\appveyor\AppData\Local\Temp\1\sbt_605036b9\target-bridge-2.12.11.jar: The process cannot access the file because it is being used by another process.
```
Does anyone know how to run the AppVeyor again?

eed3si9n:
> Does anyone know how to run the AppVeyor again?

I'm rerunning it now.

eed3si9n (approved):
The change LGTM. Thanks @adpi2!

PR #800, opened 2020-06-23
Add an assertion in MCompilations

dwijnand:
This... invariant(?).. is mentioned in the TODO.

In the pipelining PR I found it to be the root cause of an erroneous
implementation (merging multiple times).  So I think it would be good to
enforce this on construction, alas with an exception.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/800/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):
I'm not 100% sure if this is safe until we run community build against it, but I'd be happy to merge this for now.

PR #818, opened 2020-07-07
Trim scripted tests incOptions

dwijnand:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/818/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
If these only show up on error, isn't it better to keep these debug turned on so the error message (when it fails) provide useful information? i.e. what's the point of this PR?

dwijnand:
If that's the case, should it be on by default, do you think?

Meanwhile, I'll amend, keeping those lines.

PR #819, opened 2020-07-07
Merge DummyVirtualFile & MappedVirtualFile into PlainVirtualFile

dwijnand:
... by allowing its identifier to be specified, so it may serve 3 purposes.

Builds on #820


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/819/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/MappedVirtualFile.scala
```diff
@@ -13,57 +13,34 @@ package sbt
...
     rootPaths2.find { case (_, rootPath) => path.startsWith(rootPath) } match {
       case Some((key, rootPath)) =>
-        MappedVirtualFile(s"$${$key}/${rootPath.relativize(path)}".replace('\\', '/'), rootPaths)
+        PlainVirtualFile(s"$${$key}/${rootPath.relativize(path)}".replace('\\', '/'), path)
       case _ =>
         path.getFileName.toString match {
-          case "rt.jar"              => DummyVirtualFile("rt.jar", path)
-          case _ if allowMachinePath => MappedVirtualFile(s"$path".replace('\\', '/'), rootPaths)
+          case "rt.jar"              => PlainVirtualFile("rt.jar", path)
+          case _ if allowMachinePath => PlainVirtualFile(s"$path".replace('\\', '/'), path)
```
eed3si9n, line 45: Neither rt.jar nor mapped virtual file are "palin", so we're losing type information here. Also the motivation for MappedVirtualFile is to not hang on to the original Path object, which contains absolute path.

dwijnand: > Neither rt.jar nor mapped virtual file are "palin", so we're losing type information here.

Do you prefer the name "BasicVirtualFile"?  The point is it's a basic string id + Path implementation, where the id can be from the Path or from anything else.

> Also the motivation for MappedVirtualFile is to not hang on to the original Path object, which contains absolute path.

But we do have the original absolute path, it's just buried inside the `rootPathsMap` behind a key, which we need to recompute (and possibly re-allocate?).

To me, looking back at what the code was actually doing, it seemed like it was deriving an identifier for the key, based on `rootPaths`, which can be used as its id, which is persisted.  But if the `PathBasedFile` needs the `Path`, why recompute instead of just keeping a hold of the one we were given?

I'll need to look at Windows, though, which I suspect is falling prey to the random-looking `.replace('\\', '/')` usages and bare `PlainVirtualFile(_)` usages.

dwijnand: Turns out the Windows issue wasn't in any `PlainVirtualFile(_)`, but in `TestVirtualFile` _not_ replacing separators.  That's been fixed in the broken out PR, inside `BasicVirtualFileRef`'s constructor, and that's now merged.

So I could make `BasicVirtualFile` a non-final class, much like `BasicVirtualFileRef` is, and then create trivial subclasses for `PlainVirtualFile` and `MappedVirtualFile`, and even `DummyVirtualFile`.

eed3si9n: > But if the PathBasedFile needs the Path, why recompute instead of just keeping a hold of the one we were given?

Because all the absolute paths adds up in memory. See for example WindowsPath (https://github.com/openjdk/jdk/blob/master/src/java.base/windows/classes/sun/nio/fs/WindowsPath.java). When a build has 1000s of files both source and `*.class` files, we're keeping all the `C:\whatever\something\` portions in-memory. 

(The situation in Linux isn't super either since every now and then people with encrypted drive would come out and say that they're hitting the 255 character default. Scala 2.13.0 adopted 240 in https://github.com/scala/scala/pull/7497?)

This is relevant because to do subproject-dependency we basically need to keep the Analysis around for all subprojects. I should note however that my time-space-tradeoff claim here is a theoretical one. If you could, it would be cool to profile this in the large build and see if this makes actual difference.

dwijnand: Ah!  I see.  OK, I'll reconsider.  cc @retronym that might have more info or have given this some thought.

PR #820, opened 2020-07-08
Cleanup parts of the VirtualFile-related code

dwijnand:
* Rename VirtualFileWrap/XsbtPlainFile/XsbtVirtualFile to
  AbstractZincFile/ZincPlainFile/ZincVirtualFil;
* Default FileConverter#toVirtualFile(VirtualFileRef): VirtualFile
* Rework BasicVirtualDirectory & BasicMemoryFile to avoid needless
  Option and BasicVirtualDirectory boxing, & use structural sharing;
* Cleanup some of VirtualFile's docs.

Subset of #819


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/820/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/compiler-interface/src/main/java/xsbti/BasicVirtualFileRef.java
```diff
@@ -11,65 +11,48 @@
...
-        return name;
+        this.id = _id.replace('\\', '/');
+        int idx = id.lastIndexOf('/');
+        parent = idx == -1 ? "" : id.substring(0, idx + 1);
+        name = id.substring(idx + 1);
     }
 
     public String[] names() {
-        if (parent == null || parent == "") {
+        if (Objects.equals(parent, "")) {
```
eed3si9n, line 29:
```
scala> java.util.Objects.equals(null, "")
res0: Boolean = false
```

so wouldn't this return different result for `null`?

eed3si9n (changes requested):

internal/compiler-interface/src/main/java/xsbti/BasicVirtualFileRef.java
dwijnand: Yeah, but `parent` never is `null`, so I dropped that first.

But the thing here is equals vs `==`, in Java, where `==` means reference equality.  I _assume_/hope there's only 1 `""`, or at least for the literal `""`, within a JVM execution, but I wasn't sure so went with the safer option.

WDYT?

eed3si9n: You're right on this one.

```
eval new String("") eq ""
<eval>:1: warning: comparing a fresh object using `eq' will always yield false
new String("") eq ""
               ^
[info] ans: Boolean = false
```

eed3si9n (approved):

PR #828, opened 2020-07-10
Use the real 2.12 scala.reflect.io.PlainNioFile

dwijnand:
In 2.12 scala.reflect.io.PlainNioFile is private[scala] with no nioPath 
val, so first we need to define the alias in the scala package and then we
need to unwrap it with Java reflection... :sadface:

Builds on #827.

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
@@ -137,9 +137,9 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with 
               if (!ignore)
                 binaryDependency(zip.file.toPath, binaryClassName)
             }
-          case pf: xsbt.Compat.PlainNioFile =>
+          case pf: ZincCompat.PlainNioFile =>
             // The dependency comes from a class file
-            binaryDependency(pf.nioPath, binaryClassName)
+            binaryDependency(ZincCompat.unwrapPlainNioFile(pf), binaryClassName)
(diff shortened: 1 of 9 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/828/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

dwijnand:
Oh I need to rebase this.  OK, I'll do that after the chain is merged.

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -137,9 +137,9 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
               if (!ignore)
                 binaryDependency(zip.file.toPath, binaryClassName)
             }
-          case pf: xsbt.Compat.PlainNioFile =>
+          case pf: ZincCompat.PlainNioFile =>
```
dwijnand, line 140: Before this PR, in 2.12 whenever the compiler gives Zinc a `scala.reflect.io.PlainNioFile` Zinc checks if it's a `xsbt.PlainNioFile`, which it will never be, and fails to register it as a dependency.

That's what this PR is fixing.

eed3si9n (approved):

PR #836, opened 2020-07-14
Don't bomb if -Ypickle-java isn't present

dwijnand:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/836/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
```diff
@@ -100,8 +100,14 @@ final class MixedAnalyzingCompiler(
           val cp: Seq[VirtualFile] = (extraClasspath map { x =>
             config.converter.toVirtualFile(x.toAbsolutePath)
           }) ++ absClasspath
+          val scalacOpts = config.currentSetup.options.scalacOptions
+          val pickleJava = {
+            if (isPickleJava && !scalacOpts.contains("-Ypickle-java"))
+              List("-Ypickle-java")
+            else Nil
+          }
```
eed3si9n: I don't know if adding -Y flag forcefully would still make it a good default..

dwijnand: Hmm...  Yeah.  Just drop the assert, then?  WDYT?

eed3si9n: This would mess up pipelining. I think the solution is to actually solve the mixed Scala/Java compilation, and let the build tools configure the flags.

dwijnand: Why and how would it mess it up?  It's an assertion, so removing it can't actively mess anything up.

dwijnand: I don't disagree with eventually finding a solution for pipelining in mixed projects, but we can drop as it breaks your example.

eed3si9n: When a build tool (currently scripted) correctly hooks up the pipelined tasks, not passing `-Ypickle-java` would create mysterious error saying the symbols (defined in upstream Java subproject) are not found.

dwijnand: Yeah, but it's not an outlier: there's about a billion ways you can set up pipelining wrong.

dwijnand:
#843 disables pipelining but tests everything with it, which works for me.

PR #845, opened 2020-07-19
Make a number of performance optimizations

eatkins:
I noticed that running https://github.com/eatkins/scala-build-watch-performance with a local sbt build using zinc 1.4.0-M1 was slower than 1.3.10 in turbo mode. Using flamegraphs, I was able to identify to hotspots and optimize them with some relatively straightforward changes. Before these changes, I got the following results with sbt 1.4.0-SNAPSHOT:
 project | min (ms) | max (ms) | median (ms) | total (ms) | cpu % |
:------- | :------: | :------: | :-------: | :--------: | :---: |
sbt-1.4.0-SNAPSHOT (3 source files) | 151 | 263 | 166 | 34787 | 0.2
sbt-1.4.0-SNAPSHOT (5003 source files) | 747 | 1016 | 836 | 103423 | 0.1

After these changes, the performance improved:
 project | min (ms) | max (ms) | median (ms) | total (ms) | cpu % |
:------- | :------: | :------: | :-------: | :--------: | :---: |
sbt-1.4.0-SNAPSHOT (3 source files) | 140 | 346 | 149 | 27990 | 0.2
sbt-1.4.0-SNAPSHOT (5003 source files) | 458 | 574 | 470 | 68228 | 0.2

Note that there is a lot of time spent stamping source files outside of zinc, scalatest and the scala compiler so the zinc overhead was probably about 400ms before these optimizations and 100ms after. 

By comparison, for sbt 1.3.10, I got
 project | min (ms) | max (ms) | median (ms) | total (ms) | cpu % |
:------- | :------: | :------: | :-------: | :--------: | :---: |
sbt-1.3.10-turbo (3 source files) | 119 | 220 | 125 | 25906 | 3.1
sbt-1.3.10-turbo (5003 source files) | 535 | 746 | 558 | 77775 | 3.5

I couldn't quite figure out why sbt 1.4.0-SNAPSHOT was consistently 20ms slower than 1.3.10 in the 3 source file example. It was reproducible in the shell though. I ran sbt with `-Dsbt.task.timings=true` and it appeared that about 10ms of the extra time was in `compileIncremental` and the other 10ms was somewhere that didn't show up in the task timing report.

The variance seemed generally lower after these changes which is probably because I reworked things so that less garbage is generated so GC is probably running less often.

I also just benchmarked running `compile` in the sbt shell in a scenario when we just needed to compile a single source file (`object Foo`) and with the head of origin/develop in zinc, it took about 700ms when there were 5000 other source files and with these zinc changes, it dropped down to 500ms.

### DIFF
--- a/internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/JavaAnalyze.scala
+++ b/internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/JavaAnalyze.scala
@@ -192,7 +192,8 @@ private[sbt] object JavaAnalyze {
 
       def readInheritanceDependencies(classes: Seq[Class[_]]) = {
         val api = readAPI(source, classes)
-        api.groupBy(_._1).mapValues(_.map(_._2))
+        // avoid .mapValues(...) because of its viewness (scala/bug#10919)
+        api.groupBy(_._1).iterator.map { case (k, v) => k -> v.map(_._2) }
       }
 
       // Read API of non-local classes and process dependencies by inheritance
--- a/internal/zinc-classpath/src/main/scala/sbt/internal/inc/classpath/ClassLoaders.scala
+++ b/internal/zinc-classpath/src/main/scala/sbt/internal/inc/classpath/ClassLoaders.scala
@@ -78,7 +78,9 @@ final class ClasspathFilter(parent: ClassLoader, root: ClassLoader, classpath: S
         |  cp = $classpath
         |)""".stripMargin
 
-  private[this] val directories: Seq[Path] = classpath.toSeq.filter(Files.isDirectory(_))
+  private[this] val directories: Seq[Path] = classpath.toSeq.filter { p =>
+    !p.toString.endsWith(".jar") && Files.isDirectory(p)
+  }
   override def loadClass(className: String, resolve: Boolean): Class[_] = {
     val c = super.loadClass(className, resolve)
     if (includeLoader(c.getClassLoader, root) || fromClasspath(c))
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
@@ -21,6 +21,7 @@ import xsbti.compile.{ Changes, DependencyChanges, IncOptions, Output }
 import xsbti.compile.{ ClassFileManager => XClassFileManager }
 import xsbti.compile.analysis.{ ReadStamps, Stamp => XStamp }
 import scala.collection.Iterator
+import scala.collection.parallel.immutable.ParVector
 import Incremental.{ CompileCycle, CompileCycleResult, IncrementalCallback, PrefixingLogger }
 
 /**
@@ -385,7 +386,7 @@ private[inc] abstract class IncrementalCommon(
         val changed0 = new java.util.HashSet[VirtualFileRef]
         val removed0 = new java.util.HashSet[VirtualFileRef]
         val unmodified0 = new java.util.HashSet[VirtualFileRef]
-        sources.foreach {
+        new ParVector(sources.toVector).foreach {
           case f: VirtualFileRef if previousSources.contains(f) =>
             if (equivS.equiv(previous.source(f), stamps.source(f))) unmodified0.add(f)
             else changed0.add(f)
@@ -404,11 +405,12 @@ private[inc] abstract class IncrementalCommon(
 
     val removedProducts: Set[VirtualFileRef] =
       lookup.removedProducts(previousAnalysis).getOrElse {
-        previous.allProducts
+        new ParVector(previous.allProducts.toVector)
           .filter(p => {
             // println(s"removedProducts? $p")
             !equivS.equiv(previous.product(p), stamps.product(p))
           })
+          .toVector
           .toSet
       }
 
@@ -423,7 +425,7 @@ private[inc] abstract class IncrementalCommon(
           converter,
           log
         )
-      previous.allLibraries.filter(detectChange).toSet
+      new ParVector(previous.allLibraries.toVector).filter(detectChange).toVector.toSet
     }
 
     val subprojectApiChanges: APIChanges = {
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
@@ -375,20 +375,30 @@ private[inc] abstract class IncrementalCommon(
     val previousRelations = previousAnalysis.relations
 
     val sourceChanges: Changes[VirtualFileRef] = lookup.changedSources(previousAnalysis).getOrElse {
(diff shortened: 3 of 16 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/845/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -375,20 +375,30 @@ private[inc] abstract class IncrementalCommon(
...
-        val removed = previousSources -- inBoth
-        val added = (sources -- inBoth).map(x => x: VirtualFileRef)
-        val (changed0, unmodified0) =
-          inBoth.partition(f => !equivS.equiv(previous.source(f), stamps.source(f)))
-        val changed = changed0.map(x => x: VirtualFileRef)
-        val unmodified = unmodified0.map(x => x: VirtualFileRef)
+        val added0 = new java.util.HashSet[VirtualFileRef]
+        val changed0 = new java.util.HashSet[VirtualFileRef]
+        val removed0 = new java.util.HashSet[VirtualFileRef]
+        val unmodified0 = new java.util.HashSet[VirtualFileRef]
```
eed3si9n, line 388: 👍

internal/zinc-classpath/src/main/scala/sbt/internal/inc/classpath/ClassLoaders.scala
```diff
@@ -78,7 +78,9 @@ final class ClasspathFilter(parent: ClassLoader, root: ClassLoader, classpath: S
         |  cp = $classpath
         |)""".stripMargin
 
-  private[this] val directories: Seq[Path] = classpath.toSeq.filter(Files.isDirectory(_))
+  private[this] val directories: Seq[Path] = classpath.toSeq.filter { p =>
+    !p.toString.endsWith(".jar") && Files.isDirectory(p)
```
eed3si9n, line 82: 👍

internal/zinc-persist/src/main/scala/sbt/internal/inc/binary/converters/ProtobufWriters.scala
```diff
@@ -323,33 +308,26 @@ final class ProtobufWriters(mapper: WriteMapper) {
...
-    Schema.Type.Structure.newBuilder
-      .addAllDeclared(declared.asJava)
-      .addAllInherited(inherited.asJava)
-      .addAllParents(parents.asJava)
-      .build
+    val builder = Schema.Type.Structure.newBuilder
+    tpe.declared().foreach(t => builder.addDeclared(toClassDefinition(t)))
+    tpe.inherited().foreach(i => builder.addInherited(toClassDefinition(i)))
+    tpe.parents().foreach(p => builder.addParents(toType(p)))
+    builder.build
```
eed3si9n, line 322: Thanks for this! In the hindsight 2020 it makes sense that `.asJava` that I added in https://github.com/sbt/zinc/pull/799 was suboptimal.

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/JavaAnalyze.scala
```diff
@@ -192,7 +192,7 @@ private[sbt] object JavaAnalyze {
 
       def readInheritanceDependencies(classes: Seq[Class[_]]) = {
         val api = readAPI(source, classes)
-        api.groupBy(_._1).mapValues(_.map(_._2))
+        api.groupBy(_._1).iterator.map { case (k, v) => k -> v.map(_._2) }
```
eed3si9n:
```suggestion
        // avoid .mapValues(...) because of its viewness (scala/bug#10919)
        api.groupBy(_._1).iterator.map { case (k, v) => k -> v.map(_._2) }
```

internal/zinc-persist/src/main/scala/sbt/internal/inc/binary/converters/ProtobufWriters.scala
eatkins: Teamwork!

In the absence of automated performance testing, it seems inevitable that this kind of performance regression might occur. I wouldn't have noticed the regression were it not for https://github.com/eatkins/scala-build-watch-performance. That being said, a useful heuristic that I try to follow is to avoid copying data if at all possible. It's often hard to tell in isolation whether or not copying will impact performance. I also try to practice the boy scout rule of leaving things better than I found them.

internal/zinc-classfile/src/main/scala/sbt/internal/inc/classfile/JavaAnalyze.scala
eatkins: https://github.com/sbt/zinc/compare/0f2304ee2d4ab23e7038988782a90eef403c3f53..a1492e6f2a7529a41d98359277c8213fd3168fc1

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -385,7 +386,7 @@ private[inc] abstract class IncrementalCommon(
         val changed0 = new java.util.HashSet[VirtualFileRef]
         val removed0 = new java.util.HashSet[VirtualFileRef]
         val unmodified0 = new java.util.HashSet[VirtualFileRef]
-        sources.foreach {
+        new ParVector(sources.toVector).foreach {
```
eed3si9n: 👍

eed3si9n (approved):
LGTM

eed3si9n:
Let's keep this open for a few days in case @dwijnand wants to take a look as well.

retronym:
LGTM. I have a few WIP performance improvements but this looks like a great start and I'll rebase atop it!

retronym (approved):

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
eed3si9n: I'm starting to see some under-detection around 30~50% of the times when I call `zincJVM2_13/testOnly sbt.inc.IncrementalCompilerSpec`

```
[info] - should not compile anything if source has not changed *** FAILED ***
[info]   result2.hasModified() was true (IncrementalCompilerSpec.scala:57)
```

Maybe the HashSet needs to be a concurrent-ready Set?

eed3si9n: I have a fix for this here - https://github.com/sbt/zinc/pull/849/commits/4f14d025260a9c78c09e1706262a79fe5cf6be07

eatkins: Good catch. I had noticed that ci was a little flaky and was wondering what could have been going on.

PR #853, opened 2020-07-24
Fix WrappedSet.contains

eatkins:
I noticed from code inspection that the contains method of WrappedSet
was not correctly implemented. It doesn't appear that it is called in
zinc but I figured it would still be good to fix in case it ever is. (My
assumption was that only the iterator method of these Sets would be
called since they just hold the source and binary changes between
analysis runs. Avoiding creating an immutable set when when all we
needed was an iterator over the existing data was the reason for the
WrappedSet class in the first place.)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/853/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):
Oh snap.

PR #860, opened 2020-07-27
Set -Ypickle-write when required.

dwijnand:
Reverts the regression introduced in 9310afe6cf52d7feed83d25eda622d0e9256a158.

### DIFF
--- a/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
@@ -143,10 +143,11 @@ final class MixedAnalyzingCompiler(
     val isPickleJava = config.currentSetup.order == Mixed && config.incOptions.pipelining && javaSrcs.nonEmpty
 
     val earlyOut = config.earlyOutput.flatMap(_.getSingleOutputAsPath.toOption)
-    earlyOut foreach { out =>
-      if (out.toString.endsWith(".jar") && !Files.exists(out)) {
+    val pickleWrite = earlyOut.toList.flatMap { out =>
+      val sbv = scalac.scalaInstance.version.take(4)
+      if (out.toString.endsWith(".jar") && !Files.exists(out))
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/860/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
```diff
@@ -143,10 +143,11 @@ final class MixedAnalyzingCompiler(
...
 
     val earlyOut = config.earlyOutput.flatMap(_.getSingleOutputAsPath.toOption)
-    earlyOut foreach { out =>
-      if (out.toString.endsWith(".jar") && !Files.exists(out)) {
+    val pickleWrite = earlyOut.toList.flatMap { out =>
+      val sbv = scalac.scalaInstance.version.take(4)
+      if (out.toString.endsWith(".jar") && !Files.exists(out))
         scala.reflect.io.RootPath(out, writable = true).close() // creates an empty jar
-      }
+      List("-Ypickle-write", out.toString).filter(_ => sbv == "2.12" || sbv == "2.13")
```
eed3si9n, line 150: In general, Zinc should avoid adding Scalac options since the build user wouldn't be able to control it.
`-Y` flag especially has no guarantee and it's not even supported in all of 2.12 or 2.13.

dwijnand: Do you want to send a PR with a way to address this, without breaking the initial use case?

eed3si9n: What is the initial use case?

Issue #593, opened 2018-09-04
LookupImpl.lookupAnalysis is O(N) wrt classpath length


retronym:
In projects with long classpaths, I've seen this show up in profiles, as called by `memberRef` => `binaryDependency` => `externalDependency` => `getExternalApi`.

https://github.com/sbt/zinc/blob/b125d8295cea2c0afc3c3686c1853edf93c4c934/zinc/src/main/scala/sbt/internal/inc/LookupImpl.scala#L43

Could we maintain an index of the package heirarchy to the JARs that contribute to the package?

jvican:
Yes, I'll submit a PR with this solution.

Friendseeker:
I'm reviewing this issue and am interested in the setup used for profiling `lookupAnalysis`, specifically the projects with long classpaths where `lookupAnalysis` is prominent in benchmarks. I'm currently working on [related improvement](https://github.com/sbt/zinc/pull/1279) and any details about the setup or benchmarks would be extremely helpful.

Friendseeker:
c.c. https://github.com/sbt/zinc/pull/1279#issuecomment-2406060705 need to investigate which build tools are already passing in an optimized implementation of `perClasspathEntryLookup` and address the root problem.

PR #864, opened 2020-07-28
Cache the jar class definitions

eatkins:
It is slow to compute all of the classes in every jar on the classpath.
We can cache the jar class names between invocations so long as we check
that the timestamp of the jar hasn't changed. My assumption is that the
timestamps of library jars rarely change so the cache entry will rarely
be invalidated. It isn't catastrophic if it is.

To ensure that the caching isn't global, I introduce a class,
Locate.DefinesClass, which provides the same api as Locate.definesClass,
but it can cache the the jar entries. By using a class, we allow the
build tool to control whether or not they want to cache at all.

On my laptop, this reduced the average runtime for a no-op test compile in a
simple project from about 70ms to 60ms. The project had a fairly small
dependency classpath:
org/scala-lang/scala-library/2.13.3/scala-library-2.13.3.jar
org/scalatest/scalatest_2.13/3.0.8/scalatest_2.13-3.0.8.jar
com/lihaoyi/utest_2.13/0.7.4/utest_2.13-0.7.4.jar
org/scalactic/scalactic_2.13/3.0.8/scalactic_2.13-3.0.8.jar
org/scala-lang/scala-reflect/2.13.3/scala-reflect-2.13.3.jar
org/scala-lang/modules/scala-xml_2.13/1.2.0/scala-xml_2.13-1.2.0.jar
org/scala-sbt/test-interface/1.0/test-interface-1.0.jar
org/portable-scala/portable-scala-reflect_2.13/0.1.1/portable-scala-reflect_2.13-0.1.1.jar

so I suspect the difference will be more significant for projects with a
heavier dependency graph.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Locate.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Locate.scala
@@ -15,8 +15,10 @@ package inc
 
 import java.nio.file.{ Files, Path }
 import java.util.zip.{ ZipException, ZipFile }
+import java.util.concurrent.ConcurrentHashMap
+import sbt.io.IO
 import xsbti.{ PathBasedFile, VirtualFile }
-import xsbti.compile.{ DefinesClass, PerClasspathEntryLookup }
+import xsbti.compile.{ DefinesClass => XDefinesClass, PerClasspathEntryLookup }
 
 object Locate {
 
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/864/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/Locate.scala
```diff
@@ -64,7 +66,38 @@ object Locate {
...
+   * with a particular name. It will cache the set of class names for each jar file.
+   * Each jar cache entry is invalidated when the last modified time of the jar file
+   * changes.
+   *
+   */
+  sealed abstract class DefinesClass {
+    def apply(entry0: VirtualFile): XDefinesClass
+  }
+  class CachingDefinesClass extends DefinesClass {
+    private[this] val jarCache = new ConcurrentHashMap[Path, (JarDefinesClass, Long)]
```
eed3si9n, line 80: This makes me a bit nervous because there could be thousands of JAR files. Would it make sense to use Caffeine (or equivalent) that could put upper limit on the cache size and keep everything in a weak reference?

eatkins: Possibly. My guess though is that if the classpath is thousands of files that this cache will still represent a small fraction of the total memory of the process running zinc.

eatkins: Unless we shaded caffeine, we probably shouldn't put it in Zinc so we'd need to do it at the sbt level. It's a bit tricky to rework the api because I'd probably need to expose `JarDefinesClass`.

When I tested this in sbt, I added a setting `cacheClasspathEntries` that was enabled by default but could be disabled. Having the binary switch seems better for now than trying to tune the size of a caffeine cache.

That being said, it makes me wonder if we should have something that's the opposite of turbo mode that disables as many of the caches as possible.

eatkins: Also, it's worth noting that the default behavior in zinc is not cached so this shouldn't impact any downstream zinc users unless they opt in.

eed3si9n: Looking at sbt side I see:

```scala
      classpathEntryDefinesClass := {
        val converter = fileConverter.value
        val f = FileValueCache({ x: NioPath =>
          if (x.getFileName.toString != "rt.jar") Locate.definesClass(converter.toVirtualFile(x))
          else ((_: String) => false): DefinesClass
        }).get;
        { (x: File) =>
          f(x.toPath)
        }
      },
```

I guess the facility of caching per file is already there, but it's limited to per session because it's a task?

dwijnand:
Does this address https://github.com/sbt/zinc/issues/593?

eed3si9n:
@dwijnand 

> Does this address #593?

I don't think so. `lookupAnalysis(binaryClassName: String): Option[CompileAnalysis]` is about querying subproject Analysis one by one, whereas this is one is about library JAR files defining a particular class.

eatkins:
We decided to handle this in sbt instead: https://github.com/sbt/sbt/pull/5734

PR #865, opened 2020-07-28
Use toIterator instead of toStream

eatkins:
Historically Streams have not performed nearly as well as Iterator. I'm
not sure how relevant that is the case for the latest versions of scala
but there were a number of places where it was easy to switch from
Streams to Iterators so this commit does just that. There is very little
expressivity loss in the switch from Iterator to Stream.

From a performance perspective, I have been testing the zinc overhead in
a small project by looping 100 iterations of no-op Test / compile. While
not a scientific benchmark by any means, I pretty consistently found
zinc to run about 10% faster (~63ms average after this change compared
to ~70 ms before).

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/CallbackGlobal.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/CallbackGlobal.scala
@@ -41,7 +41,8 @@ sealed abstract class CallbackGlobal(
     output match {
       case single: SingleOutput => List(single.getOutputDirectoryAsPath)
       // Use Stream instead of List because Analyzer maps intensively over the directories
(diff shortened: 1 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/865/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):
Thanks!

PR #869, opened 2020-08-07
Reduce io in ClasspathCache

eatkins:
I noticed in a flame graph that we were calling Files.exists and
Files.readAttributes in the same place. It should be unnecessary to call
Files.exists because Files.readAttributes will throw a
NoSuchFileException if it doesn't exist. This would be probably be a bad
change if we were checking files that frequently didn't exist but I
doubt that's the case. We also probably should have had the catch of
NoSuchFileException regardless of whether we did the existence check
because it's always possible albeit unlikely that the file is deleted in
between the existence check and reading the attributes.

I also saw that we could avoid computing the last modified time with
IO.getModifiedTimeOrZero when using jdk > 10 since newer jdks have
millisecond timestamp precision.

### DIFF
--- a/zinc/src/main/scala/sbt/internal/inc/caching/ClasspathCache.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/caching/ClasspathCache.scala
@@ -11,15 +11,30 @@
 
 package sbt.internal.inc.caching
 
-import java.nio.file.{ Files, Path }
-import java.util.concurrent.ConcurrentHashMap
+import java.nio.file.{ Files, NoSuchFileException, Path }
 import java.nio.file.attribute.{ BasicFileAttributes, FileTime }
+import java.util.concurrent.ConcurrentHashMap
 
 import xsbti.compile.FileHash
 import sbt.internal.inc.{ EmptyStamp, Stamper }
 import sbt.io.IO
+import scala.util.control.NonFatal
 
 object ClasspathCache {
+  // copied from io
+  private val jdkTimestamps = {
+    System.getProperty("sbt.io.jdktimestamps") match {
+      case null =>
+        System.getProperty("java.specification.version") match {
+          case null => false
+          case sv =>
+            try {
+              sv.split("\\.").last.toInt >= 10
+            } catch { case NonFatal(_) => false }
+        }
+      case p => p.toLowerCase != "false"
+    }
+  }
   // For more safety, store both the time and size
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/869/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):
Thanks @eatkins!

dwijnand (approved):
> This would be probably be a bad
> change if we were checking files that frequently didn't exist but I
> doubt that's the case.

So this is adding some extra throwable CPU/allocation time to the IO of looking for a non-existent file (just had a quick read through https://shipilev.net/blog/2014/exceptional-performance/#_results).  That is assuming `Files.readAttributes` and `Files.exists` have the same performance?.


I'm not so sure using non-existing files is so rare, but nonetheless I expect the overhead to be negligible, so LGTM too.

eatkins:
My assumption was that even though filling a stack trace is expensive, performing io is even more expensive. If that's true, dropping the existence check would be a slight performance hit in the unhappy path but it should be about a 2x speedup for the happy path and less than a 2x hit in the unhappy path. On my test project, I saw pretty much the exact speedups that I expected in flamegraphs but it's certainly possible that non-existent files are more common in other projects (I think they all existed in my project).

The only clue I had about the use case was the comment `// readAttributes needs to be guarded by file.exists(), otherwise it fails` which made me think that the existence check was to avoid having `readAttributes` throw a NoSuchFileException since that is the failure mode for `readAttributes` for a file that doesn't exist. Unfortunately, the comment says nothing about when and why the file wouldn't exist so I don't know how common the failure mode is.

This issue is also further strengthens my opinion that it was/is a design mistake in scala that checked exceptions from java are ignored. Even if scala itself didn't want checked exceptions for its own apis, I think it should enforce checked exceptions in java code. At a minimum, there should be a compiler flag for this. Because it doesn't, I can basically never be sure if code in scala is going to throw or not unless I have written all of it myself which makes scala feel mildly unsafe at all times. I know checked exceptions are not a widely loved feature in java but I really don't understand why scala just ignores them given the consequence of doing so.

PR #873, opened 2020-08-10
Add replace existing flag to Files.move

eatkins:
Sbt is unable to recompile itself with 1.4.0-SNAPSHOT if any java
sources change in the main project. It throws a FileAlreadyExistsException pointing to
this line.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/ClassFileManager.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/ClassFileManager.scala
@@ -16,7 +16,7 @@ package inc
 import sbt.io.IO
 import java.io.File
 import java.util.Optional
-import java.nio.file.{ Files, Path }
+import java.nio.file.{ Files, Path, StandardCopyOption }
 
 import collection.mutable
 import xsbti.{ FileConverter, PathBasedFile, VirtualFile }
@@ -227,7 +227,7 @@ object ClassFileManager {
             if (!Files.exists(origPath.getParent)) {
               Files.createDirectories(origPath.getParent)
             } // if
-            Files.move(tmp, origPath)
+            Files.move(tmp, origPath, StandardCopyOption.REPLACE_EXISTING)
           } // if
         }
       }


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/873/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
I don't always see this error so I don't know the specific thing that triggers it. Regardless, I think the code change makes sense.

internal/zinc-core/src/main/scala/sbt/internal/inc/ClassFileManager.scala
```diff
@@ -227,7 +227,7 @@ object ClassFileManager {
             if (!Files.exists(origPath.getParent)) {
               Files.createDirectories(origPath.getParent)
             } // if
-            Files.move(tmp, origPath)
+            Files.move(tmp, origPath, StandardCopyOption.REPLACE_EXISTING)
```
eed3si9n, line 230: This change itself looks fine, but I'm increasingly starting to think that using NIO `Files` directly is an anti-pattern / source of subtle bugs.

eatkins: Is this really subtle though? It throws a noisy exception that is easy to fix. A number of the apis in sbt.io.IO are much more subtle. For example, IO.delete doesn't report if it failed. I actually like that nio forces you to be specific about your assumptions.

dwijnand (approved):
I was just in this file today, so I'm actually familiar with this code for once.  😄

LGTM too

internal/zinc-core/src/main/scala/sbt/internal/inc/ClassFileManager.scala
eed3si9n: For things like moving files, deleting files/directories, and creating them, most usage around compiler or build in general I'm used to the way `sbt.io.IO` just do the thing for the most part forcefully whereas NIO `Files` seems to be much more strict about some sort of precondition, which we may or may not be able to meet in parallel situation.

eatkins: I hear you. It's not exactly that I am in favor of nio so much as I have been burned by `IO`. Since ultimately `IO` delegates to java.io or java.nio.file, I just prefer directly working with those apis. Also because I am experienced with nio, I can tell at a glance what the performance characteristics are. That isn't true with `IO`. You're right that `IO` is generally easier to get to do the "right thing" from the sbt perspective. It's just that often the implementation isn't optimal or it isn't obvious what the semantics are without reading the source code. If I were starting from scratch, I would probably not put either `java.io.File` or `java.nio.file.Path` anywhere in the public api and then it would be easier to define an io dsl with good performance and default semantics that match the sbt use case.

eed3si9n (approved):

Issue #777, opened 2020-05-04
Epic: Integrate changes in mspnf/zinc


dwijnand:
* [x] [Build pipelining][1] (also scala/scala-dev#643) #744
  * [x] changes to the `InvalidationProfiler` (not the `synchronized`s) [[`545c917` (#744)](https://github.com/sbt/zinc/pull/744/commits/545c917961589b7f1356fc4bb128435273a07c0c) ](https://github.com/sbt/zinc/pull/744/commits/545c917961589b7f1356fc4bb128435273a07c0c)
  * [x] ~cancellation: Zinc users need access to nsc's `reporter.cancelled = true`~ (`advance` is sufficient)
* [x] [Referentially transparent (RT) builds][2]
	* [x] #782 allow paths to be mapped so that artifacts read from analyses match those on the compilation classpath
    * [x] quickApi / provenance
      * [x] extend existing `Lookup` api (which is already customizable) to include `lookupAnalyzedClass` #785
      * [x] add provenance to AnalyzedClass #786
    * [x] ~override
`ExternalHooks#hashClassPath` to compute `FileHash` directly from the file names~ (nothing TODO, already available since #427)
* [x] [Virtual source files][3] #712
* [x] [Additional changes][4]
  * [x] SafeLazy concurreny issue #740
  * [x] ~optimized UsedName's scopes~ #764 (postponed)
  * [x] ~`Map<String, Object extraHooks()` + `Hooks` methods to set and retrieve specific hooks and callbacks~ (this was for expediency, replace with typed API additions
  * [x] Make hashing childrenOfSealedClass stable #788
* [x] ~additional [performance](https://github.com/sbt/zinc/issues?q=sort:updated-desc+is:open+label:performance) tickets~ (postponed)

original diff: https://github.com/sbt/zinc/compare/develop...mspnf:1.3.x-contrib?w=1 ([formatted](https://github.com/sbt/zinc/compare/develop...dwijnand:1.3.x-formatted))

[1]: https://github.com/mspnf/zinc/blob/1.3.x-contrib/README-MORGANSTANLEY.MD#changes-related-to-pipelining
[2]: https://github.com/mspnf/zinc/blob/1.3.x-contrib/README-MORGANSTANLEY.MD#changes-related-to-rt-build
[3]: https://github.com/mspnf/zinc/blob/1.3.x-contrib/README-MORGANSTANLEY.MD#changes-related-to-in-memory-source
[...]

PR #874, opened 2020-08-10
Trailing integration changes

dwijnand:
These are trailing changes to completing #777.

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/API.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/API.scala
@@ -210,7 +210,7 @@ final class API(val global: CallbackGlobal) extends Compat with GlobalHelpers wi
       !ignoredSymbol(sym) &&
       sym.isStatic &&
       !sym.isImplClass &&
-      !sym.hasFlag(Flags.JAVA) &&
+      (!sym.hasFlag(Flags.JAVA) || global.callback.isPickleJava) &&
       !sym.isNestedClass
     }
   }
--- a/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
@@ -598,6 +598,9 @@ class ExtractAPI[GlobalType <: Global](
           "sbt-api: Unexpected nullary method type " + in + " in " + in.owner
         )
         Constants.emptyType
+      case MethodType(_, _) =>
+        reporter.echo(NoPosition, s"sbt-api: Unhandled method type $in in ${in.owner}")
+        Constants.emptyType
       case _ =>
         reporter.warning(NoPosition, "sbt-api: Unhandled type " + t.getClass + " : " + t)
         Constants.emptyType
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Lookup.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Lookup.scala
@@ -129,6 +129,21 @@ trait ExternalLookup extends ExternalHooks.Lookup {
     import scala.collection.JavaConverters._
     shouldDoIncrementalCompilation(changedClasses.iterator().asScala.toSet, previousAnalysis)
   }
+
+  /**
+   * Used to override whether we should proceed with making an early output.
+   *
+   * By default we do not make an early output in the presence of any macros
+   * because macro expansion (in a downstream subproject) requires the macro implementation
+   * to be present in bytecode form, rather than just just a pickle-containing JAR.
+   *
+   * If you're careful micromanaging the separation of macro implementations
+   * (e.g. `def impl(c: Context) = ...`) from macro definitions
+   * (e.g. `def foo: Unit = macro Foo.impl`) you can safely override this.
+   */
+  def shouldDoEarlyOutput(compileAnalysis: CompileAnalysis): Boolean = {
+    compileAnalysis.asInstanceOf[Analysis].apis.internal.values.exists(_.hasMacro)
+  }
 }
 
 trait NoopExternalLookup extends ExternalLookup {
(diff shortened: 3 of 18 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/874/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/Lookup.scala
```diff
@@ -129,6 +129,8 @@ trait ExternalLookup extends ExternalHooks.Lookup {
     import scala.collection.JavaConverters._
     shouldDoIncrementalCompilation(changedClasses.iterator().asScala.toSet, previousAnalysis)
   }
+
+  def hasAnyMacro(a: Analysis): Boolean = a.apis.internal.values.exists(_.hasMacro)
```
eed3si9n:
```suggestion
  /**
   * Used to override whether we should proceed with making an early output.
   * By default we do not make an early output in the presence of any macros
   * because macros require bytecode JARs in the classpath.
   * However, with careful micromanagement there are cases where we can safely
   * start downstream subproject compilations with just the pickle JARs.
   */
  def shouldDoEarlyOutput(analysis: CompileAnalysis): Boolean =
    analysis.asInstanceOf[Analysis].apis.internal.values.exists(_.hasMacro)
```

Reading the commit comment

> If you carefully micro-manage macro implementation scopes, you can avoid bailing here.

I'm not against the direction, but if this is the case, I think the name `hasAnyMacro` is misleading.

internal/zinc-core/src/main/scala/sbt/internal/inc/PickleJar.scala
```diff
@@ -20,7 +20,13 @@ import scala.reflect.io.RootPath
...
 object PickleJar {
   def write(pickleOut: Path, knownProducts: java.util.Set[String]): Unit = {
-    if (!Files.exists(pickleOut)) {
+    val pickleUpdatesJar = pickleOut.resolveSibling(s"${pickleOut.getFileName}-updates.jar")
+    if (Files.exists(pickleUpdatesJar)) {
+      if (Files.exists(pickleOut))
+        JarUtils.mergeJars(pickleOut.toFile, pickleUpdatesJar.toFile)
+      else
+        Files.move(pickleUpdatesJar, pickleOut)
+    } else if (!Files.exists(pickleOut)) {
```
eed3si9n: Would this fix the case of compiling `A`, then compiling `B`, and then do `use/compile` from another subproject?

Could you add repro case by adding something like https://github.com/sbt/zinc/tree/develop/zinc/src/sbt-test/source-dependencies/subproject-pipelining?

internal/compiler-bridge/src/main/scala/xsbt/API.scala
```diff
@@ -210,7 +210,7 @@ final class API(val global: CallbackGlobal) extends Compat with GlobalHelpers wi
       !ignoredSymbol(sym) &&
       sym.isStatic &&
       !sym.isImplClass &&
-      !sym.hasFlag(Flags.JAVA) &&
+      (!sym.hasFlag(Flags.JAVA) || global.callback.isPickleJava) &&
```
eed3si9n, line 213: Would this fix the Scala source depending on Java from another subproject?
If so could you add a repro case similar to https://github.com/sbt/zinc/tree/develop/zinc/src/sbt-test/source-dependencies/subproject-pipelining?

internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
```diff
@@ -598,6 +598,8 @@ class ExtractAPI[GlobalType <: Global](
           "sbt-api: Unexpected nullary method type " + in + " in " + in.owner
         )
         Constants.emptyType
+      case MethodType(_, _) =>
+        Constants.emptyType
```
eed3si9n: Hit the merge button on https://github.com/sbt/zinc/pull/689?

internal/zinc-core/src/main/scala/sbt/internal/inc/PickleJar.scala
dwijnand: Change dropped (superseded by the isPickleJava change)

internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
dwijnand: Went with the lesser change of making it echo instead of warning.

internal/zinc-core/src/main/scala/sbt/internal/inc/Lookup.scala
```diff
@@ -129,6 +129,19 @@ trait ExternalLookup extends ExternalHooks.Lookup {
...
+   *
+   * By default we do not make an early output in the presence of any macros
+   * because macro expansion (in a downstream subproject) requires the macro implementation
+   * to be present in bytecode form, rather than just just a pickle-containing JAR.
+   *
+   * If you're careful micromanaging the separation of macro implementations
+   * (e.g. `def impl(c: Context) = ...`) from macro definitions
+   * (e.g. `def foo: Unit = macro Foo.impl`) you can safely override this.
+   */
+  def shouldDoEarlyOutput(a: Analysis): Boolean = a.apis.internal.values.exists(_.hasMacro)
```
eed3si9n: This should take `CompileAnalysis` as input parameter because `Analysis` is private.

eed3si9n (approved):
LGTM

PR #880, opened 2020-08-13
Improve incremental compilation on multi module project 

### LINKED ISSUE #810: false-positives "Detected a change in a public API"
Coming from https://github.com/sbt/sbt/issues/5570#issuecomment-653599363 I also see a lot of messages like the the original poster reported

```
[debug]The com.nezasa.model.iam.User has the following regular definitions changed:
[debug]         UsedName(failsInventoryOwnerCheck,[Default]).
```

In my work project, right now this is the only change that I now have in the logs, but in previous sessions I noticed others...

Running `compile` multiple times without changing any source files always triggers this message, and If I enable better logs with `incOptions ~= (o => o.withApiDebug(true).withRelationsDebug(true)),` I can see:

```
[debug] [diff] Detected a change in a public API (com.nezasa.model.iam.User):
[debug] [diff] @scala.this#SerialVersionUID(value = 1006L) class com.nezasa.model.iam.User extends scala.this#Serializable with java.io.this#Serializable with scala.this#Product with scala.this#Equals with com.nezasa.comms.this#Logging with com.nezasa.model.iam.this#EmailRecipient with com.nezasa.model.serialization.this#NezasaModelObject with com.nezasa.model.serialization.this#NezasaModelCOObject with com.nezasa.model.serialization.this#NezasaModelTOObject with com.mohiva.play.silhouette.api.this#Identity with java.lang.this#Object with scala.this#Any {
[debug] [diff] 
[debug] [diff] 
[debug] [diff] 
[debug] [diff] }
[debug] [diff] object com.nezasa.model.iam.User extends scala.this#Serializable with java.io.this#Serializable with com.nezasa.model.iam.this#UserDatabase with com.nezasa.dao.utils.this#DaoAdapter[com.nezasa.dao.this#UserDao] with com.nezasa.comms.this#Logging with com.nezasa.model.serialization.this#NezasaModelCompanion[com.nezasa.model.iam.this#User, com.nezasa.model.iam.this#UserTO] with java.lang.this#Object with scala.this#Any {
[debug] [diff] 
[debug] [diff] 
[debug] [diff] 
[debug] [diff] }
[debug] The com.nezasa.model.iam.User has the following regular definitions changed:
[debug]         UsedName(failsInventoryOwnerCheck,[Default]).
[debug] All member reference dependencies will be considered within this context.
[debug] Files invalidated by inheriting from (external) com.nezasa.model.iam.User: Set(); now invalidating by inheritance (internally).
[debug] Getting direct dependencies of all classes transitively invalidated by inheritance.
[debug] Getting classes that directly depend on (external) com.nezasa.model.iam.User.
[debug] None of the modified names appears in source file of controllers.location.api.AirportsApiController. This dependency is not being considered for invalidation.
[debug] None of the modified names appears in source file of controllers.location.api.AreasApiController. This dependency is not being considered for invalidation.
[debug] No changes
```

However I dont understand what that diff represents...

Something that I also noted is that this is a multimodule project and although this `User` class is used pervasively in all the modules only a single module is showing this message...

jtjeferreira:
Fixes #810

Consider a multimodule build with subproject `core` and subproject `app` where `app` depends on `core`. When there are source code changes on `core` that does NOT cause invalidations on `app` then the current `Analysis` is not updated to include the subproject ("external") changes. Subsequent incremental compilations keep considering the changes despite nothing changing.

The fix is to update the Analysis file with the external changes even if there are no invalidations in the current module.

**Note**: In other words, this change would in some cases introduce changes to `Analysis` file when the output JAR itself would be identical.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -396,6 +396,7 @@ object Incremental {
         )
     }
     val hasModified = initialInvClasses.nonEmpty || initialInvSources.nonEmpty
+    val hasSubprojectChange = initialChanges.external.apiChanges.nonEmpty
     val analysis = withClassfileManager(options, converter, output, outputJarContent) {
       classfileManager =>
         if (hasModified)
@@ -413,12 +414,24 @@ object Incremental {
             1
           )
         else {
+          val analysis =
+            if (hasSubprojectChange)
+              previous.copy(
+                apis = initialChanges.external.allModified.foldLeft[APIs](previous.apis) {
+                  (apis, clazz) =>
+                    lookup.lookupAnalyzedClass(clazz, None) match {
+                      case Some(ac) => apis.markExternalAPI(clazz, ac)
+                      case _        => apis
+                    }
+                }
+              )
+            else previous
           if (earlyOutput.isDefined)
-            writeEarlyOut(lookup, progress, earlyOutput, previous, new java.util.HashSet)
-          previous
+            writeEarlyOut(lookup, progress, earlyOutput, analysis, new java.util.HashSet)
+          analysis
         }
     }
-    (hasModified, analysis)
+    (hasModified || hasSubprojectChange, analysis)
   }
 
   /**
--- a/internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
+++ b/internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
@@ -229,6 +229,9 @@ class IncHandler(directory: Path, cacheDir: Path, scriptedLog: ManagedLogger, co
     onArgs("checkDependencies") {
       case (p, cls :: dependencies, i) => p.checkDependencies(i, dropRightColon(cls), dependencies)
     },
+    onArgs("checkExists") {
+      case (p, cls :: name :: Nil, i) => p.checkExists(i, dropRightColon(cls), name)
+    },
     noArgs("checkSame") { case (p, i)   => p.checkSame(i) },
     onArgs("run") { case (p, params, i) => p.run(i, params) },
     noArgs("package") { case (p, i)     => p.packageBin(i) },
@@ -479,6 +482,12 @@ case class ProjectStructure(
         assert(expected == actual, s"Expected $expected dependencies, got $actual")
 
       assertDependencies(expected.toSet, classDeps(className))
+      ()
+    }
+
+  def checkExists(i: IncState, className: String, name: String): Future[Unit] =
+    compile(i).map { analysis =>
+      assert(analysis.apis.externalAPI(className).nameHashes().exists(_.name() == name))
       ()
     }
 
--- a/zinc/src/sbt-test/source-dependencies/subproject-api-update/build.json
+++ b/zinc/src/sbt-test/source-dependencies/subproject-api-update/build.json
@@ -1,0 +1,15 @@
+{
+  "projects": [
+    {
+      "name": "root",
+      "dependsOn": [
+        "core"
+      ],
+      "scalaVersion": "2.13.3"
+    },
+    {
+      "name": "core",
+      "scalaVersion": "2.13.3"
+    }
+  ]
+}
--- a/zinc/src/sbt-test/source-dependencies/subproject-api-update/build.sbt
+++ b/zinc/src/sbt-test/source-dependencies/subproject-api-update/build.sbt
@@ -1,0 +1,1 @@
+//ThisBuild / logLevel         := Level.Debug
--- a/zinc/src/sbt-test/source-dependencies/subproject-api-update/changes/A.scala
+++ b/zinc/src/sbt-test/source-dependencies/subproject-api-update/changes/A.scala
@@ -1,0 +1,5 @@
+package core
+
+class A {
+  protected val b: String = "A"
+}
--- a/zinc/src/sbt-test/source-dependencies/subproject-api-update/core/src/main/scala/A.scala
+++ b/zinc/src/sbt-test/source-dependencies/subproject-api-update/core/src/main/scala/A.scala
@@ -1,0 +1,5 @@
+package core
+
+class A {
+  protected val a: String = "A"
+}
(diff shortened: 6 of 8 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/880/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

zinc/src/sbt-test/source-dependencies/issue-810/test
```diff
@@ -0,0 +1,9 @@
+> compile
+
+> checkAPI core.A -513038468
```
jtjeferreira: Unfortunately the hash is not stable between java8 and java 11 and the test fails in java 8

jtjeferreira: @eed3si9n any suggestion on how to write the assertion? other than this the PR is ready for review...

eed3si9n: If the values are not stable, that sounds like a bug we should address on its own.
But regardless, if it isn't then maybe it's not a good thing to test on. Can you say something like `A.b` exists?

jtjeferreira: > If the values are not stable, that sounds like a bug we should address on its own.

Yeah I started to check why the hashes were not stable, and is due to some inherited methods from java.lang.Object having different annotations since java 9 (`@HotSpotIntrinsicCandidate` https://bugs.openjdk.java.net/browse/JDK-8076112). This can be seen by comparing the debug logs

```
java 8
[debug] [diff] ^inherited^ @scala.this#native( = ()) def hashCode(): scala.this#Int

java 11
[debug] [diff] ^inherited^ @scala.this#native( = ()) @jdk.internal.this#HotSpotIntrinsicCandidate( = ()) def hashCode(): scala.this#Int
```

Even if we ignored this annotation, the order of the methods is not exactly the same which would most likely also affect the hash...


> But regardless, if it isn't then maybe it's not a good thing to test on. Can you say something like `A.b` exists?

I am going to try that...

internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -413,12 +414,21 @@ object Incremental {
             1
           )
         else {
+          val analysis =
+            if (hasSubprojectChange)
+              previous.copy(
+                apis = initialChanges.external.allModified.foldLeft[APIs](previous.apis) {
+                  (apis, clazz) =>
+                    apis.markExternalAPI(clazz, lookup.lookupAnalyzedClass(clazz, None).get)
```
jtjeferreira: Passing `None` is okay? using `.get` is okay?

zinc/src/sbt-test/source-dependencies/issue-810/build.json
```diff
@@ -0,0 +1,15 @@
+{
```
jtjeferreira: the name of the folder `issue-810` is okay?

jtjeferreira (commented):
@eed3si9n ready for review

internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
eed3si9n: The second parameter was something that got added by @dwijnand recently for potential perf benefit build tools can make. I think it's ok to pass in `None` here.

`.get` on the other hand might not be good. Does something like this make sense? 

```suggestion
                    lookup.lookupAnalyzedClass(clazz, None) match {
                      case Some(ac) => apis.markExternalAPI(clazz, ac)
                      case _        => apis
                    }
```

zinc/src/sbt-test/source-dependencies/issue-810/build.json
eed3si9n: I prefer descriptive names. Something like `subproject-api-update`?

jtjeferreira:
@eed3si9n ready for a 2nd review

eed3si9n:
Overall LGTM. Could you squash the commits and provide a descriptive commit message explaining the background/motive for the changes plz?

I'll keep this open so others will also have the change to review.

dwijnand:
I haven't been following.  Is this to remove a log statement?

jtjeferreira:
> Overall LGTM. Could you squash the commits and provide a descriptive commit message explaining the background/motive for the changes plz?
> 
> I'll keep this open so others will also have the change to review.

Done

> I haven't been following. Is this to remove a log statement?

No. I updated the PR description

dwijnand (approved):
LGTM, thanks

eed3si9n:
@jtjeferreira Thanks for updating the commit!

@dwijnand I added a note to the description that might be relevant to provenance/quickAPI that you might want to double check. In the situation `core` adds a method that `app` does NOT use, no cycles are triggered for `app` and therefore the output JAR path remains the same, but the associated `Analysis` file will change, which means there will be 1-to-n relationship with the output JAR and `Analysis` file.

eed3si9n:
I'm guessing it _should_ be ok because the invariance of "different API hash means different JAR" would remain the same.

dwijnand:
Thanks for the heads up.  I might finally get access tomorrow, so going forward I'll be able to look at how these hooks are implemented (and what integration tests are present) so I'll rotate back to this.  The fix is good, I'm happy to merge it and forward fix how it interacts with 1-to-1/1-to-n.

internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
```diff
@@ -482,6 +485,12 @@ case class ProjectStructure(
       ()
     }
 
+  def checkExists(i: IncState, className: String, name: String): Future[Unit] =
```
eatkins, line 488: This name is a little vague. It would be more clear if it was something like checkNameExistsInClass. I found the test cases at the bottom confusing to read without looking at the source code.

eed3si9n: I agree now that you've mentioned it, but in the interest of reducing the barrier-to-merge I'm going to merge this as-is, and split your comment to a new issue.

PR #885, opened 2020-08-18
Fix "off-by-not" error in shouldDoEarlyOutput

dwijnand:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/885/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

eed3si9n:
The change looks ok, but it might be better to add a test in https://github.com/sbt/zinc/blob/develop/zinc/src/test/scala/sbt/inc/IncrementalCompilerSpec.scala?

dwijnand:
What tests were you thinking?  The honest way to observe this is if you have a custom implementation with the careful micro-managing, which we don't have a simulation of.

eed3si9n:
For this particular bug, it would've been caught with a simple lookup imple  that always returns `true`, so that's what I was thinking. Or you can do things like return true every other time, call compile, and see if early output is created.

PR #887, opened 2020-08-18
Finish wiring VirtualFileRef to lookupAnalyzedClass (aka "quickAPI")

dwijnand:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/887/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/Lookup.scala
```diff
@@ -76,7 +76,10 @@ trait ExternalLookup extends ExternalHooks.Lookup {
    *         Optional.of(analyzed) if analyzed.provenance.isEmpty => Couldn't find it.
    *         Optional.of(analyzed) => good
    */
-  def lookupAnalyzedClass(binaryClassName: String): Option[AnalyzedClass]
+  def lookupAnalyzedClass(
+      file: Option[VirtualFileRef],
+      binaryClassName: String,
+  ): Option[AnalyzedClass]
```
eed3si9n: I'd flip the parameter order here, since `file` is more or a hint.

Also I think it would be helpful to provide a documentation for what it intends to do when `Some(...)` is passed, for example in terms of how this relates to provenance. Speaking of which, provenance is a String should this be String also?

dwijnand: > I'd flip the parameter order here, since `file` is more or a hint.

Sure.

> Also I think it would be helpful to provide a documentation for what it intends to do when `Some(...)` is passed, for example in terms of how this relates to provenance.

That's up to the build tool.  The file here is based on the "classFile" for the given binary class name:

https://github.com/sbt/zinc/blob/e226ba97388d4addbde9298a1bc4040ffa5d3390/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala#L177-L191

which may be a class file or a jar file:

https://github.com/sbt/zinc/blob/e226ba97388d4addbde9298a1bc4040ffa5d3390/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala#L130-L145

Provenance is a more abstract, which is why it's just a string identifier.

eed3si9n (approved):

bishabosha:
@dwijnand do you know anywhere where the file parameter is actually used?

dwijnand:
Yeah, but in a private code base, which uses this extension.

bishabosha:
Ok thanks, may main question then is if I say had a tasty-only classpath, (e.g. with sbt pipelining) can I pass a non-existing class file to the binaryDependency callback? (which ends up calling this method) or perhaps it's appropriate to pass the tasty file instead?

Or maybe it's ok to pass a class file that's not there now because eventually it will be?

dwijnand:
Can't remember anything about these APIs, sorry.

PR #891, opened 2020-08-19
fix "return statement uses an exception to pass control ..." warning

jtjeferreira:
fix "return statement uses an exception to pass control to the caller of the enclosing named method"

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
@@ -11,7 +11,7 @@
 
 package xsbt
 
-import java.nio.file.{ Path, Paths }
+import java.nio.file.Path
 import xsbti.VirtualFile
 import xsbti.api.DependencyContext
 import DependencyContext._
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/891/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -72,13 +72,14 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
...
-          val sym = classOrModule.symbol
-          return Some(if (sym.isModule) sym.moduleClass else sym)
-        case _ => ()
+      val maybeClassOrModule = tree find {
+        case ((_: ClassDef) | (_: ModuleDef)) => true
+        case _                                => false
+      }
+      maybeClassOrModule.map { classOrModule =>
+        val sym = classOrModule.symbol
+        if (sym.isModule) sym.moduleClass else sym
```
dwijnand, line 81: collect over find/map...

jtjeferreira: I think you meant `collectFirst`, however `collectFirst` does not exist for `Tree`

dwijnand: Ah, fair enough (and yeah, collectFirst 😳)

PR #893, opened 2020-08-20
Avoid pickles being clobbered intra-cycles

dwijnand:
Previously merging the pickles updates jar only happened when the
compile cycle wasn't going to continue.  But that meant that over
multiple cycles scalac (given different sets of sources) was clobbering
the updates jar.  Now we make sure merge the jar every cycle.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/893/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #898, opened 2020-08-24
Cleanup some invalidation code

dwijnand:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/898/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -469,53 +469,35 @@ private[inc] abstract class IncrementalCommon(
...
+    val allInvalidations = (firstClassInvalidation -- recompiledClasses) ++ secondClassInvalidation
+    if (allInvalidations.isEmpty) {
       log.debug("No classes were invalidated.")
       Set.empty
     } else {
-      val allInvalidatedClasses: Set[String] = firstClassInvalidation ++ secondClassInvalidation
-      log.debug(s"Invalidated classes: ${allInvalidatedClasses.mkString(", ")}")
-      allInvalidatedClasses
+      log.debug(s"Invalidated classes: ${allInvalidations.mkString(", ")}")
+      allInvalidations
```
eed3si9n, line 497: Maybe the bug is here?
In the old code, it's checking for the diff, but returns `firstClassInvalidation ++ secondClassInvalidation`

dwijnand: Maybe.  Why does this have two different equations?

eed3si9n: It's checking if there's anything new that are being invalidated, but it returns both new and old invalidation together.

dwijnand: But the regression proves that the "old" invalidations aren't ignorable, no?  Therefore I think they shouldn't be ignored in that diff/if (i.e. do the opposite alignment that I went for)

eed3si9n: I think the regression indicates that you can ignore old when there's no new invalidation, but both old and new invalidations must be compiled together to avoid circular situation in some cases.

dwijnand: How does it indicate the former?

eed3si9n: The old-and-working code says:

```scala
    val newInvalidations = (firstClassInvalidation -- recompiledClasses) ++ secondClassInvalidation
    if (newInvalidations.isEmpty) {
      log.debug("No classes were invalidated.")
      Set.empty
```

dwijnand: OK, I don't think the regression indicates that you can ignore the old invalidations.

The question is will `firstClassInvalidation ++ secondClassInvalidation` ever be non-empty due to `recompiledClasses` alone?  That would suggests recompiledClasses invalidates itself, which sounds like an invariant that isn't going to happen.  So we can just drop the whole `-- recompiledClasses` business and align like that.

PR #878, opened 2020-08-12
Fix PickleJar to not clobber unmodified pickles

dwijnand:
Found a way to give scalac a unique file and then merge it with the previous one.

Now I just need to find a way to reproduce the original bug in a scripted test - any ideas?  It needs to somehow recognize that, for example, class A was compiled but the pickle jar should've contained the pickle for class B too.

PR #899, opened 2020-08-24
Don't auto-invalidate java sources every cycle

eatkins:
With the latest zinc, the second compilation cycle will always
invalidate all of the java sources in a given project resulting in
frequent overcompilation of these sources. This commit stops
automatically invalidating them. The invalidation logic was added in
0504f70454fdbe9db2cba4003c710888f6e31e2a. Two scripted tests needed to
be reverted to their pre-0504f70454fdbe9db2cba4003c710888f6e31e2a state
after this change.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
@@ -69,7 +69,8 @@ private[inc] abstract class IncrementalCommon(
       doCompile: CompileCycle,
       classfileManager: XClassFileManager,
       output: Output,
-      cycleNum: Int
+      cycleNum: Int,
+      isPipeline: Boolean,
   ) {
     def toVf(ref: VirtualFileRef): VirtualFile = converter.toVirtualFile(ref)
     def sourceRefs: Set[VirtualFileRef] = allSources.asInstanceOf[Set[VirtualFileRef]]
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/899/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

dwijnand:
While "all the java sources, irrespectively" is blunt, especially "every compilation cycle", it's required for by pipelining... for a reason that escapes me right now.  Hopefully it comes back after some morning caffeine.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -123,10 +123,6 @@ private[inc] abstract class IncrementalCommon(
       val nextInvalidations = result.nextInvalidations
       val current = result.analysis
 
-      val nextChangedSources: Set[VirtualFileRef] =
-        if (continue) javaSources
-        else Set.empty
```
eed3si9n: If we no longer have to do this, that would be great. This was a hack that was added to make Java symbols available to downstream Scala projects during pipelining. I wonder if this has been fixed via JAR merging implemented in https://github.com/sbt/zinc/pull/878?

eed3si9n (approved):

eatkins:
> While "all the java sources, irrespectively" is blunt, especially "every compilation cycle", it's required for by pipelining... for a reason that escapes me right now. Hopefully it comes back after some morning caffeine.

Ok. That makes sense. I updated CycleState to have an `isPipeline` parameter so that we only auto invalidate the sources if pipelining is enabled.

eatkins:
> If we no longer have to do this, that would be great. This was a hack that was added to make Java symbols available to downstream Scala projects during pipelining. I wonder if this has been fixed via JAR merging implemented in #878?

Could be. As mentioned in the comment above, I actually left the pipelined behavior in but we could easily remove it in the future for the pipelined case above. For now, I'm happy to leave it in since I have no idea what the consequences of not auto-invalidating are for pipelining.

dwijnand:
Could this follow #895, #896, and #898?  Those are all green and just waiting for a merge.

eatkins:
Sure. I'm not in a rush.

PR #901, opened 2020-08-24
Upgrade to scalatest 3.2.0

eatkins:
I was having problems compiling zinc-benchmarks with scala 2.13 that
seemed related to scalatest so I upgraded it. They changed the api in
3.1.0 so I had to run a scalafix task to update our existing tests:
https://github.com/scalatest/autofix/tree/master/3.1.x. After the
update, ZincBenchmarkSpec failed under 2.13 at runtime and I realized
that it wasn't actually testing anything, so I removed it.

Issue #579, opened 2018-08-17
Add an AuxiliaryClassFileManager to Zinc public API


allanrenucci:
From scala-js/scala-js#3419:

> Can we upstream a more generic version of https://github.com/lampepfl/dotty/blob/master/sbt-dotty/src/dotty/tools/sbtplugin/TastyFileManager.scala to Zinc so that scala-js and others can reuse it instead of reimplementing it?

I think we need something like:
```scala
abstract class AuxiliaryClassFileManager extends ClassFileManager {

  /** Given a list of class files, return the list of associated files
   *
   *  Associated files share the same life cycle as their class file.
   *  E.g. if a class file is deleted or restored, its associated files will
   *  be deleted or restored as well.
   */
  protected def associatedFiles(classes: Array[File]): Array[File]
}
```

sjrd:
Yes, that would be an ideal API for us in Scala.js.

(I think you can make `associatedFiles` protected rather than public)

sjrd:
Or perhaps even better: just give me one `.class` file at a time:
```scala
protected def associatedFiles(classFile: File): Array[File]
```

smarter:
@eed3si9n @dwijnand I'd like to bring attention to this issue again: ClassFileManager in 1.4 now uses VirtualFile and it's not clear to me if that still allows the sbt-dotty and sbt-scalajs plugin to correctly implement a custom ClassFileManager. https://github.com/sbt/zinc/pull/902 restores the old API as deprecated which should at least help with migration but doesn't seem like it's a good way forward.

PR #902, opened 2020-08-24
Restore ClassFileManager binary compatibility

eatkins:
With sbt 1.4.0-M2, the swoval scala.js project cannot compile because of
the removal of the file based delete method from ClassFileManager:
```
[error] ## Exception when compiling 27 sources to
${SWOVAL_LOCATION}/nio-js/.js/target/scala-2.12/classes
[error] java.lang.NoSuchMethodError: 'void
xsbti.compile.ClassFileManager.delete(java.io.File[])'
[error]
org.scalajs.sbtplugin.SBTCompat$$anon$1.delete(SBTCompat.scala:49)
```

To get the swoval build to work, I restored but deprecated the delete
and generate methods in ClassFileManger. I also had to add default
implementations of the VirtualFile versions of these methods. The
defaults work by converting the virtual files to regular files and
invoking the legacy apis that take file arguments.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/902/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
Builds on top of #901 which could be dropped in favor of this.

smarter:
> To get the swoval build to work, I restored but deprecated the delete
and generate methods in ClassFileManger.

Both sbt-dotty and sbt-scalajs (as you encountered) rely on these APIs, can they be properly reimplemented using the new VirtualFile stuff? Even better would be if zinc provided a higher-level API to just let the plugins declare the files that need to be deleted, as I pleaded for two years ago: https://github.com/sbt/zinc/issues/579

eatkins:
> Both sbt-dotty and sbt-scalajs (as you encountered) rely on these APIs, can they be properly reimplemented using the new VirtualFile stuff? Even better would be if zinc provided a higher-level API to just let the plugins declare the files that need to be deleted, as I pleaded for two years ago: #579

That would be great but it is beyond the scope of this PR. I simply want users (including myself) to be able to upgrade sbt versions without having to upgrade their sbt plugin versions.

smarter:
Right, I'll leave a comment in the original issue.

eed3si9n:
I thought I added https://github.com/sbt/zinc/pull/765 for Scala.JS. Did it not work?

eatkins:
It did not.

eatkins:
I use a pretty old version of scala.js in swoval: 0.6.26. I'm also using sbt-scalajs-crossproject 0.4.0.

eatkins:
I double checked with the latest sbt and zinc code in https://github.com/swoval/swoval cannot compile the scala.js project without this patch.

eed3si9n (approved):

PR #903, opened 2020-08-26
Make doc work with sbt 1.4.x

eatkins:
Scaladoc are not generated with 1.4.0-M2:
https://github.com/sbt/sbt/issues/5798. There seemed to be ~~three~~ two problems:
1. the sources were not actually specified in ScaladocBridge
2. Three was an assumption that the compiler would return only
   AbstractZincFiles in DelegatingReporter.makePosition but, in doc at
   least, it returns scala.reflect.io.PlainFile
3. ~~service loading doesn't seem to work reliably from sbt~~

~~Issues 1 and 2 are straightforward but issue 3 is more tricky. It isn't
clear to me what the assumptions are for service loading and why it
_sometimes_ works from sbt but not always. For example, in the sbt
project, the scripted actions/cross-multi-project successfully loads the
instance of CompilerInterface2 for scala 2.13.1 but not for 2.12.12. It
seems like maybe there is some kind of conflict between the sbt
metabuild loader and the scala instance class loader. At any rate, the
issue can be worked around if we use reflection to load
xsbt.CompilerBridge and xsbt.ScaladocBridge directly rather than
creating a dummy bridge and using reflection on the dummy bridge (at
least that is my interpretation of what the fallback code is doing). If
that fails, we fall back to the dummy bridge as it was before.~~

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/DelegatingReporter.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/DelegatingReporter.scala
@@ -17,6 +17,7 @@ import java.util.Optional
 import scala.reflect.internal.util.{ FakePos, NoPosition, Position }
 // Left for compatibility
 import Compat._
+import scala.reflect.io.PlainFile
 
 private object DelegatingReporter {
   def apply(settings: scala.tools.nsc.Settings, delegate: xsbti.Reporter): DelegatingReporter =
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/903/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
I'm not sure if this is the best way to fix this issue but scaladoc works in sbt 1.4.x after these changes.

eatkins:
If we modify the sbt classloading layers, service loading can work and it is not necessary to apply the changes to `AnalyzingCompiler.scala` to get doc working with sbt 1.4.x. I still think the changes might be worth making because the fallback code didn't seem to work before. That being said, I have no problem with re-pushing with only the changes to `DelegatingReporter.scala` and `ScaladocBridge.scala`.

eatkins:
Actually, I was wrong about service loading. The problem was that I had a stale compiler bridge jar in ~/.sbt/scripted/boot/fakeboot/scala-2.12.12. Once I removed that directory, service loading started working.

eed3si9n (approved):
Thanks for the fix!

PR #907, opened 2020-08-27
Invalidate java sources even on just class invalidations

dwijnand:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/907/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
```diff
@@ -335,6 +335,7 @@ case class ProjectStructure(
     (incO, sco)
   }
   val exportPipelining = incOptions.pipelining
+  val noOuroboros = incProps.getProperty("noOuroboros", "false").toBoolean
```
eatkins: Haha. I had to look ouroboros up. It's very appropriate. The only bad thing is that I'm not sure how you would figure out what it does without looking at where it's used in the code.

zinc/src/sbt-test/source-dependencies/java-class-invs/test
```diff
@@ -0,0 +1,4 @@
+> use/compile
+$ sleep 1000
```
eatkins, line 2: Why is the sleep necessary?

eed3si9n:
> I think this isn't the default setup in sbt, but I didn't check.

Self `classes` directory is definitely part of the classpath.

https://github.com/sbt/sbt/blob/1b99049fbca6439280631b800cc237a171301878/main/src/main/scala/sbt/Defaults.scala#L2345

```scala
      fullClasspath := concatDistinct(exportedProducts, dependencyClasspath).value,
```

What differentiates "full" classpath from dependency classpath is its exported products, and in a normal case the exported product is the `classes` directory.

Having the `classes` directory as part of the classpath is a fundamental building block to use _separate_ compilation for the purpose of _incremental_ compilation. In other words, given `A.scala`, `compile`, then later `B.scala`, you want `A.class` on the classpath without recompiling `A.scala`.  Ouroboro is not a good way of thinking about this, because if ClassfileManager is doing its job, it should evict any products created by the candidate source files before raw compilers are invoked. For example, if you make changes to `A.scala`, `A.class` should be moved to some other directory before `A.scala` compilation starts.

zinc/src/sbt-test/source-dependencies/java-class-invs/test
dwijnand: It's a last-modified-time-granularity thing.  There's a comment associated to the first instance of it in the test suite, which I've since shed when copying the pattern.  I've been thinking it should be handled intrinsically in scripted's compilation where, additionally, it can be both manually and globally tweaked locally as well as intelligently configured (like sbt.io.IO does, IIRC).

internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
dwijnand: Yeah, it's opaque.  But it's also terse and unique, so I was happy enough with it.

dwijnand:
OK, that makes sense, and thank you for the explanation.  Can you think of another way to test that an upstream class change causing a downstream recompilation should always include the java sources?

eed3si9n:
> Can you think of another way to test that an upstream class change causing a downstream recompilation should always include the java sources?

Maybe it would be good to open an issue to discuss the premise here? I don't see why Java sources needs to be invalidated even for pipelining.

I'm guessing that you're thinking of a scenario like this?

```scala
// core/Core.scala
object Core {
  def foo: Int = 0
  // def foo(x: Int): Int = x
}

// util/A.scala // uses Core.scala
...
// util/B.java
...

// app/App.scala // uses Core.scala and B.java
...
```

1. app/compile
2. Make a change to `core/Core.scala` that would trigger invalidation of `util/A.scala` and `app/App.scala`.
3. `app/compile` again. If it still works then it sees the symbols from `B.java` so we're all good?

dwijnand:
> Maybe it would be good to open an issue to discuss the premise here? I don't see why Java sources needs to be invalidated even for pipelining.

It's to pickle the Java APIs so they can be consumed downstream (via the early jar).

> so we're all good?

Yep, that worked a treat.  Thanks.

zinc/src/sbt-test/source-dependencies/java-class-invs/test
```diff
@@ -0,0 +1,6 @@
+> app/compile
+$ sleep 1000
+$ copy-file changes/Std.scala std/Std.scala
+$ delete lib/target/early/output.jar
+# delete early/output.jar to avoid it accidentally working by merging into an older jar
```
eed3si9n, line 5: So what this is saying is that the incremental compilation would work if `early/output.jar` is not magically deleted? If so we should account for that particular scenario instead of invalidating all Java sources in all cycles.

eed3si9n (changes requested):

zinc/src/sbt-test/source-dependencies/java-class-invs/test
dwijnand: Good point.

The difference between this working here and not working elsewhere is because zinc-scripted adds target/classes to the classpath.

There are suspicions that might not work well in pruning classfiles on java source deletions, but I need to write the test and see.  Drafting this in the meantime.

dwijnand:
Coming back to studying the difference in setups and whether it requires any adjustments within Zinc hasn't come up, so I'll drop this from your queue.

Issue #914, opened 2020-09-08
Compilation can fail with NoClassDefFoundError when low on memory


eatkins:
## steps

In the zinc project, using jdk8, G1GC and a max heap of 1150m run:
`sbt -J-Xmx1150m -Dsbt.version=1.4.0-RC1 -J-XX:+UseG1GC -J-verbose:gc "clean;publishLocal"`
## problem

```
[error] (compilerBridgeJVM2_12 / Compile / compileIncremental) java.lang.NoClassDefFoundError: xsbt/DelegatingReporter$CompileProblem
[error] (compilerBridgeJVM2_10 / Compile / compileIncremental) java.lang.NoClassDefFoundError: xsbt/DelegatingReporter$CompileProblem
[error] (compilerBridgeJVM2_10 / Compile / doc) java.lang.NoClassDefFoundError: xsbt/DelegatingReporter$CompileProblem
[error] (compilerBridgeJVM2_12 / Compile / doc) java.lang.NoClassDefFoundError: xsbt/DelegatingReporter$CompileProblem
[error] (compilerBridgeJVM2_11 / Compile / compileIncremental) java.lang.NoClassDefFoundError: xsbt/ExtractAPI$$anonfun$mkAnnotations$1$$anonfun$apply$1
[error] (compilerBridgeJVM2_11 / Compile / doc) java.lang.NoClassDefFoundError: xsbt/DelegatingReporter$CompileProblem
[error] Total time: 101 s (01:41), completed Sep 8, 2020 11:28:48 AM
```

## expectation

It should have OOMed.

## notes

Bumping the heap to 1250m, the following command works:
`sbt -J-Xmx1250m -Dsbt.version=1.4.0-RC1 -J-XX:+UseG1GC -J-verbose:gc "clean;publishLocal"`

eatkins:
I don't know what, if anything, we can do about this but I wanted to create the issue so that this behavior becomes known. I feel like I have seen weird NoClassDefFoundErrors before and never realized it could be due to insufficient memory.

smarter:
Interesting, this is probably the root cause of https://github.com/sbt/zinc/issues/750, is this a JVM bug?

eed3si9n:
[...]

PR #915, opened 2020-09-08
Hold reference to compiler bridge classloader

eatkins:
When sbt is low on memory, sometimes it will evict the classloader that
was constructed for the compiler bridge causing compilation to fail with
a mysterious NoClassDefFoundError. The compiler cache should not evict a
classloader that has any references to it. This commit reworks
AnalyzingCompiler so that we make an explicit reference to the compiler
bridge classloader and pass that loader into the methods that need to
access the classes in the loader instead of passing in the logger from
which we create an ephemeral loader.

After this change, I was unable to reproduce
https://github.com/sbt/zinc/issues/914

### DIFF
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/AnalyzingCompiler.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/AnalyzingCompiler.scala
@@ -83,14 +83,17 @@ final class AnalyzingCompiler(
     val progress = if (progressOpt.isPresent) progressOpt.get else IgnoreProgress
     val cp = classpath.map(converter.toPath)
     val arguments = compArgs.makeArguments(Nil, cp, options).toArray
+    // hold reference to compiler bridge class loader to prevent its being evicted
+    // from the compiler cache (sbt/zinc#914)
+    val loader = getLoader(log)
 
-    loadService(classOf[CompilerInterface2], log) match {
+    loadService(classOf[CompilerInterface2], loader) match {
       case Some(intf) =>
         intf.run(sources, changes, arguments, output, callback, reporter, progress, log)
       case _ =>
         // fall back to old reflection if CompilerInterface2 is not supported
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/915/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
I added printlns to the sbt CompilerCache eviction and found that it was specifically the compiler bridge loader that was being closed which was why I tried to make sure that we had a reference to the compiler bridge loader to prevent evictions.

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/AnalyzingCompiler.scala
```diff
@@ -83,14 +83,15 @@ final class AnalyzingCompiler(
     val progress = if (progressOpt.isPresent) progressOpt.get else IgnoreProgress
     val cp = classpath.map(converter.toPath)
     val arguments = compArgs.makeArguments(Nil, cp, options).toArray
+    val loader = getLoader(log)
```
eed3si9n, line 88:
```suggestion
    // hold reference to avoid the class loader getting collected by G1GC sbt/zinc#914
    val loader = getLoader(log)
```

eed3si9n: It might be worth adding a comment there to avoid future refactoring.

eatkins: 👍

eed3si9n (approved):
Thanks!

Issue #911, opened 2020-09-05
Compile cycles can loop indefinitely


eatkins:
## steps

* publish sbt with latest zinc and latest sbt code
* set `ThisBuild / SettingKey[Boolean]("usePipelining") := false` to a file local.sbt or modify build.sbt
* build https://github.com/sbt/sbt/pull/5812 (using `sbt -Dsbt.version=1.4.0-SNAPSHOT`)
* change Zinc version in `project/Dependencies.scala` to `1.4.0-SNAPSHOT`
* reload
* compile

## problem

main-settings project cycles indefinitely

```
[info]  published ivy to /Users/ethanatkins/.ivy2/local/org.scala-sbt/command_2.12/1.4.0-SNAPSHOT/ivys/ivy.xml                          
[info] compiling 9 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                     
[info] compiling 3 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                     
[info] compiling 11 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                    
[info] compiling 3 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                     
[info] compiling 11 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                    
[info] compiling 3 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                     
[info] compiling 11 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                    
[info] compiling 3 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                     
[info] compiling 11 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                    
[info] compiling 3 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                     
[info] compiling 11 Scala sources to /Users/ethanatkins/work/oss/sbt/sbt/main-settings/target/scala-2.12/classes ...                    
[...]

PR #916, opened 2020-09-09
Invalidate all sources after transitive step

eatkins:
If there are cyclic dependencies in files, it is possible to get in a
loop where we repeatedly recompile the same files. To avoid that, we can
just invalidate all of the sources if the cycle count exceeds the
IncOptions.transitiveStep parameter. This is a blunt tool but is better
than getting stuck in a loop.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
@@ -86,7 +86,12 @@ private[inc] abstract class IncrementalCommon(
       val invalidatedRefs: Set[VirtualFileRef] =
         mapInvalidationsToSources(classesToRecompile, initialChangedSources, sourceRefs, previous)
 
-      val invalidatedSources: Set[VirtualFile] = invalidatedRefs.map(toVf)
+      val invalidatedSources: Set[VirtualFile] = if (cycleNum > options.transitiveStep) {
+        debug(s"reached cycle limit ${options.transitiveStep}; invalidating all sources")
+        allSources
+      } else {
+        invalidatedRefs.map(toVf)
+      }
 
       val pruned = IncrementalCommon
         .pruneClassFilesOfInvalidations(invalidatedSources, previous, classfileManager, converter)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/916/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
Fixes #911

dwijnand (approved):
Sucks to not have a better understanding of the cause and/or a reproducing test.

But the hammer is worth it, considering how bad the bug is.

eed3si9n:
Has this always been the case, or did we regress on this because of pipelining invalidating Java etc?

eatkins:
It occurs to me that we could potentially avoid this problem if we tracked which files have already been compiled so we only ever invalidate a file once. Theoretically if there have been no additional source changes after the first cycle, I think we should still be able to use the class files generated during the first compilation run. But this is just an idea, I have no idea what might break in practice.

dwijnand:
I'd gladly take that change if it's truly safe, but it sounds too easy to be true.  Surely one can wire things in a complex enough way such that later validations cause files that have already been recompiled once to then fail to typecheck or refcheck, no?

eatkins:
You're probably right that things could theoretically break but my intuition is that would only be the case if there were really nasty cyclic dependencies between the files. Off the top of my head, I guess I could see something like this being a problem:
A.scala
```scala
object A {
  def foo = B.foo
  def bar: Int = 3
}
```
B.scala
```scala
object B {
  def foo = A.bar
}
```
So I suppose a lot of builds could potentially have incremental compilation regressions if such a scheme were introduced. On the other hand, it would be very nice if there was an incremental option `cyclic: Boolean` or something that would allow build authors to indicate that their project is acyclic so that it would be safe to only compile a source once. That could potentially be a good performance win.

It would be a nice complement to https://github.com/lihaoyi/acyclic.

eed3si9n (changes requested):
This is a pretty drastic measure for an incremental compilation, and it can have pretty significant downstream impact. (Like large codebase suddenly taking 30 min to compile stuff)
Before we do this, we should try to get it to reproduce the problem and understand what's happening. sbt 1.4.0-RC1 had some issues with previous values that was causing over-compilation, which hopefully I addressed in https://github.com/sbt/sbt/pull/5844

eatkins:
I disagree strongly. I have a very simple reproduction of the issue. It is real. I also see no good reason to ever do more than 3 cycles. At that point, I really think we should just throw up our hands and just recompile the whole project. Often 3 cycles even with just partial invalidations is slower than one cycle would have been. I also have never seen more than 3 cycles for a project in my own usage of sbt prior to discovering the looping issue.

Here is the repro:
1) publish a local version of zinc
2) checkout origin/develop in sbt/sbt
3) set `ThisBuild / SettingKey[Boolean]("usePipelining") := false` to a file local.sbt or modify build.sbt
4) clean; compile
5) change the version of zinc to 1.4.0-SNAPSHOT in project/Dependencies.scala
6) compile

After that last compile, zinc will loop forever in mainSettingsProj. I just tested this with fresh checkouts of both zinc and sbt.

eatkins:
Certainly an option would be to add an additional parameter to IncOptions called cycle limit or something if users really need a fourth or fifth cycle but a full invalidation after the transitive step seems like a reasonable default.

eed3si9n:
Thanks for the repro. I can reproduce it using sbt 1.4.0-RC1 as you noted, but I can't using sbt 1.3.13. The log looks like this:

### sbt 1.3.13

```
sbt:sbtRoot> compile
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] Compiling 1 Scala source to /private/tmp/sbt/testing/target/scala-2.12/classes ...
[info] Compiling 5 Scala sources to /private/tmp/sbt/main-command/target/scala-2.12/classes ...
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] Compiling 6 Scala sources and 1 Java source to /private/tmp/sbt/zinc-lm-integration/target/scala-2.12/classes ...
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] Compiling 1 Scala source to /private/tmp/sbt/run/target/scala-2.12/classes ...
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] Compiling 5 Scala sources to /private/tmp/sbt/main-actions/target/scala-2.12/classes ...
[info] Compiling 1 Scala source to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] Compiling 10 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] Compiling 11 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] Compiling 23 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] Compiling 132 Scala sources and 14 Java sources to /private/tmp/sbt/main/target/scala-2.12/classes ...
[info] Compiling 3 Scala sources to /private/tmp/sbt/sbt/target/classes ...
[info] Compiling 4 Scala sources to /private/tmp/sbt/scripted-sbt-redux/target/scala-2.12/classes ...
[success] Total time: 64 s (01:04), completed Sep 13, 2020 8:09:49 PM
```

so it's supposed to climb up from 1, 10, 11, then 23 Scala sources.

### sbt 1.4.0-RC1

```
sbt:sbtRoot> compile
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] compiling 6 Scala sources and 1 Java source to /private/tmp/sbt/zinc-lm-integration/target/scala-2.12/classes ...
[info] compiling 1 Scala source to /private/tmp/sbt/run/target/scala-2.12/classes ...
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] compiling 1 Scala source to /private/tmp/sbt/testing/target/scala-2.12/classes ...
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] compiling 5 Scala sources to /private/tmp/sbt/main-command/target/scala-2.12/classes ...
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] compiling 5 Scala sources to /private/tmp/sbt/main-actions/target/scala-2.12/classes ...
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] compiling 1 Scala source to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] compiling 9 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] compiling 3 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 11 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 3 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 11 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 3 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 11 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 3 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 11 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 3 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 11 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 3 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
[info] compiling 11 Scala sources to /private/tmp/sbt/main-settings/target/scala-2.12/classes ...
```

Meanwhile 1.4.0-RC1 is oscillating between 3 and 11 Scala sources instead of expanding the invalidation. This indicates that there's likely some regression that happened that is failing to actually do transitive invalidation.

eatkins:
well, you disproved my claim about 3 cycles.

eatkins:
Perhaps this is the wrong idea then. I don't really know how to debug the invalidation logic so I'll leave that to those who do.

eed3si9n:
I can take a look. It's entirely possible that given the current state of code this ends up becoming the solution, but I wanted to understand why.

eed3si9n:
<s>
I think I found the regression. In https://github.com/sbt/zinc/pull/898 `dependsOnClass` function value got refactored to be

```scala
    val dependsOnClass = findClassDependencies(_, relations)
```

but originally it was:

```scala
      if (invalidateTransitively) {
        val dependsOnClass = relations.memberRef.internal.reverse _
        ....
```

`memberRef.internal.reverse` is flip-the-arrow invalidation, so that's pretty aggressive.
</s>

https://github.com/sbt/zinc/pull/921#discussion_r487815968

eatkins:
I'm going to close this since it looks like #921 addresses the issue.

eed3si9n:
> It occurs to me that we could potentially avoid this problem if we tracked which files have already been compiled so we only ever invalidate a file once. Theoretically if there have been no additional source changes after the first cycle, I think we should still be able to use the class files generated during the first compilation run. But this is just an idea, I have no idea what might break in practice.

Ironically 1.4.0-RC1 maybe have implemented exactly that (returning only `newInvalidation`) in https://github.com/sbt/zinc/pull/898#discussion_r487889237 causing this looping issue.

PR #924, opened 2020-09-17
Upgrade Scalafmt, add NLs before+after methods

dwijnand:
Also remove FixScalafmtPlugin, whose root cause was fixed in
sbt-scalafmt 2.0.1.

Oh, and upgrade to sbt 1.3.13.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/924/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
Could we separate version bumps and style changes please?

dwijnand:
Sure.  Chronologically the version bump came after, as I had to upgrade to a version that had `newlines.topLevelStatements`, but I'm happy to split the commit.

internal/compiler-bridge/src/main/scala-2.11/xsbt/Compat.scala
```diff
@@ -16,6 +16,7 @@ import xsbti.compile.Output
 import scala.tools.nsc.Settings
 
 abstract class Compat
+
```
eed3si9n, line 19: What's the point of these blank lines?

eed3si9n:
> Chronologically the version bump came after, as I had to upgrade to a version that had newlines.topLevelStatements, but I'm happy to split the commit.

I guess we can first discuss the general style change and worry about bumping then. I'm generally not a huge fan of changing Scalafmt style since I'm ok with the existing style.

internal/compiler-bridge/src/main/scala/xsbt/API.scala
```diff
@@ -152,7 +155,7 @@ final class API(val global: CallbackGlobal) extends Compat with GlobalHelpers wi
           }
           val zincClassName = names.className
           val srcClassName = classNameAsString(symbol)
-          sourceVF foreach { source =>
+          sourceVF.foreach { source =>
```
eed3si9n, line 158: Was this done automatically? What's wrong with foreach being infix?

internal/compiler-bridge/src/main/scala/xsbt/DelegatingReporter.scala
```diff
@@ -51,20 +52,24 @@ private object DelegatingReporter {
...
-        case (Some(s), Some(l)) => s + ":" + l
-        case (Some(s), _)       => s + ":"
-        case _                  => ""
-      }
+
+    override def toString = (sourcePath0, line0) match {
+      case (Some(s), Some(l)) => s + ":" + l
+      case (Some(s), _)       => s + ":"
+      case _                  => ""
+    }
```
eed3si9n, line 60: I prefer the previous style here too.

internal/compiler-bridge/src/main/scala/xsbt/DelegatingReporter.scala
```diff
@@ -121,7 +126,8 @@ private object DelegatingReporter {
             // work around for https://github.com/scala/bug/issues/11865 by falling back to start pos
             case _: ArrayIndexOutOfBoundsException =>
               startLine
-          } else None
+          }
+        else None
```
eed3si9n, line 130: This looks good. `else` should be aligned with `if`.

internal/compiler-bridge/src/main/scala/xsbt/InteractiveConsoleFactoryBridge.scala
```diff
@@ -25,16 +26,16 @@ class InteractiveConsoleBridgeFactory extends xsbti.InteractiveConsoleFactory {
...
+    args,
+    bootClasspathString,
+    classpathString,
+    initialCommands,
+    cleanupCommands,
+    loader,
+    bindNames,
+    bindValues,
+    log
+  )
```
eed3si9n, line 39: Not a fan of this dangling.

internal/compiler-bridge/src/main/scala/xsbt/ClassName.scala
```diff
@@ -13,27 +13,19 @@ package xsbt
 
 import scala.tools.nsc.Global
 
-/**
- * Utility methods for creating (source|binary) class names for a Symbol.
- */
+/** Utility methods for creating (source|binary) class names for a Symbol. */
```
eed3si9n, line 16: Making short scaladoc one-line seems like a good idea. Not sure if it's worth doing massive code change for it, but this one I can get behind as more human-line.

eatkins:
Personally I think we should just accept what the defaults are because it's not worth the mental energy to worry about code formatting. I think we could pin to an older edition to upgrade the plugin but I just don't see that much upside in doing so. The biggest reason to avoid formatting changes is for the git history but git blame in sbt and zinc already doesn't really work because of all of the format changes and file moves over the years.

internal/compiler-bridge/src/main/scala/xsbt/ClassName.scala
eatkins: Funnily enough in light of my previous comment, this is a change I don't like because one-line scaladoc implicitly discourages documenting parameters. Even if your method initially doesn't have a parameter, if you add one then you have to completely change the scaladoc formatting from one-line to multiline.

eed3si9n:
We initially introduced Scalafmt because ppl thought sbt code base was difficult to read. That was one of the feedbacks we got in contributor feedback thing. So we applied Scalafmt. The point is that the formatting is meant to be as bland / vanilla as possible. In the beginning we debated to make the style more bland, because some of the Scalafmt defaults at the time didn't feel standard enough, but I'm fine with whatever we got for now.

eatkins:
I agree that the old sbt formatting was quite difficult to read. Like I said, I'm very pro code formatting. What is nice about the google java formatter is that they basically don't have any options and they also don't ever change the defaults. In my opinion, scalafmt changes the defaults way too often but it's still way better to use scalafmt than not. The reason I support just going with their defaults even though I don't like all of them is that I worry that if we pin to an edition, then it won't be supported in the future and I think it's good to stay on the latest version of the plugin whenever possible. Going with the defaults also means we won't argue about things like one-line scaladoc formatting. We can just agree that some of the defaults are annoying but we might not agree on which ones.

internal/compiler-bridge/src/main/scala-2.11/xsbt/Compat.scala
dwijnand: Clarity of code

internal/compiler-bridge/src/main/scala/xsbt/API.scala
dwijnand: It's distracting for some to be infix and others not, so I've just taken infix out of the picture.

internal/compiler-bridge/src/main/scala/xsbt/DelegatingReporter.scala
dwijnand: This is `newlines.beforeMultilineDef`.  I've gone with `fold`, but there's also `keep`, which I think is documented as "may be buggy", `classic`, which I assume will be removed eventually?, and `unfold` which would suck for many `= this match {` / `= foo match {` definitions.

So we can either take a risk by switching to `keep`, take another risk going with `classic`, or stick with `fold`.

Personally if a definition is fully defined in terms of a match (and that initial RHS fits on the RHS of `=`) then I think this looks neat.

internal/compiler-bridge/src/main/scala/xsbt/InteractiveConsoleFactoryBridge.scala
dwijnand: I assume you mean the `new InteractiveConsoleBridge(` folding up.  See the comment higher about the options here.

internal/compiler-bridge/src/main/scala/xsbt/ClassName.scala
dwijnand: I'm fine with that 1-off cost being paid whenever a method adds its first parameter rather than the continuous cost of 2 extra pointless (IMO) lines per method (opening and closing the scaladoc).

dwijnand:
Scalafmt's editions are dead, so that idea isn't worth thinking about (the migration for any die hard editions fans is to stick to older versions of scalafmt - which I guess means you can at least set updated on sbt-scalafmt versions).

I sweat the formatting more than most, I think, and it's held me back for years.  I was so happy when we were finally able to merge the formatting changes in sbt/sbt as it freed up my mind more to be able to actually look at the code instead of being constantly distracted by its formatting (not that the problem hasn't gone away completely).

I'm honestly jealous that the two of you can consider going with the defaults, but personally I don't always find the defaults picked to always make the code readable, which is why I've pushed for some of those defaults to be changed (reason why in this upgrade I could remove the `align.openParenCallSite`, `align.openParenDefnSite`, and `danglingParentheses` settings).  It's also why I look at the options to see if there's a better option that I find makes the code more readable.

Looking at the changes I picked, folding oneline scaladocs is probably the least important, so I can drop it if you prefer.

internal/compiler-bridge/src/main/scala/xsbt/API.scala
eed3si9n: I am generally in favor of the trend towards using standard dot-and-paren invocation style, rather than aggressive infix style. However, `foreach` taking a lambda on RHS is a clear case that _should_ be infixed as the DSL is mimicking to be a language word, and the end result is not even dot-and-paren.

eed3si9n:
It's not that I don't have preferences about styles (I clearly do), now that this is one of hobby projects that I have so little hours I can spend on, *debating about style changes* has net-negative productivity for me. I am strongly in favor of using the current version of Scalafmt as long as we can, and even when we update Scalafmt version and make style changes in the future, we do so only to counter Scalafmt default changes to preserve the existing sbt or Zinc code styling.

dwijnand:
It's a significant productivity boost for me to find a way to land this change.  I've rewritten, manually, different parts of this diff, multiple times, over the past 6 months (in local code) that I'm happy to finally have an automated way that doesn't impose any work on anyone (nor myself) and also applies the change consistently.

I'm happy to negotiate this in a call later in the week if that helps keep the cost down.

eatkins:
I didn't notice that the PR added rule changes and mistakenly thought that the changes were due to changes to the defaults in a newer version of scalafmt. That was my bad for not reading the actual commits. I was just responding to Eugene's comments.

We've had a number of debates about formatting including whether or not log messages start with capital letters. I finally realize that the bike shedding is more of an emotional drain than a productivity one for me. In the case of code formatting, I basically haven't thought about it since I started working on sbt which was after the initial scalafmt change. When I look at the old formatting, I am horrified though. I can relate to your frustration because I would have gotten very frustrated working on the old sbt code. With all that in mind, I'm trying to stick to a default position of I don't care about formatting unless it's inconsistent or I find it egregiously horrible (like the old sbt formatting). For example, I don't care for the one line doc strings but they're not awful. Since they're consistently applied, I don't strongly object to the change.

eatkins:
I also barely touch zinc so I have no basis for a strong objection.

internal/compiler-bridge/src/main/scala/xsbt/API.scala
dwijnand: One of the reason to always favour non-infix is that it makes it non-awkward to add something like `.headOption` at the end of it.  I guess with `foreach` that's different (if we ignore things like tap/pipe).  But outside of foreach I still stand by the consistency value of applying this universally.

I wouldn't be against you if you argued to the Scalafmt devs that `foreach` should be treated differently and got the default changed, or even made it as an option that you then preferred to have switched on.

dwijnand:
> mistakenly thought that the changes were due to changes to the defaults in a newer version of scalafmt

Oh a lot of them were due to that.  I initially attempted this upgrade a few weeks ago, straight to whatever the latest versions were then, and it was a terrible diff.  So then I waited till I had an open Friday afternoon (last week) and literally applied every single sbt-scalafmt and Scalafmt version bump in chronological release date order, reading the release notes to make a best attempt at preserving the existing format as well as migrating the various options changes.  Behaviour did change, Scaladocs are now formatted, other things that weren't always being applied (e.g. around infix) started to, things that were never being formatted (e.g. `... = this match {\n`) started to, etc, etc...  Then once I got to the latest version I squashed it all down so that the incremental nonsense went away (then re-split it into 2 commits, as preferred).

dwijnand:
Looking for a merge, or a path to discuss this...

eed3si9n:
I am interested in unified formatting across sbt submodules. What we have I think is good enough for the purpose of having standard-looking code base.

If there's a major pain/concern about the readability by new contributors etc, I'd to understand that first before jumping to the solution space.

dwijnand:
Sucks that we weren't able to find a compromise here that would have helped me contribute to Zinc.

PR #931, opened 2020-10-07
Fix #930: Report original problems in SourceInfos

### LINKED ISSUE #930: Reported problems are always empty on VirtualFiles
## steps

- Create an sbt project
- Add a warning inside a source file:
```scala
def foo() = ???
foo // deprecation warning on scala 2.13.3 with -deprecation flag 
```
- Print the reported problems in `compile.value`

```scala
compile.value.getAnalysis.readSourceInfos.foreach {
  case (_, infos) =>
    println(infos.getReportedProblems.size)
    println(infos.getUnreportedProblems.size)
}
```

## problem

The reported problems are empty

## expectation

The reported problems should contain the warning

## notes

The bug is located in the `AnalysisCallback` class:

https://github.com/sbt/zinc/blob/52c63b3a146b1e641897dda8055662311550d6c9/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala#L676-L690

Here `pos.sourcePath` contains an absolute path whereas in the following lines `src` contains a `VirtualFile` path.

https://github.com/sbt/zinc/blob/52c63b3a146b1e641897dda8055662311550d6c9/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala#L963-L967

All `reported` and `unreported` are lost when the `Analysis` is built.

Fixing this bug can help fixing:
https://github.com/sbt/sbt/issues/5937
https://github.com/sbt/sbt/issues/5950

adpi2:
Fixes #930 

`LoggedReporter.allProblems` return original problems so that the ` AnalysisCallback` can link each position to the corresponding `VirtualFile`

### DIFF
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/LoggedReporter.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/LoggedReporter.scala
@@ -136,7 +136,7 @@ class LoggedReporter(
       severity,
       InterfaceUtil.jo2o(rendered)
     )
-    allProblems += problem
+    allProblems += problem0
     severity match {
       case Warn | Error =>
         if (!testAndLog(transformedPos, severity)) display(problem)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/931/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

adpi2:
Should I do something to validate this against sbt?

eed3si9n:
@adpi2 Thanks for the fix.

@rtimush wrote in https://github.com/sbt/sbt/issues/5892#issuecomment-705169240:

> So, apparently what happens is that sourcePositionMapper converts virtualizes path into the real one, then the real path is used to build a map of compile problems in Incremental#problem but the lookup uses the virtualized path in Incremental#addProductsAndDeps.

It would be good if you could confirm that this patch would address https://github.com/sbt/sbt/issues/5892 by locally publishing zinc. I often use the following to bind SNAPSHOT Zinc against sbt to locally check:

```
sbt -Dsbtzinc.path=../zinc
```

rtimush:
I checked sbt/sbt#5892 with this PR and my test (https://github.com/sbt/sbt/compare/develop...rtimush:warnings-test) passed.

eed3si9n (approved):

PR #939, opened 2020-11-03
Optimize knownProducts for direct-to-JAR compilation

### LINKED ISSUE #933: Performance improvement ideas for Zinc 1.4.x
I'm profiling a build using the latest Zinc. I'll use this ticket to record the areas where Zinc could be improved.

## `JarUtils.ClassInJar`

The build is using straight-to-JAR compilation. This seems to be an allocation hotspot within:

```
   jo2o(output.getSingleOutputAsPath) match {
      case Some(so) if so.getFileName.toString.endsWith(".jar") =>
        knownProducts foreach { product =>
          new JarUtils.ClassInJar(product.id).toClassFilePath foreach { path =>
            ps.add(path.replace('\\', '/'))
          }
        }
```

The main optimization would be to avoid repeatedly calling `new File` in `ClassInJar.splitJarReference` on what I presume is always the same path (the path of the JAR), and is anyway discarded by `def toClassFilePath: Option[ClassFilePath] = splitJarReference._2`! 

`new File` internally incurs String allocations, at least on Windows, in `WinNTFileSystem.normalize`.

I am also inclined to remove use of parallel collections should be used in `knownProducts`.

This is also inefficient:
```
val Array(jar, cls) = toString.split("!")
```

retronym:
Fixes https://github.com/sbt/zinc/issues/933

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -1086,29 +1086,26 @@ private final class AnalysisCallback(
   private def knownProducts(merged: Analysis) = {
     // List classes defined in the files that were compiled in this run.
     val ps: java.util.Set[String] = new java.util.HashSet[String]()
-    val knownProducts: Vector[VirtualFileRef] =
-      merged.relations.allSources.toVector.flatMap(merged.relations.products)
+    val knownProducts: collection.Set[VirtualFileRef] = merged.relations.allProducts
 
     // extract product paths
-    jo2o(output.getSingleOutputAsPath) match {
-      case Some(so) if so.getFileName.toString.endsWith(".jar") =>
-        knownProducts foreach { product =>
-          new JarUtils.ClassInJar(product.id).toClassFilePathOrNull match {
-            case null =>
-            case path =>
-              ps.add(path.replace('\\', '/'))
-          }
+    val so = jo2o(output.getSingleOutputAsPath).getOrElse(sys.error(s"unsupported output $output"))
+    val isJarOutput = so.getFileName.toString.endsWith(".jar")
+    knownProducts foreach { product =>
+      if (isJarOutput) {
+        new JarUtils.ClassInJar(product.id).toClassFilePathOrNull match {
+          case null =>
+          case path =>
+            ps.add(path.replace('\\', '/'))
         }
-      case Some(so) =>
-        knownProducts foreach { product =>
-          val productPath = converter.toPath(product)
-          try {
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/939/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -1086,20 +1090,29 @@ private final class AnalysisCallback(
...
+    val ps: java.util.Set[String] =
+      if (AnalysisCallback.parallelKnownProducts)
+        java.util.concurrent.ConcurrentHashMap.newKeySet[String]
+      else
+        new java.util.HashSet[String]()
+    val knownProducts: Vector[VirtualFileRef] =
+      merged.relations.allSources.toVector.flatMap(merged.relations.products)
+
+    def knownProductsPar =
+      if (AnalysisCallback.parallelKnownProducts) new ParVector(knownProducts) else knownProducts
```
eed3si9n: Does this use the default fork/join scheduler? If so, could it end up competing against other tasks or potentially other Zinc compilation that are also trying to do the same thing? Or it finishes quickly enough that it won't surface? (But takes long enough time if it's not parallelized?)

eed3si9n: Looks like I added the parallel thing in  https://github.com/sbt/zinc/pull/849/commits/94ba81e908d3ec4838705125fbaee9156f2be496, but I'd be happy to see it gone if it's not working.

eed3si9n (approved):

internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
retronym: In this case I think the best idea is to keep it sequential. If this is a bottle neck, we can try to first reduce the overhead in sequential code (like I've done here, but I'm sure there are more opportunities.)  Sequential code is a bit easier to profile and tune. Using parallelism should be done with a careful understanding of whether the task size suitable for parallel collections (e.g. something like `(1 to 1000).par.count(_ % 2 == 0)` is slower than the sequential version.

I've added a commit to remove the `ParVector` altogether.

eed3si9n (approved):

PR #940, opened 2020-11-03
remove "duplicate compilations" assertion

johnduffell:
There was a new assertion introduced in https://github.com/sbt/zinc/pull/800 that there couldn't be more than one compilation with the same start time.

The start time appears to be intended as a millisecond time so it's perfectly possible that more than one can start in the same millisecond.

To further confuse matters, the failure message suggests that the compilation itself is duplicated rather than just the start time, however the start time is not a strong identifier as far as I can see.  So presumably this didn't match the intent at the time of implementation.

In practical terms this is preventing us from updating to sbt 1.4 ( as well as others https://github.com/sbt/sbt/issues/5911 by the looks), as the incremental play compiler seems to randomly trigger this assertion, presumably a race condition depending on compilation speed etc. https://github.com/guardian/support-frontend/pull/2808

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Compilations.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Compilations.scala
@@ -36,10 +36,6 @@ object Compilations {
 
 private final class MCompilations(val allCompilations: Seq[Compilation]) extends Compilations {
   // TODO: Sort `allCompilations` chronologically and enforce it in the Zinc API specification
-  assert(
-    { val xs = allCompilations.map(_.getStartTime); xs.distinct.size == xs.size },
-    "Created a Compilations with duplicate compilations"
-  )
 
   def ++(o: Compilations): Compilations = new MCompilations(allCompilations ++ o.allCompilations)
   def add(c: Compilation): Compilations = new MCompilations(allCompilations :+ c)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/940/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):
Thank you for taking action on this. LGTM from me.

jtjeferreira:
Nothing against removing the assertion, but one could also argue that the problem is in the play sbt plugin...

johnduffell:
@jtjeferreira true and I don't claim to be an expert!

From my reading, I think the issue is a combination of two things

1. play compiles all dependent projects as a dependency of itself - https://github.com/playframework/playframework/blob/316fbd61c9fc6a6081a3aeef7e773c8bbccd0b6b/dev-mode/sbt-plugin/src/main/scala/play/sbt/PlayCommands.scala#L61-L70 

2. the zinc `Compilation` object just gets the current time when it was constructed and puts it in the startTime field.
https://github.com/sbt/zinc/blob/6b357ae59a86a1238d78c4f808662e808747259b/internal/zinc-core/src/main/scala/sbt/internal/inc/Compilation.scala#L37

I think in a concurrent build system it doesn't make sense to use the current time stamp as a strong identifier, but I'm happy for there to be a more appropriate solution to the overall problem.

---

note that in the develop branch, the code I linked above has changed, but still uses the current millis time in a different bit of code https://github.com/sbt/zinc/blob/b969b72f3737b04399d43272a5af1587518d3c82/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala#L598

dwijnand:
That assertion is really handy to keep!  I agree with you that it's a bad identifier, but it's the only identifier that exists and it's used in places as an identifier.  So that still needs fixing...

johnduffell:
totally agree that the intent of the assertion is appropriate
consider using the whole tuple (both fields together) in those places as an identifier.  Then we could reinstate this check on the whole product rather than just one field.

PR #948, opened 2020-11-27
Prepare DualLoader for Scala 3

adpi2:
Change the implementation of `DualLoader` to fix the `LinkageError` exception that occurs in the Scala 3 compiler when the `xsbti.*` classes are loaded. The problem is that those classes are loaded by two different `ClassLoaders`: the sbt loader and the scala instance loader. See https://github.com/lampepfl/dotty/blob/master/sbt-bridge/src/xsbt/CompilerClassLoader.java for more details about the origin of the problem.

In the fixed implementation, the compiler classes are loaded by the `DualLoader` itself, which extends `URLClassLoader`, so that the subsequent `xsbti.*` classes can be loaded by the underlying `dual` loader which is the sbt loader. 

This fix is needed to implement the new `CompilerInterface2` in the Scala 3 compiler bridge.

### Caching and Performance

@smarter said
> The changes to DualLoader need to be carefully inspected since ClassLoader are always tricky business and the way they are cached in sbt/zinc is also quite intricate (and if classloader caching doesn't work right, it might still work but slow things done as the JIT ends up recompiling the same classes again and again, it helps to run the JVM with -verbose:class to verify that multiple calls to compile for example do not end up reloading the same compiler classes multiple times.)

The cache is not affected here: instances of `DualLoader` are still indexed by the list of jars in the scala instance + the bridge jar.

The thing that could be problematic here is that we do not rely on `scalaInstance.loader` anymore but we build a new `URLClassLoader` containing `scalaInstance.allJars`.

### Regarding Binary Compatibility

I modified the  `DualLoader` constructor but, since it is in the internal package, it should rarely impact a downstream project. There are at least 2 breakage:
- in a deprecated method of sbt that can easily be fixed by merging sbt/sbt#6172
- in the ugly piece of code in `sbt-dotty` that casts and hacks the ClassLoader. This fix is replacement of this hack.

I found no usage of `DualLoader` in Bloop nor Mill.

Also I removed `DifferentLoaders` and `NullLoader` because I found no usage in Zinc, sbt, Bloop and Mill.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -597,6 +597,11 @@ lazy val zincClasspath = (projectMatrix in internalPath / "zinc-classpath")
         "sbt.internal.inc.classpath.NativeCopyConfig.*"
       ),
       exclude[IncompatibleMethTypeProblem]("sbt.internal.inc.classpath.NativeCopyConfig.*"),
+      exclude[MissingTypesProblem]("sbt.internal.inc.classpath.DualLoader"),
+      exclude[DirectMissingMethodProblem]("sbt.internal.inc.classpath.DualLoader.this"),
+      exclude[IncompatibleMethTypeProblem]("sbt.internal.inc.classpath.DualLoader.this"),
+      exclude[MissingClassProblem]("sbt.internal.inc.classpath.DifferentLoaders"),
+      exclude[MissingClassProblem]("sbt.internal.inc.classpath.NullLoader")
     ),
   )
   .jvmPlatform(scalaVersions = scala212_213)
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/948/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

adpi2:
Closing this one because I think https://github.com/lampepfl/dotty/pull/10541 is more promising.

PR #950, opened 2020-11-27
Don't initialize symbols before finding classfile

### LINKED ISSUE #949: Zinc inspects more symbols than necessary
#### build.sbt
```scala
lazy val a = project
  .in(file("a"))

lazy val b = project
  .in(file("b"))
  .settings(
    Compile / dependencyClasspath += (a / Compile / classDirectory).value
  )

lazy val c = project
  .in(file("c"))
  .settings(
    Compile / dependencyClasspath += (b / Compile / classDirectory).value
  )
```

#### a/A.scala
```scala
package a
class A
````

#### b/B.scala
```scala
package b

object B {
  def foo(x: Option[a.A]): Unit = ()
}
```

#### c/C.scala
```scala
package c

object C {
  def bar(): Unit = b.B.foo(None)
}
```

#### scalac.sh
```sh
#!/bin/zsh
rm -rf out
mkdir -p out/{a,b,c}
scalac a/A.scala -d out/a
scalac b/B.scala -cp out/a -d out/b
scalac c/C.scala -cp out/b -d out/c
```

## problem

This compiles with scalac (using `scalac.sh`), but not with sbt:
```
$ sbt a/compile b/compile c/compile
[info] welcome to sbt 1.4.3 (Twitter Java 1.8.0_242)
[info] loading settings for project global-plugins from plugins.sbt ...
[info] loading global plugins from /Users/mduhem/.sbt/1.0/plugins
[info] loading settings for project strict-deps-build-build from metals.sbt ...
[info] loading project definition from /Users/mduhem/Desktop/strict-deps/project/project
[info] loading settings for project strict-deps-build from metals.sbt ...
[info] loading project definition from /Users/mduhem/Desktop/strict-deps/project
[success] Generated .bloop/strict-deps-build.json
[success] Total time: 0 s, completed Nov 27, 2020 11:22:30 AM
[info] loading settings for project strict-deps from build.sbt ...
[info] set current project to strict-deps (in build file:/Users/mduhem/Desktop/strict-deps/)
[success] Total time: 0 s, completed Nov 27, 2020 11:22:30 AM
[info] compiling 1 Scala source to /Users/mduhem/Desktop/strict-deps/b/target/scala-2.12/classes ...
[success] Total time: 2 s, completed Nov 27, 2020 11:22:33 AM
[info] compiling 1 Scala source to /Users/mduhem/Desktop/strict-deps/c/target/scala-2.12/classes ...
[error] /Users/mduhem/Desktop/strict-deps/c/C.scala:4:22: Symbol 'type a.A' is missing from the classpath.
[error] This symbol is required by 'value b.B.x'.
[error] Make sure that type A is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
[error] A full rebuild may help if 'B.class' was compiled against an incompatible version of a.
[error]  def bar(): Unit = { b.B.foo(Option.empty); () }
[error]                      ^
[error] one error found
[error] (c / Compile / compileIncremental) Compilation failed
[error] Total time: 0 s, completed Nov 27, 2020 11:22:33 AM
```

## expectation
The code should compile with both Zinc and scalac.

Duhemm:
Previously, we would complete symbols before trying to find the
associated classfile. This is problematic for symbols where we don't
have the classfile (for instance, when compiling with Pants'
`strict_deps` enabled).

This initialization was introduced in #758, to workaround a scalac bug
where the associated classfile of a symbol wouldn't be set before the
symbol is completed. This bug has been fixed in Scala 2.12.12 and later
(see scala/scala#8889).

Fixes #949

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
@@ -175,7 +175,7 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with 
               isSyntheticCoreClass(targetSymbol)
           )
           if (!noByteCode) {
-            classFile(targetSymbol.initialize) match {
+            classFile(targetSymbol) match {
               case Some((at, binaryClassName)) =>
                 // Associated file is set, so we know which classpath entry it came from
                 processExternalDependency(binaryClassName, at)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/950/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -175,7 +175,7 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
               isSyntheticCoreClass(targetSymbol)
           )
           if (!noByteCode) {
-            classFile(targetSymbol.initialize) match {
+            classFile(targetSymbol) match {
```
eed3si9n, line 178: Nice catch!

eed3si9n (approved):
I'm personally ok with this especially this fixes the so-called strict dependency compilation that [Bazel](https://docs.bazel.build/versions/master/user-manual.html#flag--strict_java_deps) and [Pants](https://v1.pantsbuild.org/3rdparty_jvm.html#strict-dependencies) implement to work around the "omg 100 JARs on my ClassPath" problem.

retronym (approved):

dotta:
@Duhemm Could you elaborate on why this change is safe to do?

Asking because the compiler-bridge sources are always compiled against the specific version of Scala in use on a project, and hence there are no guarantees that 2.12.12+ is being used. I guess I'm missing something...

retronym:
New zinc with old scalac will still be correct, just a bit inefficient as
it will fall into the slow path more often.

PR #956, opened 2020-12-07
Add AuxiliaryClassFiles in IncOptions

### LINKED ISSUE #579: Add an AuxiliaryClassFileManager to Zinc public API

adpi2:
Fixes #579 

Add an `AuxiliaryClassFiles` member in `IncOptions` so that the `ClassFileManager` manages the files that are associated to class files. This will be used for Scala 3 (`.tasty` files), for Scala.js (`.sjsir` files) and for Scala Native (`.nir` files).

The `auxiliaryClassFiles` member in `IncOptions` is an array so that we can manage several type of files at the same time (for instance `.tasty` and `.sjsir` files). Its default value is an empty array because Scala 2 and Java don't need any.

The implementation for `.tasty` files is provided so that it can already be reused by sbt (https://github.com/sbt/sbt/issues/6080) and also by Mill and Bloop.

### DIFF
--- a/internal/compiler-interface/src/main/java/xsbti/compile/AuxiliaryClassFiles.java
+++ b/internal/compiler-interface/src/main/java/xsbti/compile/AuxiliaryClassFiles.java
@@ -1,0 +1,16 @@
+package xsbti.compile;
+
+import java.nio.file.Path;
+
+/**
+ * Interface that associates a class file with other corresponding produced
+ * files, for instance `.tasty` files or `.sjsir` files.
+ *
+ * This class is meant to be used inside the [[ClassFileManager]] to manage
+ * auxiliary files.
+ *
+ * ClassFileManagers are forgiving on auxiliary files that do not exist.
+ */
+public interface AuxiliaryClassFiles {
+    Path[] associatedFiles(Path classFile);
+}
--- a/internal/zinc-core/src/main/java/xsbti/compile/TastyFiles.java
+++ b/internal/zinc-core/src/main/java/xsbti/compile/TastyFiles.java
@@ -1,0 +1,17 @@
+package xsbti.compile;
+
+/**
+ * Can be added to `IncOptions.auxiliaryClassFiles` so that the TASTy files are
+ * managed by the ClassFileManager.
+ */
+public class TastyFiles extends AuxiliaryClassFileExtension {
+    private static final TastyFiles _instance = new TastyFiles();
+
+    public static TastyFiles instance() {
+        return _instance;
+    }
+
+    private TastyFiles() {
+        super("tasty");
+    }
+}
--- a/zinc/src/test/scala/sbt/inc/AuxiliaryClassFilesSpec.scala
+++ b/zinc/src/test/scala/sbt/inc/AuxiliaryClassFilesSpec.scala
@@ -1,0 +1,29 @@
+package sbt.inc
+
+import sbt.io.IO
+import xsbti.compile.AuxiliaryClassFiles
+
+import java.nio.file.Path
+
+class AuxiliaryClassFilesSpec extends BaseCompilerSpec {
+  it should "allow client to add define their own auxiliary class files" in {
+    IO.withTemporaryDirectory { tempDir =>
+      val setup = ProjectSetup.simple(tempDir.toPath, SourceFiles.Foo :: Nil)
+      var callbackCalled = 0
+      val myAuxiliaryFiles = new AuxiliaryClassFiles {
+        override def associatedFiles(classFile: Path): Array[Path] = {
+          callbackCalled += 1
+          Array()
+        }
+      }
+
+      val compiler =
+        setup.createCompiler(incOptions = incOptions.withAuxiliaryClassFiles(Array(myAuxiliaryFiles)))
+
+      try compiler.doCompile()
+      finally compiler.close()
+
+      callbackCalled.shouldEqual(3)
+    }
+  }
+}
--- a/zinc/src/test/scala/sbt/inc/AuxiliaryClassFilesSpec.scala
+++ b/zinc/src/test/scala/sbt/inc/AuxiliaryClassFilesSpec.scala
@@ -1,29 +1,68 @@
 package sbt.inc
 
+import sbt.internal.inc.FileAnalysisStore
 import sbt.io.IO
-import xsbti.compile.AuxiliaryClassFiles
+import sbt.io.syntax._
+import xsbti.compile.{ AnalysisStore, ScalaNativeFiles, TastyFiles, TransactionalManagerType }
 
-import java.nio.file.Path
+import java.nio.file.Files
 
 class AuxiliaryClassFilesSpec extends BaseCompilerSpec {
-  it should "allow client to add define their own auxiliary class files" in {
-    IO.withTemporaryDirectory { tempDir =>
-      val setup = ProjectSetup.simple(tempDir.toPath, SourceFiles.Foo :: Nil)
-      var callbackCalled = 0
-      val myAuxiliaryFiles = new AuxiliaryClassFiles {
-        override def associatedFiles(classFile: Path): Array[Path] = {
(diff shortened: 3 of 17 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/956/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
Could you exercise this from https://github.com/sbt/zinc/blob/develop/zinc/src/test/scala/sbt/inc/IncrementalCompilerSpec.scala or scripted test somehow?

smarter:
Thanks for working on this!

> The auxiliaryClassFiles member in IncOptions is an array so that we can manage several type of files at the same time (for instance .tasty and .sjsir files). Its default value is an empty array because Scala 2 and Java don't need any.
> The implementation for .tasty files is provided so that it can already be reused by sbt (sbt/sbt#6080) and also by Mill and Bloop.

I  wonder if besides .tasty we shouldn't also hardcode .sjsir and .nir by default, it shouldn't break anything and it means that mill/bloop/maven/... don't have to remember to do this themselves. /cc @sjrd

internal/compiler-interface/src/main/java/xsbti/compile/AuxiliaryClassFiles.java
```diff
@@ -0,0 +1,14 @@
+package xsbti.compile;
+
+import java.nio.file.Path;
+
+/**
+ * Interface that associate a class file with other corresponding produced
+ * files, for instance `.tasty` files or `.jsir` files.
```
sjrd:
```suggestion
 * Interface that associates a class file with other corresponding produced
 * files, for instance `.tasty` files or `.sjsir` files.
```

internal/zinc-core/src/main/java/xsbti/compile/TastyFiles.java
```diff
@@ -0,0 +1,23 @@
...
+ * managed by the ClassFileManager
+ */
+public class TastyFiles implements AuxiliaryClassFiles {
+    @Override
+    public Path[] associatedFiles(Path classFile) {
+        if (classFile.endsWith(".class")) {
+            String absolutePath = classFile.toAbsolutePath().toString();
+            String prefix = absolutePath.substring(0, absolutePath.length() - 6);
+            Path tastyFile = Paths.get(URI.create(prefix + ".tasty"));
+            return new Path[] { tastyFile };
```
sjrd: This implementation indiscriminately returns a `Path` for a `.tasty` file even if it doesn't exist. Not all `.class` files have a corresponding `.tasty` file (only top-level classes have one). Is this expected, i.e., is the client supposed to ignored returned `Path`s that do not exist?

sjrd:
> I wonder if besides .tasty we shouldn't also hardcode .sjsir and .nir by default, it shouldn't break anything and it means that mill/bloop/maven/... don't have to remember to do this themselves.

We could do that. I don't have a strong opinion either way. If Zinc doesn't do it by default, sbt-scalajs and sbt-scala-native will keep doing it.

internal/zinc-core/src/main/java/xsbti/compile/TastyFiles.java
adpi2: Yes I think it is easier if the `ClassFileManager` is forgiving on auxiliary files that don't exist.
Also, as far as I can see, the `ClassFileManager` implementations are already forgiving on class files that don't exist.

I add a comment and a test case.

zinc/src/test/scala/sbt/inc/AuxiliaryClassFilesSpec.scala
```diff
@@ -0,0 +1,29 @@
+package sbt.inc
+
+import sbt.io.IO
+import xsbti.compile.AuxiliaryClassFiles
+
+import java.nio.file.Path
+
+class AuxiliaryClassFilesSpec extends BaseCompilerSpec {
+  it should "allow client to add define their own auxiliary class files" in {
```
sjrd:
```suggestion
  it should "allow client to define their own auxiliary class files" in {
```

internal/zinc-core/src/main/java/xsbti/compile/TastyFiles.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public class TastyFiles implements AuxiliaryClassFiles {
+    @Override
+    public Path[] associatedFiles(Path classFile) {
+        // not all classfiles have a corresponding TASTy file but we strongly assume that the
+        // ClassFileManager is forgiving on files that do not exist.
+        String absolutePath = classFile.toAbsolutePath().toString();
+        if (absolutePath.endsWith(".class")) {
+            String prefix = absolutePath.substring(0, absolutePath.length() - 6);
+            Path tastyFile = Paths.get(prefix + ".tasty");
```
sjrd: This implementation will only work for `Path`s that represent the real file system. It would be better, IMO, to only use methods of `Path` that work for arbitrary `Path`s. Perhaps the following:
```suggestion
        String fileNameString = classFile.getFileName().toString();
        if (fileNameString.endsWith(".class")) {
            String fileNamePrefix = fileNameString.substring(0, fileNameString.length() - 6);
            Path tastyFile = classFile.resolveSibling(fileNamePrefix + ".tasty");
```

sjrd: The same comment applies to the other implementations of `AuxiliaryClassFiles`, obviously.

eed3si9n (approved):

internal/zinc-core/src/main/java/xsbti/compile/ScalaJsFiles.java
```diff
@@ -0,0 +1,22 @@
+package xsbti.compile;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Can be added to `IncOptions.auxiliaryClassFiles` so that the sjsir files
+ * produced by the Scala.js compiler are managed by the ClassFileManager.
+ */
+public class ScalaJsFiles implements AuxiliaryClassFiles {
```
sjrd: I don't know what's the convention of this repo for acronyms in class names, but in Scala.js we typically `ScalaJS`, for example `ScalaJSPlugin`. Under that convention, it would be nice for this to be called `ScalaJSFiles`. However, that may conflict with the conventions of this repo, which should prevail.

sjrd (approved):
Overall LGTM. I have a few comments.

zinc/src/test/scala/sbt/inc/AuxiliaryClassFilesSpec.scala
adpi2: this test does not exist anymore because it was useless

internal/zinc-core/src/main/java/xsbti/compile/TastyFiles.java
adpi2: I have applied those changes and reduced the code duplication by introducing a `AuxiliaryClassFileExtension` parent class

PR #410, opened 2017-09-21
Add thread-safe implementation of AnalysisCallback

dotta:
This commit simply turns all mutable data structures into concurrent ones.

As we already had a thread-safe implementation of AnalysisCallback, I thought it was easier to share the code rather than commenting on the related PR (#409).

PR #957, opened 2020-12-18
Made AnalysisCallback thread-safe again. Re sbt/sbt#6198

dotta:
For the parallel Hydra Scala compiler to work with sbt and zinc, it's paramount
that the AnalysisCallback is thread-safe. The bulk of the work was done in
sbt/zinc#410, but the changes introduced to support pipelined compilation were
not thread-safe, hence could potentially break compilation via Hydra.

There are three categories of concurrency hazards that are fixed:

* Memory visibility issues (`@volatile` helps with this)
* Compiler/JVM optimizations leading to instructions re-ordering (`@volatile
  prevents this from happening in the sections of interest)
* Race conditions due to concurrent read/write (`AtomicBoolean` CAS heals this)

Mind that instead of `@volatile`, it was perfectly fine to use
`AtomicReference`/`AtomicBoolean`, but I opted for `@volatile` in the places
where CAS were not required. The rationale is to apply KISS.

I've tested this by building a local sbt with zinc produced out of this PR, and tested Hydra compilation in sbt 1.4 works. I don't believe there is a need for testing performance. The reason is that the performance impact of this PR is so irrelevant that is essentially impossible to observe.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -16,6 +16,7 @@ package inc
 import java.io.File
 import java.nio.file.{ Files, Path, Paths }
 import java.util.{ EnumSet, UUID }
+import java.util.concurrent.atomic.{ AtomicBoolean }
 import sbt.internal.inc.Analysis.{ LocalProduct, NonLocalProduct }
 import sbt.internal.inc.JavaInterfaceUtil.EnrichOption
 import sbt.util.{ InterfaceUtil, Level, Logger }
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/957/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

dotta:
\cc @dragos

eed3si9n (approved):
Thanks!

PR #969, opened 2021-03-23
Reuse a TreePrinter for less allocation and lock contention

retronym:
ExtractAPI stringifies trees in annotation argument position.

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
@@ -202,6 +202,42 @@ class ExtractAPI[GlobalType <: Global](
   private def reference(sym: Symbol): xsbti.api.ParameterRef =
     xsbti.api.ParameterRef.of(tparamID(sym))
 
+  // Constructing PrintWriters can cause lock contention in highly parallel code,
+  // it's constructor looks up the "line.separator" system property which locks
+  // on JDK 8.
+  //
+  // We can safely reuse a single instance, avoiding the lock contention and
+  // also reducing allocations a little.
+  private object ReusableTreePrinter {
+    import java.io._
+    private val buffer = new StringWriter()
+    private val printWriter = new PrintWriter(buffer)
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/969/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

retronym:
This is the profile that motivated the change:

image

image

internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
```diff
@@ -215,7 +252,10 @@ class ExtractAPI[GlobalType <: Global](
             xsbti.api.Annotation.of(
               processType(in, a.atp),
               if (a.assocs.isEmpty)
-                Array(xsbti.api.AnnotationArgument.of("", a.args.mkString("(", ",", ")"))) // what else to do with a Tree?
+                Array(
+                  xsbti.api.AnnotationArgument
+                    .of("", ReusableTreePrinter.mkString(a.args, "(", ",", ")"))
+                ) // what else to do with a Tree?
```
smarter, line 257: Hash it! Or at least that's what I do: https://github.com/lampepfl/dotty/blob/d4f1c261f808b185f91bb3e6f3efebd25c39646f/compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala#L619-L695 :)

retronym: Nice. It would be preferable for the compiler directly hash much more than it already does, but as you point out the test discovery feature would still need to find annotated members.

eed3si9n (approved):
Thanks @retronym!
Merging this now so it'll make it to 1.5.0.

PR #973, opened 2021-04-06
fix ScalaInstance constructor

### LINKED ISSUE #972: ScalaInstance constructor deprecated in 1.5.0 looks broken
## steps

Check code: https://github.com/sbt/zinc/blob/v1.5.0/internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala#L61-L62

## problem

`this` constructor is called without using named parameters. It looks like some parameters are inverted and the compilerJar parameter is lost.

## expectation

It think it should be:

```scala
    this(
      version,
      loader,
      loader,
      loaderLibraryOnly,
      Array(libraryJar),
      Array(compilerJar),
      allJars
      explicitActual
    )
```

adpi2:
Fix #972

### DIFF
--- a/internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala
+++ b/internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala
@@ -58,8 +58,8 @@ final class ScalaInstance(
       loader,
       loaderLibraryOnly,
       Array(libraryJar),
-      allJars,
-      Array.empty,
+      compilerJars = allJars,
+      allJars = allJars,
       explicitActual
     )
   }


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=pull/973/head&io=develop&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala
```diff
@@ -58,8 +58,8 @@ final class ScalaInstance(
       loader,
       loaderLibraryOnly,
       Array(libraryJar),
-      allJars,
-      Array.empty,
+      compilerJars = allJars,
+      allJars = allJars,
```
adpi2, line 62: We don't use `compilerJar` because it cannot be used alone. But it should be included in `allJars`.

adpi2: @slandelle

eed3si9n (approved):

PR #999, opened 2021-08-11
Force the type of an annotation's typeSymbol before checking isStatic

### LINKED ISSUE #998: False invalidation by inheritance on JDK 11
## steps

  - clone https://github.com/lrytz/sbtproj
  - use JDK 11 (OpenJDK 64-Bit Server VM Temurin-11.0.12+7)
  - sbt
    - clean
    - compile
    - debug
    - un-commment the `println` in A.scala
    - compile

`A` is recompiled, then the log shows
```
[debug] Invalidated by transitive inheritance dependency: Set(B)
```
and `B` is recompiled in a second round.

Doesn't happen on JDK 8.

lrytz:
Before 2.13, `isStatic` is implemented by
`isNonBottomSubClass(StaticAnnotationClass)`
which forces the annotation symbol's type.

In 2.13, Java annotations are identified by flags. This check doesn't
force the info, and the flags are missing if the info is still a
`ClassfileLoader`.

This leads to spurious API changes (annotation goes missing) if the
depending if the info is already forced or not.

A fix for this will be in 2.13.7, but we should still work around it
in Zinc to make sure it works correctly on 2.13.0-6.

Fixes https://github.com/sbt/zinc/issues/998

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
@@ -838,9 +838,14 @@ class ExtractAPI[GlobalType <: Global](
       }
       implicit def compat(ann: AnnotationInfo): IsStatic = new IsStatic(ann)
 
-      // scala/bug#11679 annotations of inherited members may be absent from the compile time classpath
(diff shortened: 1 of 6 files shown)


eed3si9n (approved):
Thanks!

retronym (approved):

PR #1018, opened 2021-12-21
compute hash using DigestInputStream for larger files

### LINKED ISSUE #1015: HashUtil.farmHash leads to OOME with large files
## steps
Checkout https://github.com/JetBrains/intellij-scala
Remove this line:
https://github.com/JetBrains/intellij-scala/blob/ae3cb5fe8b1c03ed2e4d3c5c4631e8e62e5ef4be/build.sbt#L27
`sbt compile`

## problem
Sbt allocates byte arrays in `HashUtil.farmHash(path: Path)` method. In case of many concurrent tasks and large files it may lead to OOME:

[error] ## Exception when compiling 0 sources to C:\Users\Nikolay.Tropin\IdeaProjects1\intellij-scala2\pluginXml\target\scala-2.13\classes
[error] ## Exception when compiling 30 sources to C:\Users\Nikolay.Tropin\IdeaProjects1\intellij-scala2\scala\dfa\target\scala-2.13\classes
[error] java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:481)
[error] java.lang.OutOfMemoryError: Java heap space
[error] java.lang.OutOfMemoryError: Java heap space
[error] java.base/java.nio.file.Files.read(Files.java:3221)
[error] java.base/java.nio.file.Files.read(Files.java:3221)
[error] java.base/java.util.concurrent.ForkJoinTask.getThrowableException(ForkJoinTask.java:600)
[error] java.base/java.nio.file.Files.readAllBytes(Files.java:3278)
[error] java.base/java.nio.file.Files.readAllBytes(Files.java:3278)
[info] shutting down sbt server
[error] sbt.internal.inc.HashUtil$.farmHash(HashUtil.scala:24)
[error] sbt.internal.inc.HashUtil$.farmHash(HashUtil.scala:24)
java.lang.OutOfMemoryError: Java heap space
Error during sbt execution: java.lang.OutOfMemoryError: Java heap space

## expectation
No crash 😄 

## notes
This version produces the same hash without copying file content to the heap. It also seem to work faster, at least for large files.
```
import java.nio.ByteBuffer
import java.nio.channels.FileChannel
import java.nio.file.StandardOpenOption

def farmHash(buffer: ByteBuffer): Long =
  LongHashFunction.farmNa().hashBytes(buffer)

def farmHash(path: Path): Long =
  farmHash(mappedBuffer(path))

private def mappedBuffer(path: Path): ByteBuffer = {
  val fileChannel = FileChannel.open(path, StandardOpenOption.READ)
  fileChannel.map(FileChannel.MapMode.READ_ONLY, 0, fileChannel.size())
}
```

niktrop:
Fixes https://github.com/sbt/zinc/issues/1015

### DIFF
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/HashUtil.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/HashUtil.scala
@@ -13,13 +13,24 @@ package sbt
 package internal
 package inc
 
-import java.nio.file.{ Files, Path }
 import net.openhft.hashing.LongHashFunction
+
+import java.nio.file.{ Files, Path }
 
 object HashUtil {
   def farmHash(bytes: Array[Byte]): Long =
     LongHashFunction.farmNa().hashBytes(bytes)
 
-  def farmHash(path: Path): Long =
-    farmHash(Files.readAllBytes(path))
+  def farmHash(path: Path): Long = {
+    import sbt.io.Hash
+
+    //allocating many byte arrays for large files may lead to OOME
+    //but it is more efficient for small files
+    val largeFileLimit = 10 * 1024 * 1024
+
+    if (Files.size(path) < largeFileLimit)
+      farmHash(Files.readAllBytes(path))
+    else
+      farmHash(Hash(path.toFile))
+  }
 }


lightbend-cla-validator:
Hi @niktrop,

[https://www.lightbend.com/contribute/cla](https://www.lightbend.com/contribute/cla)

internal/zinc-compile-core/src/main/scala/sbt/internal/inc/HashUtil.scala
```diff
@@ -13,13 +13,24 @@ package sbt
...
+    import sbt.io.Hash
+
+    //allocating many byte arrays for large files may lead to OOME
+    //but it is more efficient for small files
+    val largeFileLimit = 10 * 1024 * 1024
+
+    if (Files.size(path) < largeFileLimit)
+      farmHash(Files.readAllBytes(path))
+    else
+      farmHash(Hash(path.toFile))
```
eed3si9n, line 34: Would this basically stop checking for hash? I don't know if 10MB is good enough threshold for this if for example JAR could change with the same file name (SNAPSHOT etc). Maybe include file size since you grabbed it?

eed3si9n: Also I am not sure if grabbing file name in general is a good idea since the point of using content hash (as opposed to using last modified) was to make this machine independent and repeatable, and the absolute path could change from machine to machine.

niktrop: `sbt.io.Hash.apply` computes SHA of a file content using DigestInputStream. Maybe it's worth to copy implementation here to be more explicit.

eed3si9n: My mistake. So the difference basically would be that this would use SHA instead of Farmhash.

eed3si9n (approved):

PR #1019, opened 2022-01-10
Upgrade util to 1.6.1

slandelle:
util 1.6.0 depends on log4j 2.17.0 which is vulnerable.

util 1.6.1, targeting 2.17.1, was released and sbt was upgraded, but not zinc itself.


eed3si9n (approved):

gabrieljones:
Will this be cherry picked to 1.8.
I had to switch to 1.8 due to https://github.com/sbt/zinc/issues/1158

PR #1034, opened 2022-02-04
centralize & modernize JVM settings & increase heap size

SethTisue:
this repo uses sbt-projectmatrix for crossbuilding which means it can be very heap-hungry if you do anything across all subprojects. I think it's reasonable to go to the 2G heap size for local development, not just on CI like before.

I've also removed `-Xms` as I doubt it's helping anything

I've also removed `-server` (this is surely left over from some ancient config, it's the default these days)

and I've removed the `CodeCacheSize` setting which I'm less sure about but I suspect is also just an ancient holdover


SethTisue:
@eed3si9n I'm, like, preeeeeetty sure these are all good changes, but lmk what you think

eed3si9n (approved):
LGTM

These memory settings predate GitHub Actions, so tbh I am not sure if they are still needed.

PR #1037, opened 2022-02-13
fixes so we can build on JDK 17

SethTisue:
newer sbt-assembly version has the necessary ASM upgrade, is the main change

main motivation: I'd like to be able to do my local development work on 17. but also, it's good for the CI in this repo, and the Scala 2 community build too, to be able to run the tests on JDK 17, just in case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,9 @@ jobs:
           - os: ubuntu-latest
             java: 11
             jobtype: 1
+          - os: ubuntu-latest
+            java: 17
+            jobtype: 1
           - os: windows-latest
             java: 8
             jobtype: 2
--- a/build.sbt
+++ b/build.sbt
@@ -206,6 +206,18 @@ lazy val zinc = (projectMatrix in (zincRootPath / "zinc"))
       BuildInfoKey.map(compilerBridge213 / Compile / classDirectory)("classDirectory213" -> _._2),
     ),
     Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat,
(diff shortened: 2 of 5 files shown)


SethTisue:
@eed3si9n I noticed that `runBenchmarks` was only running on 11, so I changed to run on 17 only. not sure if that's best or whether it should run on both. I'm guessing latest-only is probably okay, but I have zero knowledge of the benchmarking setup in this repo or any history there might be behind the choice

SethTisue:
to benchmark on JDK 17, I had to move the scala/scala reference SHA to a newer one that actually compiles on JDK 17. that would potentially throw off any apples-to-apples comparison of benchmark results a little, but I'm guessing it doesn't really matter?

eed3si9n:
> that would potentially throw off any apples-to-apples comparison of benchmark results a little, but I'm guessing it doesn't really matter?

Yea. We're not really tracking long-term benchmark, it's more of a sanity check in case in case we are intending perf improvement in a PR or later observe some regression we would have some way of benchmarking it.

SethTisue:
I think I'm going to just leave the benchmarks on JDK 11. It was turning into a time sink.

SethTisue:
I ended up just leaving the benchmarks on JDK 11. It was turning into a time sink.

This is ready for merge, as far as I can see.

also validated at https://github.com/scala/community-build/pull/1535

eed3si9n (approved):

dwijnand:
> the Javadoc tool is now picky about going straight to h3 without having h2 first

So weird.  Can't choose to use a relatively smaller header without?  😕

PR #1063, opened 2022-02-13
Update sbt-scalafmt and scalafmt to latest versions

SethTisue:
~I split this up into multiple commits to aid review. One commit has the code reformatting, the other has the reformatting of the Scaladoc comments.~

The code changes seem acceptably minimal to me.

~The Scaladoc changes are more sweeping, but I think it's a one-time big-bang we should accept. There's rarely a need to git-blame and so forth on Scaladoc comments, anyway.~

Going forward, I don't think we necessarily need to take every little Scalafmt version bump if there is resulting churn in the formatting. But the version we were on before had become really outdated, so I think it's worth making this big jump forward.


.scalafmt.conf
```diff
@@ -9,8 +9,7 @@ newlines.source = keep
 
 # http://docs.scala-lang.org/style/scaladoc.html recommends Asterisk style (aka JavaDoc style)
 # scala/scala is written that way too https://github.com/scala/scala/blob/v2.12.2/src/library/scala/Predef.scala
-# TODO: switch to Asterisk
-docstrings.style = keep
+docstrings.style = Asterisk
```
eed3si9n:
```suggestion
docstrings.style = Asterisk
docstrings.wrap = no
```

I am annoyed with each Scalafmt version randomly contaminating changes into git history long after Scalafmt was first introduced. I am often too lazy to find out the backward compatible configuration incantation to keep the source code as-is, but `docstrings.wrap = no` I think is necessary because the change it makes is not only unnecessary, but often unergonomic, like turning neatly ordered list into:

```
1.
1.
1.
```

eed3si9n (changes requested):

.scalafmt.conf
SethTisue: Okay I dropped that commit.

eed3si9n (approved):

SethTisue:
> I am annoyed with each Scalafmt version randomly contaminating changes into git history long after Scalafmt was first introduced

I'll keep this in mind and seek to strongly prefer stability going forward, now that we've established a new baseline. Thanks for merging.

PR #890, opened 2020-08-19
Add means to shim out sbt-project-matrix for IntelliJ import

retronym:

Issue #1064, opened 2022-02-14
CachedHashingSpec is flaky


SethTisue:
I've seen it fail several times in recent CI runs.

```
[info] CachedHashingSpec:
[3683](https://github.com/sbt/zinc/runs/5176868906?check_suite_focus=true#step:6:3683)
[info] zinc
[3684](https://github.com/sbt/zinc/runs/5176868906?check_suite_focus=true#step:6:3684)
[info] - should cache jar generation *** FAILED ***
[3685](https://github.com/sbt/zinc/runs/5176868906?check_suite_focus=true#step:6:3685)
[info]   18 was not less than 2.5 Cache jar didn't work: 18 is >= than 50% of 5. (CachedHashingSpec.scala:91)
```

similar previous issue: #552 with fix #553; maybe even more leniency is needed?

eed3si9n:
To make the test more robust and also make it a better guard against performance regressions, an improved approach might be to run it a multiple times (like JMH) and compare the p50 (median). If n = 1 occasionally fails, that could mean that the improvement gap is within the variance range. But also since it's mostly passing, it also hopefully means that n doesn't have to be large to smoke out the p50 differences consistently, like n = 5, or n = 11 should be ok.

SethTisue:
PR: #1077 (taking a somewhat simpler approach than suggested)

PR #1075, opened 2022-03-16
remove no-longer-needed IntelliJ workaround

SethTisue:
the workaround suggested in build.sbt for importing the build into IntelliJ doesn't work anymore (the commented-out line doesn't compile, if uncommented) with the current versions of things (not sure when it broke or why)

this is my attempt to do better and it seems to work. I'm not sure yet if it's perfect, but it's a clear improvement on the status quo. we can always adjust further later, as needed

replaces #890


retronym (approved):

SethTisue:
the CI failure is a known-flaky test (#1064)

project/plugins.sbt
```diff
@@ -13,3 +13,4 @@ addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.11.0")
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.8.1")
 addSbtPlugin("com.eed3si9n" % "sbt-projectmatrix" % "0.9.0")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "1.2.0")
+addSbtPlugin("org.jetbrains.scala" % "sbt-ide-settings" % "1.1.1")
```
lrytz, line 16: TIL!

PR #1079, opened 2022-03-17
recompile downstream when Java annotation changes

### LINKED ISSUE #630: Adding @Retention(RetentionPolicy.RUNTIME) to a java annotation does not correctly recompile downstream classes
```ini
sbt.version=1.2.6
```

```java
package test;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;

//@Retention(RetentionPolicy.RUNTIME)
public @interface Test {
}
```

```java
package test;
@Test
public class Foo{
  public static void main(String[] args){
    System.out.println(Foo.class.getAnnotations().length);
  }
}
```

```bash
$ sbt run
[info] Compiling 2 Java sources to /Users/lihaoyi/test/target/scala-2.12/classes ...
0
```

```bash
$ vim Test.java
$ cat Test.java
package test;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;

@Retention(RetentionPolicy.RUNTIME)
public @interface Test {
}
$ sbt run
[info] Compiling 1 Java source to /Users/lihaoyi/test/target/scala-2.12/classes ...
0
$ sbt clean
$ sbt run
[info] Compiling 2 Java sources to /Users/lihaoyi/test/target/scala-2.12/classes ...
1
```

SethTisue:
fixes #630


SethTisue:
Jason thinks the approach sounds good and notes,

> Looks like Java annotations all have the same parent interface:
>> The direct superinterface type of an annotation interface is always
>> java.lang.annotation.Annotation (§9.1.3).
>
> So we don't need to do any sort of transitive invalidation

SethTisue:
My initial implementation temporarily, kludgily overloaded the `hasMacro` flag to mean "has macro declaration _or annotation definition_", just as a quick way to get the tests passing.

The theme for today was to figure out what to do instead. I initially assumed that I would need to add an `hasAnnotation` flag to `AnalyzedClass`, alongside the existing `hasMacro` flag, and that would mean changing both the contraband and the protobuf.

But after discussion with @dwijnand and @retronym, we decided that would be overkill. We suspect that the `hasMacro` flag exists because the invalidation situation with macros is more complex: there is the definition and the separate implementation, and when the implementation changes we need to treat that (unlike the body of an ordinary method changing) as an API change.

But here the situation is simpler, so they suggested we could just have `detectAPIChanges` in `IncrementalCommon` directly check the parents to see if `java.lang.annotation.Annotation` is present.

And that approach seems to be working out.

SethTisue:
I think this is ready to go, but there's a customer that might like to test it before merge.

SethTisue:
@dwijnand I've review-requested you, but also, do you think there's anyone else who should or might like to take a look?

dwijnand (approved):

SethTisue:
(rebased, without any changes)

eed3si9n:
I am gonna merge this so it'll be part of 1.7.0.

SethTisue:
regression report at https://github.com/sbt/sbt/issues/6969 — I'll work on a fix pronto

PR #1011, opened 2021-12-13
Updated log4j version due to a vulnerability fix

hanikhan:
Updated log4j version due to a vulnerability fix: https://www.theverge.com/2021/12/10/22828303/log4j-library-vulnerability-log4shell-zero-day-exploit

PR #1093, opened 2022-04-21
remove explicit log4j dependency from zincScripted

SethTisue:
as Eugene has noted on e.g. #1011, we get our log4j from sbt-util. it appears to me the lines removed here aren't needed

@eed3si9n you added these lines as part of #601 — it isn't clear to me what your motivation might have been at the time. it's far enough back that perhaps things were different then, I don't know

the lines don't _actually_ result in an outdated/insecure version of log4j being used (except perhaps during testing, where log4j-slf4j-impl-2.16.0.jar was brought in, but surely that JAR isn't where any insecurity is, and anyway it was only being used in `zincScripted` which isn't published and is only used during test)

but regardless, the _appearance_ that an insecure dependency _might_ be used is worth correcting

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -656,7 +656,6 @@ lazy val zincScripted = (projectMatrix in internalPath / "zinc-scripted")
     ideSkipProject := true, // otherwise IntelliJ complains
     publish / skip := true,
     name := "zinc Scripted",
-    libraryDependencies ++= log4jDependencies,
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Sounds good to me.
Eventually we're probably going to drop log4j in general from sbt/sbt, but we'll deal with this when it happens.

PR #1099, opened 2022-05-10
move scalafmt to a separate CI job

SethTisue:
it's annoying otherwise


SethTisue:
I tested this by temporarily pushing a commit with bad formatting, and it was in fact caught.

PR #1174, opened 2023-03-05
Removes system error on empty name hashes

jackcviers:
* Returning an empty array passes all tests and scripted tests
* Will prevent the error in https://github.com/sbt/sbt/issues/7157

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -958,7 +958,7 @@ private final class AnalysisCallback(
         NameHashing.merge(nm1, nm2)
       case (Some(nm), None) => nm
       case (None, Some(nm)) => nm
-      case (None, None)     => sys.error("Failed to find name hashes for " + className)
+      case (None, None)     => Array.empty
     }
   }
 


internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -958,7 +958,7 @@ private final class AnalysisCallback(
         NameHashing.merge(nm1, nm2)
       case (Some(nm), None) => nm
       case (None, Some(nm)) => nm
-      case (None, None)     => sys.error("Failed to find name hashes for " + className)
+      case (None, None)     => Array.empty
```
eed3si9n, line 961: Zinc not finding named hashes for the class or object is an error state as far as Zinc is concerned.
To put it in analogy, if a new car on the road is hitting a warning sign that says "low bridge!", the fix should be to fix the car or reroute, not remove the warning sign.

eed3si9n (changes requested):

internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
jackcviers: @eed3si9n Thank you for getting back to me. I agree. The correct fix is to ensure that dotty enters the hashes. I mentioned that in the referenced issue.

Please correct me from here to `-- End of correct me if I'm wrong portion.`

It appears that the issue is that the `sbt-api` phase in dotty occurs before any macros or, in our case, source to source transformations are run:

```
calac -Xshow-phases
WARNING: Using incubator modules: jdk.incubator.concurrent
parser
typer
inlinedPositions
sbt-deps
extractSemanticDB
posttyper
prepjsinterop
sbt-api <here's where ExtractApi lives>
SetRootTree
pickler
inlining
postInlining
staging
<our compiler plugin phase runs here, but it probably doesn't really need to.>
pickleQuotes
...
```

The way I read this, is that if you generate classes after the `sbt-api` phase, the classes that get generated will show up in the compiler outputs, but not be hashed by the `sbt-api` compiler phase. Thus, with macro annotations (and any other class generating compiler plugin or phase that occurs after `sbt-api` on a compilation unit), it is now legal to have class outputs in scalac that are, according to zinc, unhashed. The analysis callback hasn't been called after their generation. Despite this now being legal in dotty scalac, having an unhashed class symbol, even if completely synthetic, is a fatal system error in zinc. Can you explain why an unhashed class symbol is fatal? I assume it is due to performance and not correctness, but I could be very wrong.

I don't know that moving the `sbt-api` phase later in the compiler is valid. We can certainly try it.

I do know that if you set the array to empty when you don't have a hash for a class, classes generated after `sbt-api` has run still are compiled, and classes, modules, and symbols already in the source and created by the front-end of the compiler are reported  in the `sbt-api` phase, and they have hashes, so incremental compilation works as normal for them.

Generated classes after `sbt-api` have no hash. My assumption is that synthetic class symbols will change only when source changes. Those source changes will be detected via `sbt-api` calling the analysis callback, and thus their recompilation is valid and won't cause performance degradation in zinc in that case. They will cause performance degradation when the source code does not change, as the synthetic class will be identical on every compile, and thus be recompiled itself when it does not need to be recompiled.

The base assumption here is that not having a hash for a synthetic class does not result in an *invalid* compilation result, only a *slower* incremental compilation. It did not appear to cause invalid results in my testing on our class generating codebase, but that is relatively small when compared to everything zinc compiles on a daily basis, so that assumption needs validation.

// End of correct me if I'm wrong portion.

In our use-case we could probably move our transformations before `sbt-api`. 

That may not be the case for macro annotations. I can also go into dotty and try moving the `sbt-api` phase, or making the `sbt-api` phase run after the annotation macros as well as where it currently runs. The hashes in are supposed to be stable, so that might work -- it could also just fail terribly.

If it doesn't work, we can try to insert a completely new phase that calls the zinc analysis callback for a compilation unit with the synthetic class symbols after all the phases have run, ensuring they have a hash. The sbt callback is stored on the context and fetchable from that. I don't believe it is removed from the context store at any time.

If you or anyone else have any other ideas, please, speak up. You can find me in the sbt discord under the same name too.

eed3si9n: Zinc is more complicated part of sbt, and I don't claim to know everything, but I gave a talk giving overview of Zinc's general strategy, which hopefully can explain what namehash is used for
https://m.youtube.com/watch?v=h8ACmUHQ2jg

In general, incremental compiler *not* invalidating classes will result in undercompilation, meaning code that should've been recompiled due to code change are not compiled, and can result to people shipping bad JAR files.

jackcviers: Ok. If you think undercompilation is a potential problem, then we have to call the sbt callback later in dotty.

The sbt-api phase does not appear to be movable from its current execution position in my experiments yesterday in its current state. I will first add some scripted tests to this PR to confirm that undercompilation is happening when generated classes are created after the sbt-callvack is executed, and that check generated classes are not making it into jars during packaging.

Then I'll move to dotty to try running the sbt-api callback in later phases with zinc integration, probably with some modifications to the ExtractApi class that backs the phase.

I'll also talk to the Scala 3 issue spree team and try getting this on the next issue spree.

eed3si9n:
Closing this for now.

PR #1206, opened 2023-06-17
Fix writing zip64's offset

dwijnand:
ZipFileSystem's `sync` writes all the zip file entries, followed by the
central directory ("cen") entries, counting all the `written` bytes,
which it passes to write the end ("end") entry.  ZipCentralDir's `dump`
skips writing the file entries and only writes the cen entries, which
means that its `written` is missing all the bytes of the file entries.
Fortunately that missing value is recorded in the `end` as the offset
where the cen entries start, so we fix `written` with `cenoff`.  This
bug only impacted writing zip64 jars, through ZipCentralDir.

### DIFF
--- a/internal/zinc-classfile/src/main/java/sbt/internal/inc/zip/ZipCentralDir.java
+++ b/internal/zinc-classfile/src/main/java/sbt/internal/inc/zip/ZipCentralDir.java
@@ -111,7 +111,7 @@ public class ZipCentralDir {
         }
         end.centot = elist.size();
         end.cenlen = written;
-        end.write(os, written);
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @dwijnand,

We see that you have signed the Lightbend Contributors License Agreement before, however, the CLA has changed since you last signed it.
Please review the new CLA and sign it before we proceed with reviewing this pull request:

[https://www.lightbend.com/contribute/cla](https://www.lightbend.com/contribute/cla)

eed3si9n (approved):
Thanks @dwijnand!

PR #1208, opened 2023-06-25
use `-release` instead of `-target`

xuwei-k:
fix warning

> [warn] -target is deprecated: Use -release instead to compile against the correct platform API.


eed3si9n (approved):
Thanks!

PR #1228, opened 2023-07-27
Handle javac warning messages

### LINKED ISSUE #1188: Java options warning not handled
## steps

setup a build that causes javac to give a warning on an option setting.

In my case - I was using JDK 19 + Scala + javac option: "-source 17" which gives the following on compilation...

`warning: [options] system modules path not set in conjunction with -source 17`

Note there is no file or line number in this warning.

## problem

The `JavaSpecParser` can't parse this type of warning and so returns:
`Unexpected javac output: warning: [options] system modules path not set in conjunction with -source 17`

The `JavaSpecParser` seems to ignore all subsequent errors if there are any.

This makes it quite difficult to work out why my app isn't compiling when I do have errors as Metals doesn't report any errors or highlight any issues with code.  The only way I can tell that my app isn't compiling (and why) is to look in the Metals logs.

## expectation

Ideally Zinc would interpret these kind of warnings and then go on to report any further warnings/errors.

`JavaSpecParser` seems to realise when it can't interpret an error.  I'm wondering whether it would be better to report an error in this case.  Maybe even with something like `Zinc cannot interpret following error XXX - raise an issue`.  Currently, not reporting an error on failed parsing means compilation can fail silently.  Or will that cause more problems than it fixes?

Arthurm1:
Javac warnings on bad options are currently not handled so subsequent errors/warnings in code are ignored.

Fixes #1188

### DIFF
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavaErrorParser.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavaErrorParser.scala
@@ -243,6 +243,15 @@ class JavaErrorParser(relativeDir: File = new File(new File(".").getAbsolutePath
           s"javac:$error"
         )
     }
+  val javacWarning: Parser[Problem] =
+    WARNING ~ SEMICOLON ~ restOfLine ^^ {
+      case _ ~ _ ~ warning =>
+        new JavaProblem(
+          JavaNoPosition,
+          Severity.Warn,
+          s"javac:$warning"
+        )
+    }
 
   val outputSumamry: Parser[Unit] =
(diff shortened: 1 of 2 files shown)


Arthurm1:
I've now signed the CLA

eed3si9n (approved):
@Arthurm1 Thanks for the contribution!

eed3si9n:
CONTRIBUTING I think is out of date. CLA that's required is Scala CLA https://www.lightbend.com/contribute/cla/scala

Arthurm1:
@eed3si9n I think I've signed the right one now.  https://www.lightbend.com/contribute/cla/scala/check/Arthurm1 comes back with true

PR #1235, opened 2023-08-11
Delete orphan class files during incremental compilation

### LINKED ISSUE #1234: Zinc fails to delete stale .class files resulting in broken incremental compilation
## steps

I have written up a repository with instructions to reproduce: https://github.com/jackkoenig/zinc-bug-example

They're written for Linux x86_64 or MacOs x86_64, the only platform-specific aspect is that you need 2 versions of the JVM, but it's easy enough to tweak the instructions to your liking so long as you properly set `JAVA_HOME` and `PATH` to use the differing versions as described in the instructions.

## problem

**TLDR:** Zinc is not deleting `.class` files from a previous compilation, I think the `.class` file deletion logic when you have an empty `Analysis` is to blame.

See Summary (https://github.com/jackkoenig/zinc-bug-example#summary) and Commentary (https://github.com/jackkoenig/zinc-bug-example#commentary) in my linked repository: 

## expectation

Zinc should properly delete stale `.class` files so that incremental compilation will succeed.

jackkoenig:
During incremental compilation, it is possible for there to be .class files in the output classes directory that are unknown to the previous Analysis. These files can break incremental compilation. Now, Zinc will recursively walk the output directory and delete any .class files unknown to the previous Analysis ("orphan" class files) before starting the incremental compile.

I suspect this PR isn't _quite right_ but I think the test is at least decent and I think the basic logic is probably right although perhaps this sort of functionality should be folded into the `Analysis` object--I'm not sure.

Fixes #1234

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -404,6 +404,17 @@ object Incremental {
     val hasSubprojectChange = initialChanges.external.apiChanges.nonEmpty
     val analysis = withClassfileManager(options, converter, output, outputJarContent) {
       classfileManager =>
+        val nDeleted =
+          IncrementalCommon.deleteOrphanClassFiles(
+            sources,
+            previous,
+            classfileManager,
+            converter,
+            output
+          )
+        if (nDeleted > 0) {
+          incremental.log.debug(s"$nDeleted orphan classfiles deleted")
+        }
         if (hasModified)
           incremental.cycle(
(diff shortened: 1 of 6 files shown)


Friendseeker:
Thanks for your contribution to Zinc! 

This PR was inactive for a while. Do you plan to revisit this PR, or would it be okay for other people to take over this PR?

jackkoenig:
I'm happy for anyone who has the time to take over! They should just make sure to take a look at the feedback on the related issue, see https://github.com/sbt/zinc/issues/1234#issuecomment-1675678662.

PR #1240, opened 2023-08-23
Provide better guidance about Scala 3 bug reporting?

SethTisue:


.github/ISSUE_TEMPLATE/---bug-report.md
```diff
@@ -7,6 +7,10 @@ assignees: ''
 
 ---
 
+## Scala 2 or 3?
+
+Zinc is for Scala 2 only. To report a bug in incremental compilation of Scala 3 code, visit https://github.com/lampepfl/dotty/issues
```
eed3si9n, line 12: Zinc is for both Scala 2 and 3. The tricky part is that the source of incremental compilation bug often comes from the compiler bridge imple. In the same vein, we might advise that Scala 2 incremental issue should also be reported to Scalac.

SethTisue: Hmm, well, apparently I still don't understand how this works, even after all this time.

Can you suggest a wording? I'm at a loss. What we can do to help guide users here? Can users reasonably be expected to distinguish issues that might stem from one repo or the other?

I'd also be interested in reading a document that explains it, if there is one. It seems I have not retained what @smarter and others have explained to me about this.

SethTisue: In short: in a Scala 3 context, what _does_ zinc provide...?

eed3si9n: Zinc implements an incremental compilation logic that is agnostic of the specific Scala version, which is how it can compile any 2.x or 3.x. The general logic is try the first compilation, then compare the API difference, and invalidate more sources as needed. Each individual step, like actually compiling, is where we invoke the compiler via the compiler bridge.

It may not be a good idea to push triaging to the end users, esp those who aren't in core Scala ecosystem, so I'm kind of ok with the current situation of ppl reporting bugs first to Zinc and redirecting them to Scalac or Dotty, when needed. A good rule of thumb is Scala version specific incremental bug. If it happens only on 2.13 or 3.x, there's a good chance that the fix will be in the API extraction part etc of the bridge.

SethTisue: let me check my understanding: in Scala 2, zinc consists of some injected compiler phases that gather information, and then some logic to decide what needs to be recompiled. (and then actually does it via the bridge)

in Scala 3, the equivalent of those injected compiler phases is built in to the compiler? but it remains up to Zinc to decide what to do with the information the compiler emits (in other word, what recompilation to request through the bridge?)

lrytz: IIUC, Scala 2 and 3 are the same here. Compiler phases in the "bridge" that collect data and expose it through the xsbti interface. Zinc implements this interface and uses that dependency / API data to drive incremental compilation.

What's different until 2.13.11 is that the bridge implementation for Scala 2 lives in the zinc repo, while the bridge implementation for Scala 3 lives in the dotty repo. But this is now aligned with 2.13.12.

An incremental compilation bug could be in zinc or in the bridge, as Eugene says that's probably very hard to tell for users.

SethTisue: > I'd also be interested in reading a document that explains it, if there is one

to answer this myself, there's this talk by @bishabosha, which I missed in Seattle: "How does Incremental Compilation Work with Scala 3" https://www.youtube.com/watch?v=1uuFxEAiPuQ

I'll watch it on the plane to Madrid ✈️

Friendseeker:
Should we close this PR? This PR was stale for a year and I don't see how we can potentially provide better guidance.

I definitely think improvements can be made on `sbt` side as I see some incremental compilation tickets buried in `sbt` backlog and not tagging with incremental compilation (e.g. https://github.com/sbt/sbt/issues/7589). But not much on `zinc` side.

PR #1247, opened 2023-09-11
Also enable inliner for Scala 2.13

mdedetrich:
Likely an oversight since the Scala 2 inliner also works for Scala 2.13


eed3si9n (approved):
Should be ok

PR #1248, opened 2023-09-11
Update Scala 2.13

mdedetrich:
Self explanatory


eed3si9n:
Scala 2.13.12 is the first Scala 2.x compiler to implement its own compiler bridge (https://github.com/scala/scala/pull/10472), so it might need some tweaking to make it work.

mdedetrich:
I see with the failing tests

mdedetrich:
Ill close the PR since someone with more knowledge about the compiler bridge should implement it.

Issue #1229, opened 2023-07-28
IncOptions.useOptimizedSealed is broken on 2.13


smarter:
build.sbt:
```scala
lazy val root = (project in file(".")).
  settings(
    name := "sbt-hello",
    scalaVersion := "2.13.11"
  )
ThisBuild / incOptions ~= { _.withUseOptimizedSealed(true) }
```

Sealed.scala:
```scala
sealed trait Sealed
class Child1 extends Sealed
```

PatMat.scala:
```scala
object PatMat {
  def test(x: Sealed): Int = x match {
    case _: Child1 => 1
  }
}
```

1. `compile`
2. Add an extra child in Sealed.scala: `class Child2 extends Sealed`
3. `compile` again, only Sealed.scala is recompiled.

If I change `scalaVersion` to `2.12.18`, then `PatMat.scala` is recompiled as expected.
I assume the zinc scripted tests such as https://github.com/sbt/zinc/tree/develop/zinc/src/sbt-test/source-dependencies/patMat-scope only run with 2.12, it'd be great to run them with 2.13 too to catch this sort of issues.

/cc @retronym who might be aware of projects that use `useOptimizedSealed`.

lrytz:
Fixed in https://github.com/sbt/zinc/pull/1278

PR #1278, opened 2023-11-05
Fix IncOptions.useOptimizedSealed not working for Scala 2.13

### LINKED ISSUE #753: optimized sealed option may not be working for 2.13
Ref https://github.com/sbt/zinc/pull/267

## steps

zincJVM2_13/testOnly sbt.inc.NameHashingCompilerSpec

```scala
    testIncrementalCompilation(
       changes = Seq(Other -> addNewSealedChildren),
       transitiveChanges = Set(Other3, Other),
       optimizedSealed = true
    )
```

## problem

On Scala 2.13 the test fails.

## expectation

Used names may not be picking this up correctly.

Friendseeker:
In `ExtractUsedName.scala`, whenever a pattern matching involving a sealed class is encountered, the dependency is registered, but only with `Default` scope instead of `PatMat` scope.

To fix it, an extra match case is added in `handleClassicTreeNode` to register the dependency in `PatMat` scope.

Closes #753 #1229

### DIFF
--- a/internal/compiler-bridge-test/src/test/scala/sbt/internal/inc/ExtractUsedNamesSpecification.scala
+++ b/internal/compiler-bridge-test/src/test/scala/sbt/internal/inc/ExtractUsedNamesSpecification.scala
@@ -3,6 +3,7 @@ package internal
 package inc
 
 import org.scalatest.diagrams.Diagrams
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #1282, opened 2023-11-15
Invalidate macro source when its dependency changes

dwijnand:
Invalidate macro classes that transitively depend on any of the recompiled classes

The macro expansion tree can depend on the behavioural change of any upstream code change,
not just API changes, so correctness requires aggressive recompilation of downstream classes.

Technically the macro doesn't need to be recompiled - it's the classes downstream of the macro
that need to be recompiled, so that the macros can be re-expanded.  But recompiling is the most
straightforward way to signal any classes downstream of the _macro_ that they need to recompile.

Also, note, that this solution only works for behavioural changes in sources within the same
subproject as the macro.  Changes in behaviour in upstream subprojects don't cause downstream
macro classes to recompile - because downstream projects only have visibility of the upstream
API, and if it changed, which is insufficient, and upstream projects have no other way than
their API to signal to downstream.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
@@ -492,9 +492,10 @@ private[inc] abstract class IncrementalCommon(
 
   /** Invalidates classes and sources based on initially detected 'changes' to the sources, products, and dependencies.*/
   def invalidateInitial(
-      previous: Relations,
+      previousAnalysis: Analysis,
       changes: InitialChanges
   ): (Set[String], Set[VirtualFileRef]) = {
+    val previous = previousAnalysis.relations
     def classNames(srcs: Set[VirtualFileRef]): Set[String] = srcs.flatMap(previous.classNames)
     def toImmutableSet(srcs: java.util.Set[VirtualFileRef]): Set[VirtualFileRef] = {
       import scala.collection.JavaConverters.asScalaIteratorConverter
@@ -521,8 +522,15 @@ private[inc] abstract class IncrementalCommon(
         invalidateClassesExternally(previous, externalAPIChange, isScalaSource)
       }.toSet
     }
+    val byMacroUsage = {
+      for {
+        class1 <- invalidatedClasses
+        class1Client <- previous.memberRef.internal.reverse(class1)
+        if previousAnalysis.apis.internalAPI(class1Client).hasMacro
+      } yield class1Client
+    }
 
-    val allInvalidatedClasses = invalidatedClasses ++ byExtSrcDep
+    val allInvalidatedClasses = invalidatedClasses ++ byExtSrcDep ++ byMacroUsage
     val allInvalidatedSourcefiles = addedSrcs ++ modifiedSrcs ++ byProduct ++ byLibraryDep
 
     if (previous.allSources.isEmpty)
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
@@ -176,7 +176,7 @@ private[inc] abstract class IncrementalCommon(
           if (isFullCompilation) Set.empty[String]
           else
             invalidateAfterInternalCompilation(
-              analysis.relations,
+              analysis,
               newApiChanges,
               recompiledClasses,
               cycleNum >= options.transitiveStep,
@@ -451,12 +451,13 @@ private[inc] abstract class IncrementalCommon(
    * @return A list of invalidated class names for the next incremental compiler run.
    */
   def invalidateAfterInternalCompilation(
-      relations: Relations,
+      analysis: Analysis,
       changes: APIChanges,
       recompiledClasses: Set[String],
       invalidateTransitively: Boolean,
       isScalaClass: String => Boolean
   ): Set[String] = {
+    val relations = analysis.relations
     val initial = changes.allModified.toSet
     val dependsOnClass = findClassDependencies(_, relations)
     val firstClassInvalidation: Set[String] = {
@@ -477,7 +478,14 @@ private[inc] abstract class IncrementalCommon(
     if (secondClassInvalidation.nonEmpty)
       log.debug(s"Invalidated due to generated class file collision: ${secondClassInvalidation}")
 
-    val newInvalidations = (firstClassInvalidation -- recompiledClasses) ++ secondClassInvalidation
+    // Invalidate macros that transitively depend on any of the recompiled classes
+    val thirdClassInvalidation = {
+      val transitive = IncrementalCommon.transitiveDeps(recompiledClasses, log)(dependsOnClass)
+      (transitive -- recompiledClasses).filter(analysis.apis.internalAPI(_).hasMacro)
+    }
+
+    val newInvalidations =
+      (firstClassInvalidation -- recompiledClasses) ++ secondClassInvalidation ++ thirdClassInvalidation
     if (newInvalidations.isEmpty) {
       log.debug("No classes were invalidated.")
       Set.empty
@@ -485,17 +493,16 @@ private[inc] abstract class IncrementalCommon(
       if (invalidateTransitively) {
         newInvalidations ++ recompiledClasses
       } else {
-        firstClassInvalidation ++ secondClassInvalidation
+        firstClassInvalidation ++ secondClassInvalidation ++ thirdClassInvalidation
       }
     }
   }
 
   /** Invalidates classes and sources based on initially detected 'changes' to the sources, products, and dependencies.*/
   def invalidateInitial(
-      previousAnalysis: Analysis,
+      previous: Relations,
       changes: InitialChanges
   ): (Set[String], Set[VirtualFileRef]) = {
-    val previous = previousAnalysis.relations
     def classNames(srcs: Set[VirtualFileRef]): Set[String] = srcs.flatMap(previous.classNames)
     def toImmutableSet(srcs: java.util.Set[VirtualFileRef]): Set[VirtualFileRef] = {
       import scala.collection.JavaConverters.asScalaIteratorConverter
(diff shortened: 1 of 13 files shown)


SethTisue:
@Friendseeker you've been doing invalidation stuff recently (which is noticed and much appreciated!) — perhaps you'd like to do a little reviewing on this one?

Friendseeker:
> @Friendseeker you've been doing invalidation stuff recently (which is noticed and much appreciated!) — perhaps you'd like to do a little reviewing on this one?

Thank you for considering me for this review; I really appreciate it! I must admit, my experience with Scala macros is quite limited. However, I'm willing to take a closer look at the PR and provide my comments where I can.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -521,8 +522,15 @@ private[inc] abstract class IncrementalCommon(
         invalidateClassesExternally(previous, externalAPIChange, isScalaSource)
       }.toSet
     }
+    val byMacroUsage = {
+      for {
+        class1 <- invalidatedClasses
+        class1Client <- previous.memberRef.internal.reverse(class1)
+        if previousAnalysis.apis.internalAPI(class1Client).hasMacro
```
Friendseeker: What if the dependency is transitive?

For example, in `macro-use` test case, if we create a new file `Data2.scala`:

```scala
class Data2 extends Data {}
```

And in `Macro.scala`, replace

```scala
val suffix = new Data().suffix
```

To 
```scala
val suffix = new Data2().suffix
```

Then `zincScripted/Test/run macros/macro-use` fails

Friendseeker:
@dwijnand I've reviewed the changes in your PR and understand the modifications you've made. However, to get a complete picture, could you please provide some insight into the reasoning behind this approach? It appears to be a quick fix, and I'm curious if there were specific constraints or considerations that led to this solution.

dwijnand:
I went with the first fix that worked, not being able to tell where I should look to slot this fix in.  In particular I ended up moving my change to "initial changes" from `detectAPIChanges` and `invalidateAfterInternalCompilation` where I was originally looking.  I'm open to suggestions and ideas.  Meanwhile I'll study your transitive counter-example.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
dwijnand: I couldn't tell why I wasn't reproducing your finding here, so I pushed the change as a commit to see whom CI corroborates with.

Friendseeker: What...

My sincere apology, now I also cannot reproduce the finding.

Maybe I ran the "counterexample" on wrong branch or something...

Friendseeker: My sincere apology again. My first comment was incorrect. Another level of indirection is needed.

Here's the test case that causes a fail.

[macro-use.zip](https://github.com/sbt/zinc/files/13417129/macro-use.zip)

Here's the test log on my machine

[scripted-macros-macro-use.log-1.log](https://github.com/sbt/zinc/files/13417133/scripted-macros-macro-use.log-1.log)

Friendseeker: That was my fault. Try

```scala
// Data2.scala
package lib

class Data2 {
  def suffix = new Data().suffix
}
```

instead. Then the dependency recorded in `previous.memberRef.internal` becomes

`Macro -> Data2, Data2 -> Data`

And therefore `class1Client <- previous.memberRef.internal.reverse(class1)` fails to iterate through `Macro`, resulting in missing invalidation.

Here's a zip of the test case and test log on my local machine.
[macro-use.zip](https://github.com/sbt/zinc/files/13418729/macro-use.zip)

[scripted-macros-macro-use.log-1.log](https://github.com/sbt/zinc/files/13418736/scripted-macros-macro-use.log-1.log)

Friendseeker: Also, while I cannot immediately construct a counter-example, would it be safer to place the check at end of every cycle instead of just the start of the first cycle?

While doing the check for every cycle might be a massive performance hit, I have a concern that new Macro dependencies might be introduced in later cycles, such that the initial check won't catch them.

dwijnand: Done

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -477,15 +478,22 @@ private[inc] abstract class IncrementalCommon(
     if (secondClassInvalidation.nonEmpty)
       log.debug(s"Invalidated due to generated class file collision: ${secondClassInvalidation}")
 
-    val newInvalidations = (firstClassInvalidation -- recompiledClasses) ++ secondClassInvalidation
+    // Invalidate macros that transitively depend on any of the recompiled classes
+    val thirdClassInvalidation = {
```
Friendseeker, line 495: LGTM in terms of transitive dependency correctness. There's probably still some edge cases left but this did cover the edge cases I can think of.

Preferably someone can do a performance review. If the performance is acceptable then I personally think this is good to go. If not, we might need to think of a way to optimize computation of `thirdClassInvalidation` (e.g. 2-way BFS, some tweaked version of union find...), or fall back to initial approach of checking at `invalidateInitial` only.

dwijnand: I don't think computing `thirdClassInvalidation` would impact performance, because `invalidateClassesInternally` already calls `transitiveDeps` - although there it operates on the classes in `changes`, which is a subset of `recompiledClasses` we're operating on here (because we care about classes that changed in behaviour, not just those that changed in API).  Let me know if I'm missing something.

Friendseeker: Thanks for the explanation! Overall LGTM.

eed3si9n (approved):
Thanks for the contribution (and thanks to reviewers as well)! I'm happy to merge this as-is, but my only ask would be to elaborate the motivation for this PR in the PR description as "problem such and such doesn't work. \n\n solution here's the approach to fix it." section.

Friendseeker:
I think I found an edge case.

If transitive dependencies of Macro lives in different projects, then `transitiveDeps` won't find cross-project transitive dependencies, as each project has their own Analysis (while the `transitiveDeps` call only uses information from the current project's analysis).

To address this issue, when performing dependency traversal, we need to traverse through both internal dependency (dependency within a project) and external dependency (project to project dependency), fetching new analyses during traversal when necessary. We can probably implement this logic via creating a new version of `dependsOnClasses`, or do it some other way if there's a more reasonable way to do it.

---

We can either address the edge case in this PR, or merge this PR first and create a new PR to address this edge case. I think either is fine.

dwijnand:
> We can either address the edge case in this PR, or merge this PR first and create a new PR to address this edge case. I think either is fine.

I think we should defer this part, perhaps to never.

The first reason is because I think it's common practice in projects to have a subproject dedicated for a macro, and that subproject doesn't have any dependencies.  And the reason for that (IIRC and if things haven't changed) because declaring a macro dependency only puts the macro jar on the compilation classpath so it's present during macro expansion, without any of it (potential) dependencies - thus only scala-library, scala-reflect and the jdk of course.  Now technically a project could develop that jar as separate subprojects and then assemble it into one jar, and then could suffer this problem, but... perhaps that reality doesn't exist and even if so, it's "only" incremental builds.

The second reason is I can't find a good way to implement that.  You mentioned that to address this you need to traverse external dependencies.  That won't work.  The intra-subproject solution we have here takes the "recompiledClasses" (which might or might not have API changes), e.g. `lib.InternalApi`, and deduces if there are any macro classes, e.g. `lib.Macro`, that (transitively) depend on any of those recompiled classes.  And if there are, the source files associated for those macro classes are recompiled.  If the macro is in a downstream project, we don't have the information locally.  If, instead, the current project is the one that contains the macro and the code that changes are upstream, e.g. `util.InternalApi`, then we don't even have visibility on `util.InternalApi` because `lib.Macro` only has visibility of `util.Data` the public API, and that API didn't change.

I'll detail that in the PR description and in the code.

Friendseeker:
> > We can either address the edge case in this PR, or merge this PR first and create a new PR to address this edge case. I think either is fine.
> 
> 
> 
> I think we should defer this part, perhaps to never.
> 
> 
> 
> The first reason is because I think it's common practice in projects to have a subproject dedicated for a macro, and that subproject doesn't have any dependencies.  And the reason for that (IIRC and if things haven't changed) because declaring a macro dependency only puts the macro jar on the compilation classpath so it's present during macro expansion, without any of it (potential) dependencies - thus only scala-library, scala-reflect and the jdk of course.  Now technically a project could develop that jar as separate subprojects and then assemble it into one jar, and then could suffer this problem, but... perhaps that reality doesn't exist and even if so, it's "only" incremental builds.
> 
> 
> 
> The second reason is I can't find a good way to implement that.  You mentioned that to address this you need to traverse external dependencies.  That won't work.  The intra-subproject solution we have here takes the "recompiledClasses" (which might or might not have API changes), e.g. `lib.InternalApi`, and deduces if there are any macro classes, e.g. `lib.Macro`, that (transitively) depend on any of those recompiled classes.  And if there are, the source files associated for those macro classes are recompiled.  If the macro is in a downstream project, we don't have the information locally.  If, instead, the current project is the one that contains the macro and the code that changes are upstream, e.g. `util.InternalApi`, then we don't even have visibility on `util.InternalApi` because `lib.Macro` only has visibility of `util.Data` the public API, and that API didn't change.
> 
> 
> 
> I'll detail that in the PR description and in the code.

Thanks for the explanation. Let's defer it then!

PR #1287, opened 2023-11-26
Use the most up-to-date analysis for binary to source class name lookup

### LINKED ISSUE #1268: Missing invalidation when deleting source file
Zinc misses an invalidation when deleting a source file.

The issue shows up with an existing test case in the zinc repo: https://github.com/sbt/zinc/tree/develop/zinc/src/sbt-test/source-dependencies/empty-package.

For some reason, the test passes when it runs with zinc's `scripted` implementation. It fails when reproducing it in a small sbt project (make sure to `set incOptions := incOptions.value.withRecompileAllFraction(1.0)`), or when running it with sbt's `scripted` implementation (used in scala/scala).

To reproduce it using scala/scala:

  - Check out this branch https://github.com/lrytz/scala/tree/missing-invalidation
  - Run `set sbtTest/scriptedBufferLog := false` and `set ThisBuild / Compile / packageDoc / publishArtifact := false` in sbt
  - Run `sbtTest/scripted source-dependencies/empty-package`

Output from the scala/scala test: https://gist.github.com/lrytz/61930adc6ed4f7d0648bbc1abf3b0bef

For reference, output when running the test in the zinc repo: https://gist.github.com/lrytz/e42076ecadc49c770465f97c12a0602b

The issue reproduces with 2.13.12, older 2.13 releases, and also with 2.12.18 and 3.3.1. I tested 2.13.12 with sbt 1.9.6 (where it uses the binary bridge) and 1.9.4 (where it uses the bridge compiled from zinc sources), it happens in both cases.

Transfered from https://github.com/scala/bug/issues/12887 - I was first thinking this is a new issue in scala/scala, but it is in fact an old issue in zinc (possibly the compiler bridge).

Friendseeker:
This PR fixes #1268

### Cause

During Dependency Extraction, `internalBinaryToSourceClassName` is used to map binary class name to source class name. It is constructed at the first cycle. Hence in subsequent cycles, it is outdated. This results in missing internal member reference dependency between `a.Use` `pkgName.Test`.

To fix it, we rebuild lookup using the most up-to-date analysis.

### Other changes

This PR also removed `internalSourceToClassNamesMap`. It is currently not used, but it has the same staleness problem as `internalBinaryToSourceClassName`. So might as well remove it to prevent someone from accidentally using it in the future.

### Why `empty-package` erroneously passed on zinc repo

The complier bridge did recognize a dependency between `a.Use` and `pkgName.Test` exists. However, since `internalBinaryToSourceClassName` was stale, zinc failed to map `pkgName.Test` back to its `.scala` source file, and instead misclassified the dependency as a external library dependency. This can be observed in the [zinc test log](https://gist.github.com/lrytz/e42076ecadc49c770465f97c12a0602b):

```
[debug] previous = Stamps for: 4 products, 2 sources, 1 libraries
[debug] current source = Set(${BASE}/pro/Use.scala)
[debug] Invalidating '${BASE}/pro/target/classes/pkgName/Test.class' because pkgName.Test$ came from analysis Analysis: 2 Scala sources, 4 classes
```

Then in `invalidateInitial`, all sources that depends on `pkgName.Test` library is invalidated via `val byLibraryDep = changes.libraryDeps.flatMap(previous.usesLibrary)`

### Validating the fix

`empty-package` test is modified to assert dependency between `a.Use` `pkgName.Test` must exist after the 2nd compilation run. 

Running the modified `empty-package` on current latest `develop` branch commit now results in [test failure](https://gist.github.com/Friendseeker/7ea5ed3ef7d5f1a54f5abbfe5fbfb15a).

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -152,10 +152,6 @@ object Incremental {
     log.debug(s"[zinc] IncrementalCompile -----------")
     val previous = previous0 match { case a: Analysis => a }
     val currentStamper = Stamps.initial(stamper)
-    val internalBinaryToSourceClassName = (binaryClassName: String) =>
-      previous.relations.productClassName.reverse(binaryClassName).headOption
-    val internalSourceToClassNamesMap: VirtualFile => Set[String] = (f: VirtualFile) =>
-      previous.relations.classNames(f)
 
     val earlyJar = extractEarlyJar(earlyOutput)
     val pickleJarPair = earlyJar.map { p =>
@@ -183,8 +179,6 @@ object Incremental {
           runProfiler.timeCompilation(startTime, System.nanoTime() - startTime)
         },
         new AnalysisCallback.Builder(
-          internalBinaryToSourceClassName,
-          internalSourceToClassNamesMap,
           lookup.lookupAnalyzedClass(_, _),
           currentStamper,
           options,
@@ -250,13 +244,7 @@ object Incremental {
     log.debug("[zinc] compileAllJava")
     val previous = previous0 match { case a: Analysis => a }
     val currentStamper = Stamps.initial(stamper)
-    val internalBinaryToSourceClassName = (binaryClassName: String) =>
-      previous.relations.productClassName.reverse(binaryClassName).headOption
-    val internalSourceToClassNamesMap: VirtualFile => Set[String] =
-      (f: VirtualFile) => previous.relations.classNames(f)
     val builder = new AnalysisCallback.Builder(
-      internalBinaryToSourceClassName,
-      internalSourceToClassNamesMap,
       lookup.lookupAnalyzedClass(_, _),
       currentStamper,
       options,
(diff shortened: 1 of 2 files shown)


Friendseeker:
TODO:

- [x] The edited `empty-package` test passes in current latest 'develop' commit.
  - Expected: edited `empty-package` test should fail.

eed3si9n (approved):
I feel slightly nervous about deleting things, but hopefully this is ok.

Friendseeker:
> I feel slightly nervous about deleting things, but hopefully this is ok.

Here's my chain of thought:

The only case for problem to emerge is for  `incHandlerOpt` to be `None` while `internalBinaryToSourceClassName` lookup is still needed. For Scala compilation,  `incHandlerOpt` is never  `None`. For Java compilation,  `incHandlerOpt` is `None`, but `internalBinaryToSourceClassName` is called in one place only:

```scala
// JavaAnalyze.scala
(getMappedSource(fromBinaryName), getMappedSource(onBinaryName)) match {
  case (Some(fromClassName), Some(onClassName)) =>
    trapAndLog(log) {
      analysis.classDependency(onClassName, fromClassName, context)
    }
  case (Some(fromClassName), None) =>
    trapAndLog(log) {
      val cachedOrigin = classfilesCache.get(onBinaryName)
      for (file <- cachedOrigin.orElse(loadFromClassloader())) {
        val binaryFile: Path = remapClassFile(file)
        analysis.binaryDependency(
          binaryFile,
          onBinaryName,
          fromClassName,
          source,
          context
        )
      }
    }
  case (None, _) => // It could be a stale class file, ignore
```

In the above code, `analysis.binaryDependency` calls `internalBinaryToSourceClassName`, however it only happens when `getMappedSource(onBinaryName)` is `None`, which means the source file does not exist anyways.

---

Therefore, if the above chain of thought is sound, this change should be okay.

PR #1288, opened 2023-11-27
Register inheritance relationship for SAM-type lambda

### LINKED ISSUE #830: SAM type change causes undercompilation
This is yet another example of Zinc lagging Scala language spec change. 
https://www.scala-lang.org/news/2.12.0/#lambda-syntax-for-sam-types

## steps

Haven't minimized it but I noticed while coding. First you have a Java interface in one subproject.

```java
import java.nio.file.Path;
import java.util.Optional;

public interface SingleOutput {
    public Path getOutputDirectory();

    public default Optional<Path> getSingleOutput() {
        return Optional.of(getOutputDirectory());
    }
}
```

In another subproject, use it as  SAM-type lambda:

```scala
val output: SingleOutput = () => RootFilePath.resolve("out")
```

Then change the Java interface..

```java
import java.io.File;
import java.util.Optional;

public interface SingleOutput {
    public File getOutputDirectory();

    public default Optional< File > getSingleOutput() {
        return Optional.of(getOutputDirectory());
    }
}
```

## problem

The Scala code is not invalidated. So locally `compile` keeps succeeding. On CI the build fails if it's compiled from scratch.

## expectation

Invalidate correctly.

Friendseeker:
During dependency extraction, register inheritance dependency for each SAM-type lambda, by fetching SAM type from tree attachment.

Fixes #830

### DIFF
--- a/internal/compiler-bridge/src/main/scala_2.13/xsbt/Compat.scala
+++ b/internal/compiler-bridge/src/main/scala_2.13/xsbt/Compat.scala
@@ -28,6 +28,12 @@ abstract class Compat {
       func(a.original)
     }
   }
+
+  protected def processSAMAttachment(f: Function)(addDependency: Symbol => Unit): Unit = {
+    f.attachments.get[SAMFunction].foreach(sam => {
+      addDependency(sam.samTp.typeSymbol)
+    })
+  }
 }
 object Compat {
   // IR is renamed to Results
--- a/internal/compiler-bridge/src/main/scala-2.11/xsbt/Compat.scala
+++ b/internal/compiler-bridge/src/main/scala-2.11/xsbt/Compat.scala
@@ -21,6 +21,9 @@ abstract class Compat {
   import global._
 
   protected def processOriginalTreeAttachment(in: Tree)(func: Tree => Unit): Unit = ()
+
+  protected def processSAMAttachment(f: Function)(addDependency: Symbol => Unit): Unit =
+    ()
 }
 object Compat {
   // IR is renamed to Results
--- a/internal/compiler-bridge/src/main/scala-2.12/xsbt/Compat.scala
+++ b/internal/compiler-bridge/src/main/scala-2.12/xsbt/Compat.scala
@@ -24,6 +24,12 @@ abstract class Compat {
   protected def processOriginalTreeAttachment(in: Tree)(func: Tree => Unit): Unit = {
     Compat.OriginalTreeTraverser.Instance.traverseOriginal(in)(func)
   }
+
+  protected def processSAMAttachment(f: Function)(addDependency: Symbol => Unit): Unit = {
+    f.attachments.get[SAMFunction].foreach(sam => {
+      addDependency(sam.samTp.typeSymbol)
+    })
+  }
 }
 object Compat {
   // IR is renamed to Results
--- a/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
@@ -475,6 +475,16 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with 
         val sym = if (tree.symbol.isModule) tree.symbol.moduleClass else tree.symbol
         localToNonLocalClass.resolveNonLocal(sym)
         super.traverse(tree)
+
+      case f: Function =>
+        processSAMAttachment(f)(symbol => {
+          addDependency(symbol)
+          // Not using addInheritanceDependency as it would incorrectly classify dependency as non-local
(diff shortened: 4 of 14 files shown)


internal/compiler-bridge/src/main/scala_2.13/xsbt/Compat.scala
```diff
@@ -28,6 +28,15 @@ abstract class Compat {
       func(a.original)
     }
   }
+
+  protected def processSAMAttachment(f: Function)(addInheritanceDependency: Symbol => Unit)
+      : Unit = {
+    f.attachments.get[SAMFunction].foreach { sam =>
+      {
+        addInheritanceDependency(sam.samTp.typeSymbol)
```
Friendseeker: I mainly want to request for feedback for whether the inheritance relationship is registered properly in `processSAMAttachment`, As, the existing logic to register inheritance relationship has a rich history behind for which I don't fully understand. For example, I don't quite know why it is necessary to call both `addDependency` and `addInheritanceDependency` at the same time. I also don't know what is the exact purpose of the `addTypeDependencies` calls.

Existing Logic:
```scala
case Template(parents, self, body) =>
  // use typeSymbol to dealias type aliases -- we want to track the dependency on the real class in the alias's RHS
  def flattenTypeToSymbols(tp: Type): List[Symbol] =
    if (tp eq null) Nil
    else
      tp match {
        // rt.typeSymbol is redundant if we list out all parents, TODO: what about rt.decls?
        case rt: RefinedType => rt.parents.flatMap(flattenTypeToSymbols)
        case _               => List(tp.typeSymbol)
      }

  val inheritanceTypes = parents.map(_.tpe).toSet
  val inheritanceSymbols = inheritanceTypes.flatMap(flattenTypeToSymbols)

  debuglog(
    "Parent types for " + tree.symbol + " (self: " + self.tpt.tpe + "): " + inheritanceTypes + " with symbols " + inheritanceSymbols
      .map(_.fullName)
  )

  inheritanceSymbols.foreach { symbol =>
    addInheritanceDependency(symbol)
    addDependency(symbol)
  }

  inheritanceTypes.foreach(addTypeDependencies)
  addTypeDependencies(self.tpt.tpe)

  traverseTrees(body)
```

Friendseeker:
@SethTisue Since this PR includes compiler bridge changes that will be ported to [scala/scala](https://github.com/scala/scala), may I also get a review from you or other maintainers of [scala/scala](https://github.com/scala/scala)?

Friendseeker:
- [x] investigate why the fix did not work on scala/scala repo: https://github.com/scala/scala/pull/10617

internal/compiler-bridge/src/main/scala_2.13/xsbt/Compat.scala
lrytz: @Friendseeker I have never looked into zinc's dependency tracking, so not speaking out of experience.

I found some documentation in [Relations.scala](https://github.com/sbt/zinc/blob/develop/internal/zinc-core/src/main/scala/sbt/internal/inc/Relations.scala). Incremental compilation returns an `Analysis` containing the dependency analysis, the `relations: Relations` field is where the depdencies are represented.

It seems worth reading through that file. The top-level comment explains "internal" vs "external" vs "library" deps. The class has fields
  - `memberRef`: "dependency relation between classes introduced by member reference"
  - `inheritance`: "class dependency relation between classes introduced by inheritance"
  - `localInheritance`: see example in comment

See also [DependencyContext.java](https://github.com/sbt/zinc/blob/develop/internal/compiler-interface/src/main/contraband-java/xsbti/api/DependencyContext.java).

So it sounds like "inheritance dependencies" is a subset of "dependencies".

`addTypeDependencies` despite its similar name to `addDependency`/ `addInheritanceDependency`  is not a different kind of dependency registered in the analysis. It's a method that traverses all components of a `Type` and ultimately invokes `addDependency` for symols found within that type.

I guess this all means that we should also call `addDependency(sam.samTp.typeSymbol)` here. Again I really don't have experience, so I'm not sure. Maybe you can compare dependencies registered for `val f: () => Int = () => 1` vs `val f: A = () => 1`? [DependencySpecification.scala](https://github.com/sbt/zinc/blob/develop/internal/compiler-bridge-test/src/test/scala/sbt/internal/inc/DependencySpecification.scala) might be a good place to implement that.

Friendseeker: Thanks for the detailed explanation! I will look into comparing dependencies registered in [DependencySpecification.scala](https://github.com/sbt/zinc/blob/develop/internal/compiler-bridge-test/src/test/scala/sbt/internal/inc/DependencySpecification.scala) once I have some chunks of free time.

lrytz (approved):

PR #1289, opened 2023-11-28
Encode parent trait private members in `extraHash` to propagate `TraitPrivateMembersModified` across external dependency

Friendseeker:
This PR fixes #662

### Cause of the issue

To determine if a class has API change `TraitPrivateMembersModified`, `extraHash`, which is a hash of private API of a trait is computed. Direct comparison on `extraHash` is used to determine if a trait has private member modified.

When we have the following dependency `c.C -> b.B -> a.A`, (`c.C` extends `b.B`, `b.B` extends `a.A`), for which `c.C, b.B, a.A` belongs to different projects, the dependencies are external, hence when `A` has private member modified, the following happens

- `A` is recompiled
- `B` is recompiled since API of `A` has change `TraitPrivateMembersModified` and `B` extends `A`.
- API of `B` is not changed, hence `C` is not recompiled even though `C` extends `B`.

### Fix
When we compute `extraHash` for a child trait, use the parent traits' `extraHash`. Therefore, when parent has private members modified, child `extraHash` would change, hence child would have `TraitPrivateMembersModified`, propagating it.

PR #1290, opened 2023-11-28
Include internal dependency in `extraHash` computation

Friendseeker:
### Issue

PR #1289 addressed under-compilation for traits with transitive external inheritance dependencies by propagating `TraitPrivateMembersModified` from parent trait to child traits.

However, when the hierarchy consists a mix of internal & external dependencies, `TraitPrivateMembersModified` would fail to propagate, resulting in under compilation.

For example, say `c.C -> b.B -> a.A`, (`c.C` extends `c.B`, `c.B` extends `a.A`), for which `c.C` and `c.B, a.A` are on different projects, then when `a.A` 's private member is changed, the propagation stops at `c.B` since computation of `c.B`'s `extraHash` only takes account of `a.A`'s external dependency, while `a.A` is part of `c.B`'s internal dependency. Hence `c.C` never gets recompiled.

### Fix

Include internal dependency in `extraHash` computation.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -892,9 +892,11 @@ private final class AnalysisCallback(
             incHandlerOpt match {
               case Some(handler) =>
                 val analysis = handler.previousAnalysis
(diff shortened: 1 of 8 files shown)


eed3si9n (approved):
Thanks!

PR #1294, opened 2023-12-01
Increase timing resolution in CacheHashingSpec

szeiger:
I get times of 0µs vs ~30µs running this test on an M2 Max. The test fails at millisecond precision because usually both values are 0.


Friendseeker:
I don't quite understand why `/ 1000000` is not removed from `timeUs`, but otherwise LGTM.

```scala
 def timeUs[R](block: => R): Long = {
    val t0 = System.nanoTime()
    block // call-by-name
    val t1 = System.nanoTime()
    (t1 - t0) / 1000000
  }
```

Thanks for the PR!

szeiger:
Huh, you're right, it still measures milliseconds. So it must have been a different problem. I'm not sure why it failed originall, I can't reproduce it anymore.

Issue #1296, opened 2023-12-02
`compiler-bridge-test` & `zinc-scripted` do not show up in IntelliJ


Friendseeker:
## steps

Open zinc via IntelliJ

## problem

`compiler-bridge-test` & `zinc-scripted` are not properly recognized by IntelliJ.

Hence none of the smart IDE features work when editing files from `compiler-bridge-test` & `zinc-scripted`.

## expectation

The two projects should are recognized by IntelliJ.

## notes

This is certainly possible. In fact the following changes make IntelliJ recognize `compiler-bridge-test` & `zinc-scripted`.

```diff
Index: build.sbt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/build.sbt b/build.sbt
--- a/build.sbt	(revision acd9ad679eae93105b31daf3fe054c13b1e5218c)
+++ b/build.sbt	(date 1701506401387)
@@ -533,11 +533,11 @@
  * (Zinc API Info, which transitively depends on IO).
  */
 lazy val compilerBridgeTest = (project in internalPath / "compiler-bridge-test")
-  .dependsOn(zinc.jvm(scala213) % "compile->compile;test->test", compilerInterface.jvm(false))
+  .dependsOn(zinc.jvm(scala212) % "compile->compile;test->test", compilerInterface.jvm(false))
   .settings(
     name := "Compiler Bridge Test",
     baseSettings,
-    scalaVersion := scala213,
+    scalaVersion := scala212,
     compilerVersionDependentScalacOptions,
     // we need to fork because in unit tests we set usejavacp = true which means
     // we are expecting all of our dependencies to be on classpath so Scala compiler
@@ -548,6 +548,7 @@
     Test / javaOptions += "-Xmx1G",
     Test / javaOptions += s"-Dzinc.build.compilerbridge.scalaVersion=${scalaVersion.value}",
     publish / skip := true,
+    ideSkipProject := false,
   )
 
 val scalaPartialVersion = Def.setting(CrossVersion.partialVersion(scalaVersion.value))
@@ -656,7 +657,6 @@
   .enablePlugins(BuildInfoPlugin)
   .settings(
     baseSettings,
-    ideSkipProject := true, // otherwise IntelliJ complains
     publish / skip := true,
     name := "zinc Scripted",
     Compile / buildInfo := Nil, // Only generate build info for tests
@@ -668,7 +668,7 @@
[...]

PR #1297, opened 2023-12-02
Make `compiler-bridge-test` show up in IntelliJ

Friendseeker:
Partially fix https://github.com/sbt/zinc/issues/1296

Validating the fix

Screenshot 2023-12-02 at 10 19 16 AM

As shown in screenshot, `compiler-bridge-test` now shows up as a project in IDE, with smart IDE features (e.g. Type Inference) working.

PR #1298, opened 2023-12-02
Make `zinc-scripted` show up in IntelliJ

Friendseeker:
Fixes other half of https://github.com/sbt/zinc/issues/1296

Validate the fix:

Screenshot 2023-12-02 at 12 20 52 PM

(As shown in screenshot, smart IDE features are now working)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -652,11 +652,10 @@ lazy val zincClassfile = (projectMatrix in internalPath / "zinc-classfile")
 
 // re-implementation of scripted engine
 lazy val zincScripted = (projectMatrix in internalPath / "zinc-scripted")
-  .dependsOn(zinc % "compile;test->test")
+  .dependsOn(zinc % "compile;test->test", compilerBridge % "compile;test->test")
   .enablePlugins(BuildInfoPlugin)
   .settings(
     baseSettings,
-    ideSkipProject := true, // otherwise IntelliJ complains
     publish / skip := true,
     name := "zinc Scripted",
     Compile / buildInfo := Nil, // Only generate build info for tests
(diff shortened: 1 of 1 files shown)


Friendseeker:
@SethTisue 

Since this PR is changing some logics written by you, would you mind to take a look at this PR? 

(and preferably also take a look at https://github.com/sbt/zinc/pull/1297, since both PR are closely related).

---

I will be busy for the rest of December since I have a couple exams to deal with, so I might not be able to immediately respond to PR feedbacks, but I will still try my best to address feedbacks ASAP.

build.sbt
```diff
@@ -667,9 +666,6 @@ lazy val zincScripted = (projectMatrix in internalPath / "zinc-scripted")
   )
   .defaultAxes(VirtualAxis.jvm, VirtualAxis.scalaPartialVersion(scala212))
   .jvmPlatform(scalaVersions = List(scala212))
-  .configure(
-    _.dependsOn(compilerBridge210, compilerBridge211, compilerBridge212, compilerBridge213)
```
eed3si9n, line 671: This makes `zincScripted` depend on all compiler bridge JARs whereas the one above will only depend on the ones for Scala 2.12? I wonder if this was meant to create build order dependency so calling the scripted test would force rebuilding compiler bridges when we change the bridge imple. Could you check if that still works plz?

SethTisue:
I no longer remember what I might have been thinking. I'm not even confident what commits you might be referring to — sbt/zinc@4d394395c23d3d44007877b19c83994bd27361d0 ? regardless, probably I was just fumbling around

if you ask a more specific question it might ring a bell in my memory, but otherwise we should probably just rely on Eugene's say-so; I see he already merged the other PR

eed3si9n:
Closing this for now for inactivity. Feel free to reopen this if you want to pick this up.

PR #1259, opened 2023-09-18
Refix compact names w/o breaking local names

dwijnand:
The previous fix phase-traveled to after flatten, which is the threshold
for Symbol#name to start returning flattened names.  Unfortunately there
is a bug (one of a few in nsc) in that the owner of the local class
symbols isn't changed as a part of a lambda lift info transform, it's
changed as a side-effect of doing lambda lift tree transform.
Additionally the flattened name is cached on first use, which in this
case means using the wrong owner - that's why the compiler failed in
pekko, because there's a check in the backend that the same classfile
(by name) isn't emitted twice.

Turns out, we don't need the full names of method local classes, so we
can push that condition a bit higher in registerGeneratedClasses,
avoiding useless work, useless phase travel, and erroneous caching.
We also just straight up avoid phase traveling from xsbt-api (which runs
after typer) all the way to flatten (which is almost at the end) for the
flatten, compactified name, by reimplementing `flattenedName`, with the
correct, recursive, flattened owner and name.

By the by, the test case is pretty convoluted, despite trying to make it
simpler - how many times have you defined and instantiated a
method-local anonymous class, defining one of its methods using a case
object?

PR #1300, opened 2023-12-03
Make scripted test run with Scala 2.13 by default

Friendseeker:
Scala 2.13 is significantly more popular than Scala 2.12, hence it would cover more demographics for Zinc Scripted Test to run with 2.13 by default instead of 2.12.

[scala/scala](https://github.com/scala/scala) has scripted test running at 2.13, but ideally we catch 2.13 specific issue via Zinc CI directly, instead of relying on [scala/scala](https://github.com/scala/scala) CI.


Friendseeker:
Two test failures revealed by this PR

- `source-dependencies/compactify`
- `source-dependencies/compactify-nested`

@dwijnand Would you mind to give it a quick look? Just want to know if these are false positives or not.

dwijnand:
No, they're true positives.  But only because as of 2.13.12 we use in-sources compiler bridges (https://github.com/scala/scala/pull/10472) and the fix that is merged here (#1259) has been merged upstream (https://github.com/scala/scala/pull/10542) but hasn't been released yet.

lrytz:
@dwijnand are you sure we are using the binary bridge? Because IIUC that decision is not done by zinc, but by the build tool (https://github.com/sbt/sbt/blob/v1.9.7/zinc-lm-integration/src/main/scala/sbt/internal/inc/ZincLmUtil.scala#L93-L97).

In this case this would mean the scripted testing framework implementation within the zinc repo, no?

The two tests may be failing because `-Xmax-classfile-name` in 2.12 defaults to 255, but the setting was removed in 2.13 with a new default of 240 (https://github.com/scala/scala/pull/7497).

This might be the reason for `compactify-nested` failing, @Friendseeker see https://github.com/scala/scala/blob/2.13.x/test/sbt-test/source-dependencies/compactify-nested/test for the 2.13 expected class file names.

Not sure about the `compatcify` test.

internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
```diff
@@ -118,9 +118,12 @@ class IncHandler(directory: Path, cacheDir: Path, scriptedLog: ManagedLogger, co
 
   def initBuildStructure(): Unit = {
     val build = initBuild
+    val defaultScalaVersion = "2.13.12"
```
Friendseeker, line 121: Would it be better for `defaultScalaVersion` to be stored in `sys.props` instead?

e.g. https://github.com/sbt/zinc/blob/3e0ea937035c0319faad32899fe5ce842259be2c/internal/compiler-bridge-test/src/test/scala/sbt/internal/inc/CompilingSpecification.scala#L33-L36

Then we can conveniently setup CI to run scripted tests against both 2.12 & 2.13 in a clean manner. Despite 2.12 is phasing out, it might be still worthy it to keep 2.12 coverage for the slim chance that some issue only surface in 2.12.

Friendseeker: @xuwei-k Since you did lots of work on the CI & Testing side, would you mind to give a quick look on this PR? Your opinion would be valuable.

Friendseeker:
According to https://www.jetbrains.com/lp/devecosystem-2023/scala/ more than 1/3 of users use Scala 2.12. I guess the PR can wait

Issue #932, opened 2020-10-08
Feature request: Build and return `SourceInfos` when incremental compilation fails


adpi2:
When compilation fails, Zinc throws a `xsbti.CompileFailed` which contains:
- `arguments: Array[String]`
- `problems: Array[Problem]`

Based on the previous analysis and the `CompileFailed` instance I am trying to build an accurate description of the entire codebase but I am lacking information.

I don't know which of the old problems are still valid. Typically, a warning in a file is not reported again if the file is not recompiled, but I don't know if the file has been recompiled or not.

Ideally, the `CompileFailed` exception could contain an instance of `SourceInfos` which would give us the list of problems, the old ones that are still valid and the new ones.

At least, as an alternative, I would like to know which files have been recompiled.

adpi2:
@Friendseeker Since you closed #1306, should we assume this issue cannot be fixed?

Friendseeker:
@adpi2 Might want to double check with the Scala 2 compiler folks but I do suspect this problem cannot be fixed. The issue is that the compiler early stops and does not necessarily return a complete list of all problems in the recompiled files.

adpi2:
Re-reading [your comment in #1306](https://github.com/sbt/zinc/pull/1306#discussion_r1423608153) I think it's fine that some errors are not reported. What I need is a complete view of the reported errors so that I don't have to buffer then in the `BuildServerReporter` in sbt.

It's kind of the consensus already: there are some errors, we fix them, and then some more errors can appear. Everybody seems okay with that.

So I think we should merge #1306

Friendseeker:
@adpi2 Glad to hear your assessment. I shall work on getting #1306 merged. This is also a great opportunity to get #1309 merged as both require `AnalysisCallback3`. As long as we can get #1306 and #1309 merged in quick succession we can avoid the need to introduce `AnalysisCallback4`.

eed3si9n:
When we bump the AnalysisCallback, we probably want to send PRs to Scala 2.13 and Scala 3 so the Scala teams don't have to figure out what to change.

Friendseeker:
Just realized there's still work to do. Need to test the below cases. I don't remember my PR covering these cases so probably need to implement these in another PR.

- For mixed scala / java compilation or during `compileAllJava` workflow
  - Successful compilation should save java warnings in Analysis
  - Unsuccessful compilation (during java compile) should also throw an exception containing `SourceInfos`

Friendseeker:
@adpi2 

[...]

PR #1306, opened 2023-12-09
[1.x] Emit `SourceInfos` when incremental compilation fails

Friendseeker:
This PR implements https://github.com/sbt/zinc/issues/932 via building `SourceInfos` from analysis callback and pass them to `handleErrors` in `CompilerBridge.scala`.

c.c. @adpi2

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -590,7 +590,7 @@ private final class AnalysisCallback(
     progress: Option[CompileProgress],
     incHandlerOpt: Option[Incremental.IncrementalCallback],
     log: Logger
-) extends xsbti.AnalysisCallback2 {
+) extends xsbti.AnalysisCallback3 {
   import Incremental.CompileCycleResult
 
   // This must have a unique value per AnalysisCallback
@@ -1064,6 +1064,25 @@ private final class AnalysisCallback(
           externalDeps,
           libDeps
         )
+    }
+  }
+
+  def getSourceInfos: SourceInfos = {
+    // Collect Source Info from current run
+    val sources = reporteds.keySet ++ unreporteds.keySet ++ mainClasses.keySet
+    val sourceToInfo = sources.map { source =>
+      val info = SourceInfos.makeInfo(
+        getOrNil(reporteds.iterator.map { case (k, v) => k -> v.asScala.toSeq }.toMap, source),
+        getOrNil(unreporteds.iterator.map { case (k, v) => k -> v.asScala.toSeq }.toMap, source),
+        getOrNil(mainClasses.iterator.map { case (k, v) => k -> v.asScala.toSeq }.toMap, source)
+      )
+      (source, info)
+    }.toMap
+    val sourceInfoFromCurrentRun = SourceInfos.of(sourceToInfo)
+    // Collect reported problems from previous run
+    incHandlerOpt.map(_.previousAnalysisPruned) match {
+      case Some(prevAnalysis) => prevAnalysis.infos ++ sourceInfoFromCurrentRun
+      case None               => sourceInfoFromCurrentRun
     }
   }
 
--- a/internal/zinc-testing/src/main/scala/xsbti/TestCallback.scala
+++ b/internal/zinc-testing/src/main/scala/xsbti/TestCallback.scala
@@ -15,12 +15,12 @@ import java.io.File
 import java.nio.file.Path
 import java.{ util => ju }
 import ju.Optional
-
-import xsbti.api.{ DependencyContext, ClassLike }
+import xsbti.api.{ ClassLike, DependencyContext }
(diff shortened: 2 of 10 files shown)


internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
```diff
@@ -260,4 +261,6 @@ void problem(String what,
     boolean isPickleJava();
 
     Optional<T2<Path, Path>> getPickleJarPair();
+    
+    ReadSourceInfos getSourceInfos();
```
Friendseeker: I want to mainly request for feedback on whether the public API change is done correctly. (I am not sure about the detailed history, but API changes there seems to have caused trouble before since in-house compiler bridges depend on this interface.)

Friendseeker: Did some more research and ehh... I guess create a `AnalysisCallback3` is the way to go, like what is done in https://github.com/sbt/zinc/pull/1226

internal/zinc-testing/src/main/scala/xsbti/TestCallback.scala
```diff
@@ -153,6 +153,8 @@ class TestCallback extends AnalysisCallback2 {
   override def isPickleJava: Boolean = false
 
   override def getPickleJarPair = Optional.empty()
+
+  override def getSourceInfos: ReadSourceInfos = null
```
adpi2: Should it be `SourceInfos.empty` instead?

internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -1068,6 +1068,25 @@ private final class AnalysisCallback(
...
+    val sources = reporteds.keySet ++ unreporteds.keySet ++ mainClasses.keySet
+    val sourceToInfo = sources.map { source =>
+      val info = SourceInfos.makeInfo(
+        getOrNil(reporteds.iterator.map { case (k, v) => k -> v.asScala.toSeq }.toMap, source),
+        getOrNil(unreporteds.iterator.map { case (k, v) => k -> v.asScala.toSeq }.toMap, source),
+        getOrNil(mainClasses.iterator.map { case (k, v) => k -> v.asScala.toSeq }.toMap, source)
+      )
+      (source, info)
+    }.toMap
+    val sourceInfoFromCurrentRun = new MSourceInfos(sourceToInfo)
```
adpi2: What about that?

```suggestion
    val sourceInfoFromCurrentRun = SourceInfos.of(sourceToInfo)
```

adpi2 (commented):
Thanks a lot!

Here are some small suggestions but apart from that it looks good to me.

internal/zinc-testing/src/main/scala/xsbti/TestCallback.scala
Friendseeker: `TestCallback.scala` lives in `zinc-testing`, therefore it only has access to `compiler-interface` (I assume the reason behind it is that `zinc-testing` should not depend on any zinc internal.

```scala
lazy val zincTesting = (projectMatrix in internalPath / "zinc-testing")
  .dependsOn(compilerInterface)
```

Since `SourceInfos.empty` lives in `zinc-core`, `TestCallback.scala` cannot directly use it.

Friendseeker: I agree that a `null` is a code smell. I had added `TestSourceInfos` class providing an `ReadSourceInfos` implementation for testing purpose.

internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -1068,6 +1068,25 @@ private final class AnalysisCallback(
     }
   }
 
+  def getSourceInfos: SourceInfos = {
+    // Collect Source Info from current run
+    val sources = reporteds.keySet ++ unreporteds.keySet ++ mainClasses.keySet
+    val sourceToInfo = sources.map { source =>
```
Friendseeker, line 1073: - [ ] Issue: I did some more manual testing and found out that `reported` isn't always complete. When one of the recompiled file has compilation error, warnings and errors from other recompiled files may or may not get reported by the compiler.

I will try to think if this can be worked around. This is quite an ugly issue to deal with as the compiler indeed do not have responsibility to analyze every given file, when it already found out one of the file contains an error.

Friendseeker:
@adpi2 Reopened this PR and added a unit test.

@eed3si9n c.c. https://github.com/sbt/zinc/issues/932#issuecomment-2376210214 for reason for revisiting this PR. I am thinking if this can be merged along with #1309 as both involves adding `AnalysisCallback3`.

eed3si9n (approved):

eed3si9n:
LGTM. could you squash the commits and add problem and solution descriptions in the commit message plz?

Friendseeker:
@eed3si9n Squashed!

PR #1308, opened 2023-12-12
Add common VSCode metals temp folders to `.gitignore`

Friendseeker:
Before

Screenshot 2023-12-12 at 2 30 32 PM

(3k temp files from VSCode, metals and bloop detected by `Git`)

After

Screenshot 2023-12-12 at 2 30 43 PM

(0 temp file detected)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ zinc/src/test/resources/bin
 metals.sbt
 .bsp
 .DS_Store
+.vscode
+.metals
+.bloop


SethTisue:
In general we don't mind merging these sorts of additions, as there's no harm in it. But you might be happier just adding them to your personal global gitignore, then you won't need to have them added to every repo you work in.

Friendseeker:
> In general we don't mind merging these sorts of additions, as there's no harm in it. But you might be happier just adding them to your personal global gitignore, then you won't need to have them added to every repo you work in.

Ah thanks! I completely forgot that personal `.gitignore` is a thing!

You are a godsend. I was wondering why some metal devs were apparently working on Zinc related stuff and they did not complain about the 3k temp files.

eed3si9n (approved):
Yea, what Seth said.

Issue #963, opened 2021-01-27
MatchError in ApiPhase


dwijnand:
## steps

Sorry, I don't have exact steps, but after this happening a few times in a row in the same hour I thought I'd still report it to get the ball rolling.  Basically I rebased a branch and compiled...

## problem

```
[error] ## Exception when compiling 566 sources to /d/scala/build/quick/classes/library
[error] scala.MatchError: /d/scala/src/library/scala/collection/Seq.scala (of class scala.reflect.io.PlainFile)
[error] xsbt.API$ApiPhase.processScalaUnit(API.scala:62)
[error] xsbt.API$ApiPhase.processUnit(API.scala:57)
[error] xsbt.API$ApiPhase.apply(API.scala:53)
[error] scala.tools.nsc.Global$GlobalPhase.applyPhase(Global.scala:454)
[error] scala.tools.nsc.Global$GlobalPhase.run(Global.scala:401)
[error] xsbt.API$ApiPhase.run(API.scala:35)
[error] scala.tools.nsc.Global$Run.compileUnitsInternal(Global.scala:1515)
[error] scala.tools.nsc.Global$Run.compileUnits(Global.scala:1499)
[error] scala.tools.nsc.Global$Run.compileSources(Global.scala:1491)
[error] scala.tools.nsc.Global$Run.compileFiles(Global.scala:1605)
[error] xsbt.CachedCompiler0.run(CompilerBridge.scala:163)
[error] xsbt.CachedCompiler0.run(CompilerBridge.scala:134)
[error] xsbt.CompilerBridge.run(CompilerBridge.scala:39)
[error] sbt.internal.inc.AnalyzingCompiler.compile(AnalyzingCompiler.scala:92)
```

## expectation

Non-partial pattern matches...

## notes

After a `rm build/quick/classes/library/scala/collection/Seq*` it fixed it self, thankfully:

```
> mimaReportBinaryIssues
[info] root: mimaPreviousArtifacts not set, not analyzing binary compatibility
[info] running scala.tools.docutil.ManMaker "fsc, scala, scalac, scaladoc, scalap" /d/scala/target/scalaDist/resource_managed/main/doc/tools /d/scala/target/scalaDist/resource_managed/main/genman
[info] compiling 11 Scala sources to /d/scala/build/quick/classes/library ...
[info] compiling 17 Scala sources to /d/scala/build/quick/classes/library ...
[info] compiling 2 Scala sources to /d/scala/build/quick/classes/reflect ...
[...]

PR #1309, opened 2023-12-15
[1.x] Generate `AbstractZincFile` during `-sourcepath` workflow

### LINKED ISSUE #1304: MatchError in `ApiPhase.processScalaUnit` with `-sourcepath` compiler option
Setup:
  - small sbt project with two source files for classes A and B
  - set the `-sourcepath` compiler flag
  - change source file B
  - touch (save) the source file A without modifications

Zinc's incremental compilation runs the compiler on the changed source file B. For class A, the compiler finds both a classfile and a source file. Because the [source file has a more recent timestamp it creates a SourcefileLoader](https://github.com/scala/scala/blob/v2.13.12/src/compiler/scala/tools/nsc/symtab/SymbolLoaders.scala#L190-L192) for symbol A, which invokes `compileLate` and creates a plain `BatchSourceFile`.

The API phase crashes because it expects an `xsbti.VirtualFile`: https://github.com/sbt/zinc/blob/v1.9.5/internal/compiler-bridge/src/main/scala/xsbt/API.scala#L62

I'm encountering this when switching branches in scala/scala.

(edit from Dale: switched A and B in the setup, to match the description below.)

Friendseeker:
This PR fixes #1304, #963

### Issue

https://github.com/sbt/zinc/blob/5eb0de1fdfd921f0d1b6a9302fe42593e7e27922/internal/compiler-bridge/src/main/scala/xsbt/API.scala#L62

In certain codepath, `unit.source.file` is a `PlainFile`, which causes a `MatchError`.

### Fix

When `unit.source` is built, convert `unit.source.file` to `AbstractZincFile`.

### Validating the fix

The PR includes reproduction scripted test by [lrytz](https://github.com/lrytz). The test passes after the fix.


internal/compiler-bridge/src/main/scala/xsbt/CallbackGlobal.scala
```diff
@@ -78,6 +78,16 @@ sealed class ZincCompiler(settings: Settings, dreporter: DelegatingReporter, out
     extends CallbackGlobal(settings, dreporter, output)
     with ZincGlobalCompat {
 
+  override def getSourceFile(f: AbstractFile): BatchSourceFile = {
+    val file = f match {
+      case plainFile: PlainFile =>
+        val vf = callback.asInstanceOf[AnalysisCallback3].toVirtualFile(plainFile.file.toPath)
+        new ZincPlainFile(vf.asInstanceOf[PathBasedFile])
```
Friendseeker: The conversion logic is... hacky.

I would appreciate any feedback suggesting a better solution.

internal/compiler-bridge/src/main/scala/xsbt/CallbackGlobal.scala
```diff
@@ -78,6 +79,19 @@ sealed class ZincCompiler(settings: Settings, dreporter: DelegatingReporter, out
...
     with ZincGlobalCompat {
 
+  override def getSourceFile(f: AbstractFile): BatchSourceFile = {
+    val file = (f, callback) match {
+      case (plainFile: PlainFile, callback3: AnalysisCallback3) =>
+        val virtualFile = callback3.toVirtualFile(plainFile.file.toPath)
+        virtualFile match {
+          case underlying: PathBasedFile => new ZincPlainFile(underlying)
+          case _                         => f
+        }
```
dwijnand: I think this is `AbstractZincFile(callback3.toVirtualFile(plainFile.file.toPath))`, btw.

dwijnand (changes requested):
See https://github.com/sbt/zinc/issues/1304#issuecomment-1875188256

eed3si9n (approved):

Friendseeker:
@eed3si9n While the PR is in `Changes requested` status, would it be possible for you to dismiss the stale review and force merge it if everything looks okay?

My concern is that since this PR involves `AnalysisCallback3` and https://github.com/sbt/zinc/pull/1306 is already merged in, if a Zinc release is made before the PR is merged, we would need to introduce `AnalysisCallback4`.

eed3si9n:
Yea. Now that AnalysisCallback is AnalysisCallback3, we should just land this.

Issue #537, opened 2018-05-02
-opt:l:inline breaks incremental compilation


eed3si9n:
This was originally reported by @eparejatobes as https://github.com/sbt/sbt/issues/4125#issuecomment-385992117

*Copying here the relevant portion of the link above; see [ohnosequences/sbt4125](https://github.com/ohnosequences/sbt4125) for more*

## steps

```scala
scalacOptions ++= Seq(
  "-unchecked",
  "-opt-warnings:_",
  "-opt:l:inline",
  "-opt-inline-from:**"
)
```

Open a shell and 

``` bash
git checkout "x=1"
```

Now run `sbt` on a different shell and do `~test`. Once it runs for the first time,

``` bash
git checkout "x=2"
```

Take a look at the test passing, and the definitions of `x` and `y`.

## problem

> Incremental compilation when used together with the optimizer generates incorrect bytecode: modifications in methods/vals are not propagated to wherever they are being inlined.

## expectation

> The incremental compiler should not generate incorrect bytecode.

I'd say we revise the expectation to be that change to `x=2` is reflected in compilation and run.

## notes

I am not sure how prevalent these optimization flags are, but we should probably print warning or invalidate all sources, similar to 50% threshold mechanism.

smarter:
Basically, if a method is inlineable, its body is part of its API signature. We deal with that in dotty by storing the pretty-printed body of the method in a signature in ExtractAPI, though it'd be better to store a hash: https://github.com/lampepfl/dotty/blob/a3c4ec793545fd2b922eb469f7fc4d4a7d26b05f/compiler/src/dotty/tools/dotc/sbt/ExtractAPI.scala#L593-L602

eed3si9n:
Interesting. That would give better usability compared to invalidating all sources :)

eparejatobes:
> Basically, if a method is inlineable, its body is part of its API signature (...)

What about `def f = { g(0): @inline }`?

smarter:
Does that actually do anything in Scala 2.12 ?

eparejatobes:
> Does that actually do anything in Scala 2.12 ?

per the docs it should, and I just checked it in the repo linked above; `javap` yields

### with call-site @inline

```
public int z();
    Code:
       0: iconst_1
       1: getstatic     #19                 // Field issues/sbt4125/inlineFunTimes$.MODULE$:Lissues/sbt4125/inlineFunTimes$;
       4: ifnonnull     9
       7: aconst_null
       8: athrow
[...]

PR #1310, opened 2023-12-15
Invalidate sources that depends on `@inline` methods

### LINKED ISSUE #4125: -opt:l:inline breaks incremental compilation
Given

``` scala
@inline def x = 2

// somewhere else
def z = x + 1 
```

changes in `x` above after a successful compilation will not propagate to `z`; if we make `def x = 3` then we will have `z == 3`, not `4`.

I'm aware of the [2.12 Release Notes](https://www.scala-lang.org/news/2.12.0/) stating that

> We recommend enabling inlining only in production builds, as sbt’s incremental compilation does not track dependencies introduced by inlining.

but IMHO mentioning it upfront doesn't make it less of a bug: either incremental compilation works correctly when inlining is enabled (which it should), or we are given a way of disabling it (as it was already asked for in #1282).

Friendseeker:
This PR fixes https://github.com/sbt/sbt/issues/4125, https://github.com/sbt/zinc/issues/537.

### Issue

When `@inline` method implementation changes, the public API of `@inline` method does not change, hence Zinc does not invalidate sources that depends on `@inline` method.

### Solution

The behaviour of `@inline` method is similar to macro methods. Hence, we can mark `@inline` method as macro and reuse existing Zinc invalidation logic for macro invalidation.

### TODOs

- This PR does not handle call-site `@inline`. To handle that case, we need to follow https://github.com/sbt/zinc/issues/537#issuecomment-421753875 and drop call-site `@inline` at compiler side.
- Even without `@inline` annotation, compiler may still sometimes inline a method. The compiler keep a log of inlined methods, so in a follow up PR, we can utilize that information instead of directly checking `@inline` annotation.

### DIFF
--- a/internal/compiler-bridge/src/main/scala-2.11/xsbt/Compat.scala
+++ b/internal/compiler-bridge/src/main/scala-2.11/xsbt/Compat.scala
@@ -31,6 +31,10 @@ object Compat {
 
   // IMain in 2.13 accepts ReplReporter
   def replReporter(settings: Settings, writer: PrintWriter) = writer
+
+  implicit final class SettingsCompat(val settings: Settings) extends AnyVal {
+    @inline final def optInlinerEnabled: Boolean = false
+  }
 }
 
 /** Defines compatibility utils for [[ZincCompiler]]. */
--- a/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/ExtractAPI.scala
@@ -522,6 +522,7 @@ class ExtractAPI[GlobalType <: Global](
     val absOver = s.hasFlag(ABSOVERRIDE)
     val abs = s.hasFlag(ABSTRACT) || s.hasFlag(DEFERRED) || absOver
     val over = s.hasFlag(OVERRIDE) || absOver
+    val hasInline = s.annotations.exists(_.symbol.tpe == typeOf[scala.inline])
     new xsbti.api.Modifiers(
       abs,
       over,
@@ -529,7 +530,7 @@ class ExtractAPI[GlobalType <: Global](
       s.hasFlag(SEALED),
       isImplicit(s),
       s.hasFlag(LAZY),
-      s.hasFlag(MACRO),
+      s.hasFlag(MACRO) || hasInline,
       s.hasFlag(SUPERACCESSOR)
     )
   }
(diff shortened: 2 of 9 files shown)


eed3si9n (approved):
Thanks!

Friendseeker:
Hi Eugene! I was [asking users on Scala discord](https://discord.com/channels/632150470000902164/632150470000902166/1186040861037969418) about whether this issue bothers Scala users in practice.

The feedback I got is that basically everyone already has setup to only use optimizer when producing release binary. Hence this issue do not actually impact anyone.

Should we instead not proceed with the PR and instead close the two linked issues https://github.com/sbt/sbt/issues/4125, https://github.com/sbt/zinc/issues/537 as won't fix?

eed3si9n:
Since the change itself is relatively small, I see no reason we should reject a good change. While it's true, that once you start using optimizer many things would break so people have turned from past injuries to turn them off during normal development, but perhaps changes like this would gradually improve the situation.

lrytz:
This PR looks right and is an improvement. However in the vast majority of cases methods are not inlined because they carry the `@inline` annotation, but due to other inliner heuristics. The official documentation even recommends to avoid using `@inline` in general (https://docs.scala-lang.org/overviews/compiler-options/optimizer.html#inliner-heuristics-and-inline). So I don't think this PR will help much in practise, but it can be considered a first step?

Friendseeker:
> However in the vast majority of cases methods are not inlined because they carry the `@inline` annotation, but due to other inliner heuristics.

I had never worked with optimizer, but I think it would be possible for optimizer to provide information for each inlining that took place (caller & callee, along with the class name caller and callee belongs to) and store them in `CompilationUnit`, and zinc `Analyzer` phase can then use the information to handle inlining dependency. (e.g. via introducing a new dependency called `DependencyByInlining` and handle them similarly as `DependencyByInheritance`.). This would catch inlining dependency due to heuristics.

If that solution is indeed reasonable, we can then decide whether we should implement the solution in a follow up PR.

dwijnand:
It's a little confusing that `isMacro` doesn't mean isMacro any more :-/

lrytz:
The inliner can already produce a log, so adding the right calls to record dependencies should be doable.

```
➜ sandbox cat A.scala
object A {
  def f = 1
}

object B {
  def g = A.f
}

object C {
  def h = B.g
}
➜ sandbox qsc -opt:inline:'**' A.scala -Vinline _
Inlining into B$.g: inlined A$.f (the callee is a small trivial method). Before: 7 ins, after: 22 ins.
Inlining into C$.h: inlined B$.g (the callee is a forwarder or alias method). Before: 7 ins, after: 40 ins.
```

But I'm not convinced it's worth the effort. Enabling the optimizer in development is discouraged as it increases compilation times. How about issuing a warning when zinc performs an incremental compilation (not all sources) with the optimizer enabled?

PR #1312, opened 2023-12-21
Do not use `invalidationResults` to compute next invalidations

### LINKED ISSUE #1311: Overcompilation with incremental, mixed Java/Scala file edits in 1.9.0-RC1
This is a follow up to [this issue](https://github.com/sbt/sbt/issues/6183) and [the fix for the related zinc issue](https://github.com/sbt/zinc/pull/1182).  I took jdsalchow's [minimum repro repo](https://github.com/jdsalchow/compile_loop), I set the sbt version to 1.3.13 (known working version from [our affected project](https://github.com/NetLogo/NetLogo)), and things worked as normal:

```
sbt:compile_loop> compile
[info] Compiling 3 Scala sources and 5 Java sources to /CCL-Repos/compile_loop/target/scala-2.13/classes ...
[info] Non-compiled module 'compiler-bridge_2.13' for Scala 2.13.6. Compiling...
[info]   Compilation completed in 8.224s.
[success] Total time: 12 s, completed May 3, 2023, 2:36:45 PM
```

I edited `A.scala` and `ABase.java` (just whitespace changes, no real code edits) and saw:

```
sbt:compile_loop> compile
[info] Compiling 1 Scala source to /CCL-Repos/compile_loop/target/scala-2.13/classes ...
[info] Compiling 1 Scala source and 1 Java source to /CCL-Repos/compile_loop/target/scala-2.13/classes ...
[success] Total time: 1 s, completed May 3, 2023, 2:37:04 PM
```

Which seems fine.

Then I updated the sbt version to `1.9.0-RC1`.  I cleaned the repo and the first compile was the same as before:

```
sbt:compile_loop> compile
[info] compiling 3 Scala sources and 5 Java sources to /CCL-Repos/compile_loop/target/scala-2.13/classes ...
[info] Non-compiled module 'compiler-bridge_2.13' for Scala 2.13.6. Compiling...
[info]   Compilation completed in 9.728s.
[success] Total time: 12 s, completed May 3, 2023, 2:38:16 PM
```

But after the non-code whitespace edits I saw:

```
sbt:compile_loop> compile
[info] compiling 1 Scala source and 1 Java source to /CCL-Repos/compile_loop/target/scala-2.13/classes ...
[info] compiling 1 Scala source and 1 Java source to /CCL-Repos/compile_loop/target/scala-2.13/classes ...
[info] compiling 1 Scala source and 1 Java source to /CCL-Repos/compile_loop/target/scala-2.13/classes ...
[info] compiling 3 Scala sources and 5 Java sources to /CCL-Repos/compile_loop/target/scala-2.13/classes ...
[success] Total time: 1 s, completed May 3, 2023, 2:38:30 PM
```

It's now invalidating all Scala and Java files and recompiling the whole source tree.  For this small repo that only takes a few seconds, but for our repo it takes 90 seconds or more.  Since things were working as expected in 1.3.13 this seems like a regression.

Editing one Scala file and one Java file seems necessary to trigger the issue.  If I modify just one type or the other it compiles just the changed files.

Friendseeker:
### Issue

`invalidationResults` in `AnalysisCallback` is computed twice, once in API phase, once in dependency phase. It is not updated again after Java analysis, hence `invalidationResults` contains incorrect information for Java sources.

### Fix

Do not use `invalidationResults` when computing `CompileCycleResult`.

### Trivia

I am not 100% certain yet, but this is probably the root cause of https://github.com/sbt/sbt/issues/6183, as

- `invalidationResults` is added in Zinc 1.4.0 and https://github.com/sbt/sbt/issues/6183 appears after Zinc 1.4.0 is released
- [Reproduction](https://github.com/jdsalchow/compile_loop) of infinite compile by jdsalchow now terminates before `invalidateTransitively` flag is set with the PR change.

c.c. @LaCuneta This should fix #1311 or at least partially fix it (if #1311 has multiple causes).

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -937,7 +937,7 @@ private final class AnalysisCallback(
         if (!writtenEarlyArtifacts) // writing implies the updates merge has happened
           mergeUpdates() // must merge updates each cycle or else scalac will clobber it
       }
-      incHandler.completeCycle(invalidationResults, getAnalysis)
+      incHandler.completeCycle(None, getAnalysis)
     } else {
       throw new IllegalStateException(
         "can't call AnalysisCallback#getCycleResultOnce more than once"
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -1090,14 +1090,14 @@ private final class AnalysisCallback(
 
   override def dependencyPhaseCompleted(): Unit = {
     val incHandler = incHandlerOpt.getOrElse(sys.error("incHandler was expected"))
-    if (invalidationResults.isEmpty) {
+    if (earlyOutput.isDefined && invalidationResults.isEmpty) {
       val a = getAnalysis
       val CompileCycleResult(continue, invalidations, merged) =
         incHandler.mergeAndInvalidate(a, false)
       // Store invalidations and continuation decision; the analysis will be computed again after Analyze phase.
       invalidationResults = Some(CompileCycleResult(continue, invalidations, Analysis.empty))
       // If there will be no more compilation cycles, store the early analysis file and update the pickle jar
-      if (earlyOutput.isDefined && !continue && lookup.shouldDoEarlyOutput(merged)) {
+      if (!continue && lookup.shouldDoEarlyOutput(merged)) {
         writeEarlyArtifacts(merged)
       }
     }
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -937,7 +937,7 @@ private final class AnalysisCallback(
         if (!writtenEarlyArtifacts) // writing implies the updates merge has happened
           mergeUpdates() // must merge updates each cycle or else scalac will clobber it
       }
-      incHandler.completeCycle(None, getAnalysis)
+      incHandler.completeCycle(invalidationResults, getAnalysis)
     } else {
       throw new IllegalStateException(
         "can't call AnalysisCallback#getCycleResultOnce more than once"
--- a/zinc/src/sbt-test/source-dependencies/cyclic-dependency/A.scala
+++ b/zinc/src/sbt-test/source-dependencies/cyclic-dependency/A.scala
@@ -1,0 +1,13 @@
+/*
+ * Zinc - The incremental compiler for Scala.
+ * Copyright Scala Center, Lightbend, and Mark Harrah
+ *
+ * Licensed under Apache License 2.0
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+class A() extends ABase()
+
--- a/zinc/src/sbt-test/source-dependencies/cyclic-dependency/ABase.java
+++ b/zinc/src/sbt-test/source-dependencies/cyclic-dependency/ABase.java
@@ -1,0 +1,14 @@
+/*
+ * Zinc - The incremental compiler for Scala.
+ * Copyright Scala Center, Lightbend, and Mark Harrah
+ *
+ * Licensed under Apache License 2.0
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+public abstract class ABase {
+    H h;
+}
--- a/zinc/src/sbt-test/source-dependencies/cyclic-dependency/H.scala
+++ b/zinc/src/sbt-test/source-dependencies/cyclic-dependency/H.scala
@@ -1,0 +1,12 @@
+/*
+ * Zinc - The incremental compiler for Scala.
+ * Copyright Scala Center, Lightbend, and Mark Harrah
+ *
+ * Licensed under Apache License 2.0
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+class H() extends HBase()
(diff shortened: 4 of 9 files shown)


internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -937,7 +937,7 @@ private final class AnalysisCallback(
         if (!writtenEarlyArtifacts) // writing implies the updates merge has happened
           mergeUpdates() // must merge updates each cycle or else scalac will clobber it
       }
-      incHandler.completeCycle(invalidationResults, getAnalysis)
+      incHandler.completeCycle(None, getAnalysis)
```
eed3si9n: I see that this change may fix the particular bug regarding Scala / Java mixed situation, but this particular change doesn't feel right. Isn't this called once as the name suggest?

Friendseeker: Would you mind to elaborate on how `getCycleResultOnce` or `completeCycle` being called once is significant?

I would like to learn more about why the change may not be the right thing to do and investigate alternative approaches.

eed3si9n: Generally, the code in this area was contributed in https://github.com/sbt/zinc/pull/744 to implement build pipelining. I was involved in the code rewrite process to bring this into Zinc, but it's been a long while and I barely understood it back then.

The reason why the change didn't sit right with me (sorry about this vagueness) is because `completeCycle` is defined as:

```scala
    /**
     * Merge latest analysis as of analyzer into pruned previous analysis and inform file manager.
     */
    def completeCycle(
        prev: Option[CompileCycleResult],
        partialAnalysis: Analysis
    ): CompileCycleResult
```

So when there was a previous result either via API phase, or from the dependency phase, simply dropping this information would have some breakage down the road. If we're going to pretend there was no previous `CompileCycleResult`, my guess is we should do so elsewhere, like inside of `completeCycle(...)`.

>  Would you mind to elaborate on how `getCycleResultOnce` or `completeCycle` being called once is significant?

Maybe I confused myself the once being important to the timing, but I did find an interesting comment made by 2020 me - https://github.com/sbt/zinc/pull/744#issuecomment-623628523:

> If we try to pipeline mixed Java/Scala subprojects, the Analysis information created by JavaAnalyze will not be available to downstream subprojects. Instead it will use the Analysis information created via `-Ypickle-java + xsbt-dependency` phase etc. I'm not sure if we lose some accuracy because of this.

and in https://github.com/sbt/sbt/pull/5703 I implemented `compileJava` task in sbt, which is called only if pipelining is enabled. You wrote:

> [invalidationResults] is not updated again after Java analysis, hence `invalidationResults` contains incorrect information for Java sources.

so I guess we do have loss of information when there's a mix of Scala and Java, and it's not pipelined. Since we're the one running [AnalyzingJavaCompiler](https://github.com/sbt/zinc/blob/develop/zinc/src/main/scala/sbt/internal/inc/javac/AnalyzingJavaCompiler.scala) I wonder if it's possible to somehow pass in the right information to AnalysisCallback?

Friendseeker: Thanks for your explanation, Eugene!

> So when there was a previous result either via API phase, or from the dependency phase, simply dropping this information would have some breakage down the road. If we're going to pretend there was no previous CompileCycleResult

If we discard `invalidationResult` when calling `completeCycle`, then `completeCycle` would just call `mergeAndInvalidate`, which reconstructs the most up-to-date analysis from `partialAnalysis` (being `getAnalysis`) and pruned analysis, and then compute `CompileCycleResult` using the reconstructed analysis.

Since `partialAnalysis` contains information extracted from `API` and `Dependency` phase, I thought  `completeCycle(None, getAnalysis)` should still correctly account for information extracted from `API` and `Dependency` phase.

> Since we're the one running [AnalyzingJavaCompiler](https://github.com/sbt/zinc/blob/develop/zinc/src/main/scala/sbt/internal/inc/javac/AnalyzingJavaCompiler.scala) I wonder if it's possible to somehow pass in the right information to AnalysisCallback?

Correct me if I misinterpreted this comment, but if the intention of the comment is to making sure that `invalidationResults` correctly accounts Java analysis in the first place by this point that `completeCycle` is called, I think we still have to call `mergeAndInvalidate` in `AnalyzingJavaCompiler`, which basically yields a very similar solution (instead of calling `mergeAndInvalidate` in `completeCycle`, we call it before in `AnalyzingJavaCompiler`.) I can see the benefit of that solution (it maintains the invariant that `invalidationResults` is always up-to-date), but that requires adding a new method to `AnalysisCallback` as `mergeAndInvalidate` was previously not exposed in `AnalysisCallback`.

---

This might be out of scope of this PR and might be a real stretch. But... can we get rid of `invalidationResults` altogether? When we need to know about invalidation, we can always just reuse partial and pruned analysis to compute invalidation (via calling `mergeAndInvalidate`). In fact, other than the one call to `incHandler.completeCycle(invalidationResults, getAnalysis)`, the exact content of `invalidationResults` is fetched nowhere. Rather various usage of `invalidationResults`  merely checks if `invalidationResults` is empty or not. Less variable equals to less bug.

eed3si9n: My guess is that `invalidationResults` var is/was necessary to hand of these information mid-way through the last cycle.

pipelining

I guess one light change you could try is adding `earlyOutput.isDefined &&` to
https://github.com/sbt/zinc/blob/1c809a60a11a710aa4b016aa28f7e8a44da68a98/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala#L1093
that way we can avoid calling `incHandler.mergeAndInvalidate(a, false)` needlessly if we're not gonna use it, and `invalidationResults` won't be assigned.

internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -1090,14 +1090,14 @@ private final class AnalysisCallback(
 
   override def dependencyPhaseCompleted(): Unit = {
     val incHandler = incHandlerOpt.getOrElse(sys.error("incHandler was expected"))
-    if (invalidationResults.isEmpty) {
+    if (earlyOutput.isDefined && invalidationResults.isEmpty) {
```
eed3si9n, line 1093: Could you also revert the line 940 change also?

eed3si9n (approved):
Thanks!

PR #1315, opened 2023-12-26
add scalafmt convertToNewSyntax setting

xuwei-k:
prepare Scala 3.4

- https://github.com/lampepfl/dotty/pull/18813


eed3si9n (approved):
Thanks!

PR #1316, opened 2023-12-27
Invalidate macro client when type parameter changes

### LINKED ISSUE #1171: Undercompilation when macros are used (mainargs)
## Problem

When using the macro-based CLI library mainargs, I noticed an undercompilation issue

When changing the `@doc` annotation of one argument in `foo.Config`, a recompile results in only one recompiled file `Config.scala`. 

The file `ConfigParser.scala` gets not recompiled, yet it contains a `lazy val usageText = parser.helpText()`. `parser` is a `private[this] lazy val parser: ParserForClass[Config] = mainargs.ParserForClass[Config]`.

The expectation is, that also `ConfigParser.scala` gets recompiled. This does not happen. This results in an outdate/stale help message.

## Info to reproduce

* This reproducer uses the following versions:
  * Mill 0.10.11
  * Zinc 1.8.0
  * Scala 2.13.10
  * mainargs 0.3.0

* The reproducer: https://github.com/lefou/mainargs-undercompilation-reproduction

## Steps to reproduce

* Compile and run reproduction project

```
$ mill clean
[1/1] clean
$ mill foo.run --help
[33/46] foo.compile
Compiling compiler interface...
[info] compiling 2 Scala sources to /tmp/mainargs-repro/out/foo/compile.dest/classes ...
[info] done compiling
[46/46] foo.run
apply
  --help  Print this help message.

```

* Change the contents of the `@doc` annotation 

```
$ sed -i "s/Print this help message/Print another message/" foo/src/Config.scala
```

* Re-run the reproduction build and notice, that only one source file gets recompiled and the `--help` output has not changed.

```
$ mill foo.run --help
[33/46] foo.compile
[info] compiling 1 Scala source to /tmp/mainargs-repro/out/foo/compile.dest/classes ... 
[info] done compiling
[46/46] foo.run
apply
  --help  Print this help message.

```

## expectation

The compiler should also recompile the file which depends on `Config.scala`.

Friendseeker:
Fixes #1171

### Issue

When a macro takes some type parameter `T`, e.g. `def hasAnyField[T]: Boolean = macro hasAnyFieldImpl[T]`. Changes in `T` require macro client to be invalidated if macro implementation inspects `T` (e.g. fetching members of `T`).

### Fix

When implementation of `T` is invalidated, invalidate all macros clients that use `T` as type parameter using invalidation logic from https://github.com/sbt/sbt/pull/1778


zinc/src/sbt-test/macros/macro-type-change-3/pending
```diff
@@ -0,0 +1,3 @@
+> app/run false
```
Friendseeker: Plan: leave this part in the future (either in another PR, or as Dale suggested... never). This is the same problem in https://github.com/sbt/zinc/pull/1282#issuecomment-1832399563, where a mechanism to track transitive external dependency is needed.

eed3si9n (approved):
overall this LGTM

a meta observation is that this adds a new enum value to the compiler interface and changes the bridge code, and given that Scala 3.x and 2.13.12+ currently use compiler bridge implementation of their own, this fix will only work for Scala 2.11, 2.12, and 2.13.11 (and some range below 11), and potentially some future versions of Scala 3.x or 2.13.x if the respective maintainers accept similar changes. To avoid confusion, perhaps we should also try to get some feedback from Scala 2 and Scala 3 maintainers like @nicolasstucki (Dale's already listed a reviewer), so if they have any concerns we'll try to coordinate it here.

internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala
```diff
@@ -110,6 +110,7 @@ final class Dependency(val global: CallbackGlobal) extends LocateClassFile with
     val memberRef = processDependency(DependencyByMemberRef, false) _
     val inheritance = processDependency(DependencyByInheritance, true) _
     val localInheritance = processDependency(LocalDependencyByInheritance, true) _
+    val macroExpansion = processDependency(DependencyByMacroExpansion, false) _
```
nicolasstucki: I would sugest naming it `scala2MacroExpansion` to avoid any confusion if we need something similar in the future for Scala 3 macros.

nicolasstucki (commented):
Otherwise LGTM

dwijnand (approved):

bishabosha:
I will test the added test case in scala 3, see what is needed on that side

SethTisue:
@bishabosha ping

bishabosha:
I added the test cases in https://github.com/dotty-staging/dotty/commit/304c41a714a50016b443c87640081a0eb479682f, which fail, so we would need to reimplement the logic in Scala 3 - is there a special reason to introduce the new `DependencyContext` enum case? or is it just for nicer debugging? or that inverse invalidation computation can't be replicated without adding the case

sjrd:
Posting my 2 cents at @bishabosha's request: in general, you cannot solve this problem using the current approach. A macro can look at *anything*. Even if now you're solving the case where it looks at the members of a type given as a parameter, you're probably not solving the case where it looks at the members of the type of a local variable in a block that is passed to it as argument (to name a random example).

There are only 2 reasonable choices when it comes to macro expansion:
* Assume macros behave as a regular methods, and therefore record dependencies exactly like regular methods. I macros look at other things, there will be undercompilation.
* Actually track *exactly* what any particular macro expansion looks at.

Anything else in-between places an arbitrary boundary on what you can and cannot track, and that boundary will be broken next month by the next macro that comes in.

In order to actually track exactly what macro expansions look at, you would need something very similar to what we use in the Scala.js optimizer and emitter: a so-called "knowledge guardian" that gates every call to the `Quotes` API and records all the calls made by a macro as its dependencies. If you want to know more about that, you can find all the details in my only research paper ever: https://lampwww.epfl.ch/~doeraene/publications/oopsla16-incremental-optimizer.pdf

Friendseeker:
> I added the test cases in [dotty-staging/dotty@304c41a](https://github.com/dotty-staging/dotty/commit/304c41a714a50016b443c87640081a0eb479682f), which fail, so we would need to reimplement the logic in Scala 3 - is there a special reason to introduce the new `DependencyContext` enum case? or is it just for nicer debugging? or that inverse invalidation computation can't be replicated without adding the case

The inverse invalidation indeed can't be replicated. It is similar to the inheritance case, where a special type of dependency is needed to encode necessary information for invalidation.

> Posting my 2 cents at @bishabosha's request: in general, you cannot solve this problem using the current approach

Indeed `typesTouchedDuringMacroExpansion` is currently, only an approximation. However, this is still a step over the status quo. I still need to read the paper in detail, but I believe future work on bringing knowledge guardian to Dotty (if it isn't yet done) can directly reuse the PR's logic by replacing `typesTouchedDuringMacroExpansion` with the exact types gathered from `Quotes` API calls.

eed3si9n:
@sjrd wrote:

> There are only 2 reasonable choices when it comes to macro expansion:
> 
> * Assume macros behave as a regular methods, and therefore record dependencies exactly like regular methods. I macros look at other things, there will be undercompilation.
> * Actually track _exactly_ what any particular macro expansion looks at.
> 
> Anything else in-between places an arbitrary boundary on what you can and cannot track, and that boundary will be broken next month by the next macro that comes in.

I think I agree with the overall points, but there's some subtlety due to the nature of Zinc. To prevent invalidation spreading like wildfire, Zinc uses namehashing as a heuristic optimization to limit the invalidation only to the client code that uses the name of the changed API (https://www.slideshare.net/EugeneYokota/analysis-of-zinc-scalasphere-2019#24).

I think adopting Strategy 1 and assume that macros behave like regular method, or inlined methods like Scala 3, is fine. But what we should do, is basically treat lexical elements going into the macro application as an input parameter of a tree generator. For the case of mainargs macro the callsite looks like:

```scala
  private[this] lazy val parser: ParserForClass[Config] = mainargs.ParserForClass[Config]
```

so the input in this case is just the parameter `[T]` of `def apply[T]: ParserForClass[T]`. So I feel like the solution presented in this PR is still somewhat within the realm of Strategy 1, just undoing the namehashing optimization for macro application, similar to the special-cased inheritance cases.

Issue #1129, opened 2022-09-18
Stamp locking slows down large builds


viveksjain:
## steps

1. Pick a large scala repo, like https://github.com/apache/spark
2. Do a full compilation
3. Make a one-line change in some source file
4. See how long a dependent test takes to recompile

## problem

I was investigating incremental scala compilation performance, which takes O(30s) for one-line change on a large internal repo and found that quite a large fraction of time is spent blocked on Stamp locks. About 16s on this global Stamper lock https://github.com/sbt/zinc/blob/fc9e66ac83bc2bb99937fd79a657e4ad79858f06/internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala#L248 and 8s on the InitialStamps lock https://github.com/sbt/zinc/blob/fc9e66ac83bc2bb99937fd79a657e4ad79858f06/internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala#L420. Note that the lock time numbers do not correspond to wall-clock because zinc has internal parallelism, calling from multiple threads.

## notes

We can improve the former by not using `timeWrapBinaryStamps` when calling zinc, but the latter seems to be used internally in zinc to wrap the `ReadStamps` that is passed in and cannot be removed. I am not very familiar with the code, but initial ideas:
1. The InitialStamps lock seems to just be for `getAll*Stamps`, but I couldn't actually see where these methods are used. Is supporting them really necessary?
5. The global Stamper lock seems to be to cache hash results. The shared data structure is the `cache` though, so it should lock on that. Using something like ConcurrentHashMap.update would likely be even faster.

eed3si9n:
@viveksjain Thanks for the report. For the sake of reproducibility, could you suggest exact line someone could make in the Spark code base that would reproduce this behavior please?

viveksjain:
Good question, when I do spark incremental builds with mvn or sbt it seems I'm not actually able to reproduce this issue (sbt takes 15-20s but profiling shows only 300ms is spent on Stamper). I don't know enough about how mvn/sbt calls into zinc, and how it's different from how we are doing it in our internal repo (with custom bazel rules), to properly understand why. Does sbt maintain a global `timeWrapBinaryStamps` instance so that pretty much everything remains cached?

I guess as it currently stands this issue isn't particularly actionable without a repro, feel free to close it.

eed3si9n:
[...]

PR #1317, opened 2023-12-28
Use `ConcurrentHashMap` in Stamp.scala

Friendseeker:
### Issue

The two `synchronized` locks in `Stamp.scala` sometimes take too long to be grabbed (#1129)

### Mitigation

Not sure about the root cause of the issue. But we can at least make lock more granular via using `ConcurrentHashMap` instead of `synchronized`.

As bonus this change addresses [a potential race condition](https://discord.com/channels/632150470000902164/632150470000902166/1189795104207163433). When `InitialStamps.source` and `InitialStamps.getAllSourceStamps` is called concurrently, the access is not thread-safe. Such access pattern is probably very rare / never happens in practice, but nonetheless an improvement that it is addressed.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala
@@ -237,23 +237,27 @@ object Stamper {
   }
 
   private[sbt] def timeWrap(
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

Issue #578, opened 2018-08-16
Make `this` names context-dependent


jvican:
It occurred to me that because of the name heuristics, `this` can be a common source of conservative recompilation when changes in the source code happens. We fixed a similar issue with constructors in https://github.com/sbt/zinc/pull/288, and the same rationale applies here. I'm optimistic that, just like that change, this improvement can reduce unnecessary compilations and increase the efficiency of the recompilations.

jvican:
Another common name is `package`, which we may just want to outright remove.

jvican:
Like `this`, synthesized defs like the ones generated by default arguments (and maybe delayed init?) should also be prepended with the class name they are owned by to minimize recompilations. Currently: `<init>$default$1`, should be: `foo;MyBar;<init>$default$1` just as in the PR in 
https://github.com/sbt/zinc/pull/288 (/cc @smarter)

jozanek:
I think this can be closed: `this` is never recorded as a used name. `This`/`Super` trees carry the enclosing class symbol, so both bridges record the class name ([ExtractUsedNames.scala#L283-L287](https://github.com/sbt/zinc/blob/b4695540f189db6c0456177b88a0bf8f3fcc71ff/internal/compiler-bridge/src/main/scala/xsbt/ExtractUsedNames.scala#L283-L287)), and `ExtractAPI` never emits a definition named `this`. A throwaway spec covering `this.x`, `A.this.x`, `this.type`, self aliases and secondary constructors confirms this.

Only leftover from the second comment: constructor default getters are recorded as `<init>$default$1` but named `p;A;init;default;1` in `ExtractAPI`, so that name hash is inert. Benign, since any API change to a default also changes `p;A;init;`. Happy to open a small PR aligning `mangledName` if wanted.

eed3si9n:
Yea, let's close this and open a new issue for default parameters.

eed3si9n:
The default argument is already fixed by https://github.com/sbt/zinc/pull/1324.

PR #1324, opened 2024-01-03
Avoid spurious recompilations when unrelated constructor with default argument changes

Friendseeker:
### Issue

Constructor sometimes has name `<init>$default$number` (e.g.  `<init>$default$1`,  `<init>$default$2`) instead of `<init>`. Hence two unrelated constructors have the same name.

#288 ensured that two unrelated `<init>` constructor have different names by prepending constructor name with class name, however #288 did not handle `<init>$default$`.

### Fix

Prepend `<init>$default$` with the class name.

c.c. https://github.com/sbt/zinc/pull/288, https://github.com/sbt/zinc/issues/578#issuecomment-417599988

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/ClassName.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/ClassName.scala
@@ -58,6 +58,9 @@ trait ClassName extends Compat {
   protected def constructorNameAsString(cls: Symbol): String =
     cls.fullName(';') ++ ";init;"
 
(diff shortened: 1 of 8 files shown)


Friendseeker:
I am deliberating on whether the PR is actually a good change to have.

This particular case `<init>$default$` is rare (unlike `<init>` which is always a used name), while the extra logic introduced by this PR adds overhead to all `Def` handling. It also makes API phase slightly more confusing.

---

I shall request for review and let reviewer determine if this is actually good to have.

eed3si9n (approved):
Nice!

PR #1331, opened 2024-01-23
Don't use autoBoot / filterLibrary for 2.13 and 3

lrytz:
Passing `-bootclasspath /path/to/scala-library.jar` to the Scala compiler is only done for historical reasons. Putting the Scala library on the ordinary classspath is equivalent.

This commit adds a `noboot` factory for ClasspathOptions which doesn't use `-bootclasspath` for Scala 2.13 and Scala 3.

This change was made in the context of [SIP-51, unfreeze the Scala library](https://docs.scala-lang.org/sips/drop-stdlib-forwards-bin-compat.html) / https://github.com/sbt/sbt/pull/7480, but it's just a cleanup and not actually required.

### DIFF
--- a/internal/zinc-compile-core/src/main/java/xsbti/compile/ClasspathOptionsUtil.java
+++ b/internal/zinc-compile-core/src/main/java/xsbti/compile/ClasspathOptionsUtil.java
@@ -16,7 +16,6 @@ package xsbti.compile;
  * that create typical classpath options based on the desired use-case.
  */
 public interface ClasspathOptionsUtil {
-
     /**
      * Define a manual {@link ClasspathOptions} where the client manages everything.
      */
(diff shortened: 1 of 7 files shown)


internal/zinc-compile-core/src/main/java/xsbti/compile/ClasspathOptionsUtil.java
```diff
@@ -34,6 +33,16 @@ public static ClasspathOptions boot() {
         return ClasspathOptions.of(true, false, false, true, true);
     }
 
+    /**
+     * Define {@link ClasspathOptions} where the Scala standard library is present in the classpath.
+     */
+    public static ClasspathOptions noboot(String scalaVersion) {
+        if (scalaVersion.startsWith("3.") || scalaVersion.startsWith("2.13."))
```
eed3si9n: For Scala 4 preparedness:

```suggestion
        if (!scalaVersion.startsWith("2.") || scalaVersion.startsWith("2.13."))
```

internal/zinc-compile-core/src/main/java/xsbti/compile/ClasspathOptionsUtil.java
```diff
@@ -46,6 +55,19 @@ public static ClasspathOptions auto() {
...
     }
 
+    /**
+     * Define auto {@link ClasspathOptions} where:
+     * 1. the Scala standard library is present in the classpath;
+     * 2. the Compiler JAR is present in the classpath;
+     * 3. the extra JARs present in the Scala instance are added to the classpath.
+     */
+    public static ClasspathOptions autoNoboot(String scalaVersion) {
+        if (scalaVersion.startsWith("3.") || scalaVersion.startsWith("2.13."))
```
eed3si9n:
```suggestion
        if (!scalaVersion.startsWith("2.") || scalaVersion.startsWith("2.13."))
```

eed3si9n (approved):
Thanks!

PR #1363, opened 2024-05-30
[2.x] Fix the CI

adpi2:
- Bump `utilVersion` to 2.0.0-alpha11
- Ignore scripted tests
- Minor changes to pass the CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
       if: ${{ matrix.jobtype == 2 }}
       shell: bash
       run: |
-        sbt -v -Dfile.encoding=UTF-8 -Dsbt.supershell=never "crossTestBridges" "zincRoot/test" 
+        sbt -v -Dfile.encoding=UTF-8 -Dsbt.supershell=never "Test/compile" "crossTestBridges" "zincRoot/test" 
       # "zincScripted/Test/run"
     - name: Build and test (3)
       if: ${{ matrix.jobtype == 3 }}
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -7,27 +7,30 @@ object Dependencies {
...
-  val allScalaVersions = Seq(scala212, scala210, scala211, scala213)
+  val compilerBridgeVersions = Seq(scala212, scala210, scala211, scala213)
   val scala212_213 = Seq(defaultScalaVersion, scala213)
   val scala3_only = Seq(scala3)
 
-  private val ioVersion = nightlyVersion.getOrElse("1.9.7")
-  private val utilVersion = nightlyVersion.getOrElse("2.0.0-alpha8")
+  private val ioVersion = nightlyVersion.getOrElse("1.10.0")
+  private val utilOrganization = "ch.epfl.scala"
+  private val utilVersion = nightlyVersion.getOrElse("2.0.0-alpha11")
```
adpi2: I published the utils of https://github.com/sbt/sbt/pull/7554 to the `ch.epfl.scala` org and version `2.0.0-alpha11`

eed3si9n: alpha10 is already from the branch  btw.

adpi2: Oh yes you are right. I don't know why I thought we would also need `collections` which is missing on `2.0.0-alpha10`. But we don't.

eed3si9n (approved):

PR #1366, opened 2024-06-19
Use a repeatable hash code for empty files

szeiger:
The current `emptyStampCode` is a system identity hash code that changes between runs in different JVMs or ClassLoaders. This leads to unnecessary invalidations, especially when you use `ConsistentAnalysisFormat` and expect to get identical outputs for identical inputs. Since it is chosen arbitrarily, there is no reason not to hard-code an arbitrary value to get repeatable outputs.

(In our Bazel-based Scala tooling we currently use a copy of `ClasspathCache` with this fix which is wired into the build via `ExternalHooks.Lookup`, which adds a lot of boilerplate for such a small change.)


eed3si9n (approved):
Thanks!

PR #1370, opened 2024-07-03
[2.x] Create empty output jar before compilation

adpi2:
In some Java version (e.g., temurin:11.0.23), the Scala 3 compiler crashes if the jar output file does not exist.


adpi2:
Actually it does not solve the issue, because the compiler always overwrites the jar file with its own and then crashes.

PR #1371, opened 2024-07-03
[2.x] Fix analysis of Java compilation to jar

adpi2:
The goal of this PR is to fix the `java-mixed` test with straight-to-JAR compilation enabled. The difficulty here was to remap class files from the previous JAR (Scala compilation) to class files in the output JAR.

But most of the PR is a refactoring of `JavaAnalyze` to reduce nesting and parameter passing.

This will be needed by https://github.com/sbt/sbt/pull/7592


adpi2:
Actually I have the same or similar issue with `JavaThenScala` so I am going to fix it too.

adpi2:
Superseded by #1376

PR #1373, opened 2024-07-03
Parse column/position information from javac error messages

### LINKED ISSUE #1372: Parse column/position information from javac error messages
IntelliJ IDEA expects compiler error messages to be reported with a `(line, column)` pair. `xsbti.Position` instances produced for error messages coming from `javac` do not contain a value for `pointer`.

https://youtrack.jetbrains.com/issue/SCL-22799/Missing-Line-and-Column-Info-in-Compilation-Error-Message-for-Java-SBT-project

Technically, there is enough information to calculate this, by combining the `line` and `offset` information. But this requires reading the source file from disk, which I'm trying to avoid.

At the moment, both `sbt.internal.inc.javac.DiagnosticsReporter.PositionImpl` and `sbt.internal.inc.javac.JavaPosition` implement `def pointer: Optional[Integer] = o2jo(None)`.

It should be possible to parse this information and include it in the `Position` instances.

vasilmkd:
Resolves #1372 .

I've made a binary incompatible change to `sbt.internal.inc.javac.JavaPosition`. What is the binary compatibility guarantee of this package, given that it has `internal` in the name?

I can re-add the old constructor to preserve binary compatibility. However, even if I add the constructor once more, one of its parameters is no longer uses in the same way as it was before.

Any advice or preference on this matter is appreciated.

### DIFF
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/DiagnosticsReporter.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/DiagnosticsReporter.scala
@@ -72,6 +72,8 @@ object DiagnosticsReporter {
       override val line: Optional[Integer],
       override val lineContent: String,
       override val offset: Optional[Integer],
+      override val pointer: Optional[Integer],
+      override val pointerSpace: Optional[String],
       override val startOffset: Optional[Integer],
       override val endOffset: Optional[Integer],
       override val startLine: Optional[Integer],
@@ -81,8 +83,6 @@ object DiagnosticsReporter {
   ) extends xsbti.Position {
     override val sourcePath: Optional[String] = o2jo(sourceUri)
     override val sourceFile: Optional[File] = o2jo(sourceUri.map(new File(_)))
-    override val pointer: Optional[Integer] = o2jo(Option.empty[Integer])
-    override val pointerSpace: Optional[String] = o2jo(Option.empty[String])
 
     override def toString: String =
       if (sourceUri.isDefined) s"${sourceUri.get}:${if (line.isPresent) line.get else -1}"
@@ -200,6 +200,7 @@ object DiagnosticsReporter {
       def endPosition: Option[Long] = checkNoPos(d.getEndPosition)
 
       val line: Optional[Integer] = o2jo(checkNoPos(d.getLineNumber) map (_.toInt))
+      val pointer: Optional[Integer] = o2jo(checkNoPos(d.getColumnNumber) map (_.toInt))
       val offset: Optional[Integer] = o2jo(checkNoPos(d.getPosition) map (_.toInt))
       val startOffset: Optional[Integer] = o2jo(startPosition map (_.toInt))
       val endOffset: Optional[Integer] = o2jo(endPosition map (_.toInt))
@@ -241,11 +242,20 @@ object DiagnosticsReporter {
           case _ => noPositionInfo
         }
 
+      val pointerSpace = pointer.map[String] { p =>
+        lineContent.toList.take(p.intValue()).map {
+          case '\t' => '\t'
+          case _    => ' '
+        }.mkString
+      }
+
       new PositionImpl(
         sourcePath,
         line,
         lineContent,
         offset,
+        pointer,
+        pointerSpace,
         startOffset,
         endOffset,
         startLine,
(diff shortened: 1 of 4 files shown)


internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/DiagnosticsReporter.scala
```diff
@@ -241,11 +242,20 @@ object DiagnosticsReporter {
           case _ => noPositionInfo
         }
 
+      val pointerSpace = pointer.map[String] { p =>
+        lineContent.toList.take(p.intValue()).map {
+          case '\t' => '\t'
+          case _    => ' '
+        }.mkString
+      }
```
jtjeferreira: this doesn't work?
```suggestion
        lineContent.take(p.intValue()).map {
          case '\t' => '\t'
          case _    => ' '
        }
      }
```

vasilmkd: I followed the code from: https://github.com/sbt/zinc/blob/4eacff2a9bf5c8750bfc5096955065ce67f4e68a/internal/compiler-bridge/src/main/scala/xsbt/DelegatingReporter.scala#L137-L140.

Should we address it in both places?

vasilmkd:
According to https://github.com/sbt/zinc/blob/4eacff2a9bf5c8750bfc5096955065ce67f4e68a/build.sbt#L435

binary incompatible changes are ignored in `sbt.internal.*`, so the change I made should be fine. MiMa certainly doesn't say anything in the CI.

vasilmkd:
I'm working on expanding the tests in `javaErrorParserSpec`.

vasilmkd:
Can someone please definitively confirm that `xsbti.Position#pointer()` is 0-based?

eed3si9n:
> I've made a binary incompatible change to `sbt.internal.inc.javac.JavaPosition`. What is the binary compatibility guarantee of this package, given that it has `internal` in the name?

The binary compatibility is more for the potential Zinc-as-library users, but Zinc devs may need to worry about potential compatibility issues since there may be Scala compiler interaction via compiler bridge. In this case, it should be ok since the interaction is only with javac.

internal/zinc-compile-core/src/test/scala/sbt/internal/inc/javac/JavaCompilerSpec.scala
```diff
@@ -192,20 +195,43 @@ class JavaCompilerSpec extends UnitSpec with Diagrams {
...
+      lineContent: Option[String] = None
+  ): Boolean = {
     def lineContentCheck = lineContent forall (content => p.position.lineContent contains content)
     def lineNumberCheck = p.position.line.isPresent && (p.position.line.get == lineno)
-    lineNumberCheck && lineContentCheck
+    def columnCheck = {
+      val res = p.position.pointer.isPresent && (p.position.pointer.get == colno)
+      if (!res) {
+        println(s"got ${p.position().pointer().get}, expected $colno, problem $p")
+      }
```
eed3si9n, line 210: I am not sure if println out to stdout is a great idea.

eed3si9n:
> Can someone please definitively confirm that `xsbti.Position#pointer()` is 0-based?

I think `xsbti.Position` started out forwarding Scala 2.x compiler, and here's what Scala 2.13.14 prints with `-Xprint-pos`:

```scala
$ scalac --version
Scala compiler version 2.13.14 -- Copyright 2002-2024, LAMP/EPFL and Lightbend, Inc.

$ scalac -Vprint:parser -Xprint-pos A.scala
[[syntax trees at end of                    parser]] // A.scala
[0:63]package [8:15]example {
  [17:63]object Main extends [29:63][37:40]App {
    [29]def <init>() = [29]{
      [NoPosition][NoPosition][NoPosition]super.<init>();
      [29]()
    };
    [45:61][45:52]println([53:60]"hello")
  }
}
```

This looks 0-based. For example, `[8:15]example`:

```
package example
0123456789012345
```

internal/zinc-compile-core/src/test/scala/sbt/internal/inc/javac/JavaCompilerSpec.scala
eed3si9n: Never mind. This is just in the test.

vasilmkd: Ah, I'm so sorry, I used this for debugging the tests and forgot to remove it. I will open a follow up to remove it.

vasilmkd:
> The binary compatibility is more for the potential Zinc-as-library users, but Zinc devs may need to worry about potential compatibility issues since there may be Scala compiler interaction via compiler bridge. In this case, it should be ok since the interaction is only with javac.

Do you want me to try and address the problem?

> I can re-add the old constructor to preserve binary compatibility. However, even if I add the constructor once more, one of its parameters is no longer uses in the same way as it was before.

As I said, I only need some advice on how to handle this specific problem.

eed3si9n:
> Do you want me to try and address the problem?

Since Zinc should be the only code invoking JavaPosition's `apply(...)`, I don't think we need to do anything further?

vasilmkd:
Personally, I was quite happy to see the `sbt.internal.*` MiMa exclusions. 😅

PR #1376, opened 2024-07-08
[2.x] Fix mixed compilation to JAR

adpi2:
For the second compilation step of a mixed compilation (Scala then Java, or vice versa), Zinc copies the current output Jar to a temporary jar. All dependencies on this temporary jar must be declared as internal class dependencies. To do so we we cannot use the declared class file (the temporary jar), instead we use the binary class name, which is stored in a `binaryNameToClassName` map.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -43,7 +43,7 @@ ThisBuild / version := {
   nightlyVersion match {
     case Some(v) => v
     case _ =>
-      if ((ThisBuild / isSnapshot).value) "2.0.0-alpha12-SNAPSHOT"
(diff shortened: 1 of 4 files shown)


eed3si9n:
If possible, could you add a scripted test for this please?

adpi2:
@eed3si9n the scripted tests are disabled in this `wip/sbt-2.x` branch. But I have 2 scripted tests in https://github.com/sbt/sbt/pull/7592 that would fail without this.

I should backport this PR to 1.x and add a scripted test there, but I am not yet sure if it is possible to configure compilation to JAR.

eed3si9n (approved):

PR #795, opened 2020-06-16
Misc scripted/build fixes/improvements

dwijnand:

PR #597, opened 2018-09-17
Enable straight-to-jar compilation

lukaszwawrzyk:
This PR represents my work on implementing straight to jar compilation in zinc (https://github.com/sbt/zinc/issues/305).

To enable the feature, one should specify output as .jar file. This will cause scalac to write files to jar directly. This worked before the PR, but obviously some stuff had to be adjusted for the actual incremental compilation to work.

# Most important things that had to be handled
 - pruning - deleting files from jar instead of from a folder
 - merging - if we have a previous output jar it has to be merged with the one produced in next compilation
 - javac output - javac is not able to compile to jar, I am zipping the output and merging it with scalac output
 - representing jared products - I used sytanx like: `/develop/zinc/target/output.jar!sbt/internal/inc/Compile.class`
 - code that relies on classes being plain files - had to be ifed to work with jars as well

Most of the details related to the feature are in `sbt.internal.inc.STJ` object and something similar in `xsbt.STJ` as it is difficult to share code between those.

# Optimizations
ZipFileSystem was not good enough as it was rewriting jars. What I done is I took that code and kept only the code that allows to manipulate the index. Reading and writing the index to file from ZipFileSystem was most efficient implementation I could find. On top of that I implemented the required operations in performant manner.
 - merging: The jars are concated except for the index. Then the indices are read from both files, merged and written the new index at the end.
 - deleting: Files are only removed from the index.
 - transactional class file manager: The index is stored at the beginning of compilation. In case of a failure, The old index is written back to the jar,
 - reading stamps: The whole index is read once and timestamps are cached.

# Other changes
API of `ReadStamps` was altered - I added `reset()` method to easly allow reseting cache of product timestamps between compilations. This is because I made the stamper stateful to avoid reopening jar for each product (that would kill the performance).

Scripted tests were updated to run for both STJ and regular compilation. Running scripted with STJ requires changing a hardcoded flag in `IncHandler`. I was just using it for development. I am open for discussion on how to do it properly and whether we should e.g. run all scripted tests twice for each build and how to implement that.

Slight addition was to allow disabling compression with a flag when exporting analysis (performance).

Another small addition is to ignore change of `-d` option for javac as it is overriden anyway.

This PR also contains multiple changes related to Windows file system. Mostly about closing things properly to avoid locks (mostly jars). This includes fixing scripted tests on Windows (large percentage of them were failing because of inability to clear the temp dir between tests). There are still a couple of them that are flaky, but I didn't investigate it further.

---

```
[...]

PR #1377, opened 2024-07-09
Fix mixed compilation to jar

adpi2:
This PR is a backport of #1368 and #1376


adpi2:
It seems the scripted tests of compilation to JAR was automated by `bin/run-ci-scripted-to-jar.sh` in https://github.com/sbt/zinc/pull/597 but then it disappeared in https://github.com/sbt/zinc/pull/795, probably because of the migration from Drone to Travis CI.

Enabling that again, it fails with too many errors. I should try to fix those but I'll do it in another PR.

eed3si9n (approved):

PR #1382, opened 2024-08-07
Fix #1140: overcompilation when using a class directory as a library

### LINKED ISSUE #1140: This line seems to contain a bug: if classpathEntry turns out to be a directory, this comparison unsurprisingly always fails:
## original description

        This line seems to contain a bug: if classpathEntry turns out to be a directory, this comparison unsurprisingly always fails:
```
Invalidating '.../build/classes/scala/main/.../Foo.class' because
 .../build/classes/scala/main/....class (farm(77c283a23fed88f2)) != .../build/classes/scala/main (absent)
                                                                         ^^^ RHS is a ***directory***.
```
The `resolve` step was completely dropped in this PR.

_Originally posted by @mernst-github in https://github.com/sbt/zinc/pull/712#discussion_r1002970496_

adpi2:
Fix https://github.com/sbt/zinc/issues/1140


eed3si9n (approved):
Thanks!

PR #1386, opened 2024-09-02
Reuse standardFileManager in LocalJavaCompiler

dwijnand:


dwijnand:
Looks like it's not thread safe to share?

PR #1395, opened 2024-09-11
Copy bytes directly instead of using `scala.reflect.io.Streamable`

rochala:
`scala.reflect.io.Streamable` is way slower than copying bytes directly. Depending on buildtool implementations of `VirtualFile` and `FileConverter` the code path may end up running this method which for smaller files (1Kb) is 10x slower and for larger ones (20Kb+) even 20x slower. This can add up for bigger codebases.

### DIFF
--- a/internal/compiler-bridge/src/main/scala/xsbt/AbstractZincFile.scala
+++ b/internal/compiler-bridge/src/main/scala/xsbt/AbstractZincFile.scala
@@ -12,7 +12,6 @@
 package xsbt
 
 import xsbti.{ PathBasedFile, VirtualFile }
-import scala.reflect.io.Streamable
 
 private trait AbstractZincFile extends scala.reflect.io.AbstractFile {
   def underlying: VirtualFile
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

Please note that both Scala 2.13 and Scala 3 have their own compiler bridges now.

rochala:
> Thanks!
> 
> Please note that both Scala 2.13 and Scala 3 have their own compiler bridges now.

Oh, I wasn't aware of this. I can't find Scala 3 bridge sources, do you mind sharing the link to them ?

lrytz:
IIRC, Scala 3 rewrote the compiler bridge in Java: https://github.com/scala/scala3/tree/main/sbt-bridge/src

With what you know / learned by creating this PR, maybe you can find out if a similar change is needed in Scala 3 as well.

rochala:
Scala 3 is not using reflect.io API at all, so it should not be affected

lrytz:
👍  thanks. Are you going to submit a forward-port to 2.13? https://github.com/scala/scala/blob/2.13.x/src/sbt-bridge/scala/tools/xsbt/AbstractZincFile.scala

Issue #1396, opened 2024-09-16
A lot of code compiled on a small change (local variable name change)


OndrejSpanel:
## steps

I am afraid I need help to reproduce the issue in a smaller project. My project is SBT based, Scala 3, several modules, JS + JVM, with about 500 Scala source files.

## problem

When I make a small change in my project, like renaming a local variable in a function, a lot of files (>400) is compiled.

## expectation

On any change which is internal to a single function and does not affect its interface (parameters or return value) only one file should be compiled.

## notes

It is possible something in my sources is contributing to the problem, but given current tools, it is quite hard to learn what exactly is the issue. The complete build log obtained using `Level.Debug` has over 50 000 lines.

I attach an excerpt obtained using `ThisBuild / logLevel := Level.Debug` from my sbt build which was triggered after I renamed a local variable from `add` to `toAdd` :

```
[debug] compilation cycle 1
[info] compiling 2 Scala sources to C:\Dev\tempi\jvmCommon\target\scala-3.5.0\classes ...

[debug] compilation cycle 2
[info] compiling 36 Scala sources to C:\Dev\tempi\jvmCommon\target\scala-3.5.0\classes ...

[debug] compilation cycle 3
[info] compiling 50 Scala sources to C:\Dev\tempi\jvmCommon\target\scala-3.5.0\classes ...

[debug] compilation cycle 4
[info] compiling 481 Scala sources and 9 Java sources to C:\Dev\tempi\jvmCommon\target\scala-3.5.0\classes ...

```

eed3si9n:
Did you start seeing this behavior with some specific version sbt (if so could you identify the first version it does this)?

Do you use macros, lots of implicits, or package objects? Normally Zinc won't invalidate if the change is limited to a local variable (a variable inside of a function or a method).

Friendseeker:
> The complete build log obtained using `Level.Debug` has over 50 000 lines.

Would you like to share the log? There's still a chance that we can extract something useful out of the log. Ideally also share the entire scala project.

The log hopefully contains reason for invalidation somewhere. This is indeed a peculiar issue as changes such as “after I renamed a local variable from add to toAdd” should not have any impact at all. I still need to double check but I believe Zinc does not grant local variable in a function a name, so none of the name hashes should change.

---

Also does the log always start with

`[debug] compilation cycle 1
[info] compiling 2 Scala sources to C:\Dev\tempi\jvmCommon\target\scala-3.5.0\classes ...`

[...]

PR #1397, opened 2024-09-17
[1.x] Refine early return condition in `invalidateAfterInternalCompilation`

Friendseeker:
### Issue

In https://github.com/sbt/zinc/issues/1396, the project takes 4 cycles to compile when transitive step is set to 3, yet the project only takes 3 cycles when transitive step is above 3.

This indicates transitive invalidation interferes with cycle stopping condition. Even though everything is complied by cycle 3, when `invalidateTransitively` flag is set, we enter cycle 4 regardless.

### Solution

In `invalidateAfterInternalCompilation`, we should use a non-transitive version of `firstClassInvalidation` to compute `newInvalidations`, while using a transitive version of `firstClassInvalidation` to compute all invalidations when `invalidateTransitively` flag is set.

### Testing

A scripted test is added. Without the PR it takes 5 iterations. With the PR it takes 4 iterations.


eed3si9n (approved):
Thanks!

Issue #1399, opened 2024-09-19
Remove the use of XOR hashing in `extraHash` computation


Friendseeker:
https://github.com/sbt/zinc/blob/1899979a26356ccfdc601d5c46992e7ce77020e0/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala#L883-L889

We use `parentsHashes.fold(currentExtraHash)(_ ^ _) ` which is a bad way to combine hashes together. If we use XOR to combine different hashes, identical hashes gets cancelled out. Since we are doing hierarchical hashing here I have a concern that such cancellation would happen.

Consider to replace with `parentsHashes.hashCode` instead.

PR #1405, opened 2024-09-23
[1.x] Use `hashCode` instead of XOR hash in `extraHash` computation

Friendseeker:
### Issue

From #1399

> We use `parentsHashes.fold(currentExtraHash)(_ ^ _) ` which is a bad way to combine hashes together. If we use XOR to combine different hashes, identical hashes gets cancelled out. Since we are doing hierarchical hashing here I have a concern that such cancellation would happen.

Also nowhere else in Zinc codebase do we use XOR to combine hashes, so there is also a consistency issue.

### Fix

Replace XOR with `hashCode`.


Friendseeker:
I guess this is not a trivial change after all... CIs are all failing

Shall investigate further

eed3si9n (approved):

PR #1408, opened 2024-09-24
[1.x] Run `AnalysisFormatBenchmark` during CI

Friendseeker:
c.c. #1406

PR #1406, opened 2024-09-23
[2.x] Run `AnalysisFormatBenchmark` during CI

Friendseeker:
With this PR

- `AnalysisFormatBenchmark` is run automatically during CI similar to other benchmarks.
- Benchmarks are run against PR target branch instead of hardcoded `1.10.x`/`develop` branch

PR #1411, opened 2024-09-27
[2.x] Use hashCode instead of XOR hash in extraHash computation 

Friendseeker:
Forward ports #1405


eed3si9n:
If you don't mind, could you try merging the whole `1.10.x` in? It shouldn't be relatively close - https://github.com/sbt/zinc/compare/develop...1.10.x

Friendseeker:
Yes I can definitely do that. #1408 is a bit problematic though.

Once I figure out https://github.com/sbt/zinc/pull/1406#issuecomment-2367252239 we can proceed with merging the whole `1.10.x` in.

Btw Eugene do you have any intel on what might be causing https://github.com/sbt/zinc/pull/1406#issuecomment-2367252239? As you were recently working with consistent analysis and I saw that you recently worked with the zip files in be7073f3619b7ae8e74e6d50f4b9b939b1cff1c8.

PR #1423, opened 2024-10-03
[2.x] update `scalacOptions` for Scala 3

xuwei-k:
fix warnings

```
[warn] bad option '-Xlint' was ignored
[warn] bad option '-Ywarn-dead-code' was ignored
[warn] bad option '-Ywarn-numeric-widen' was ignored
[warn] bad option '-Ywarn-value-discard' was ignored
```


eed3si9n (approved):
Thanks!

PR #1425, opened 2024-10-03
[2.x] avoid deprecated eta-expand syntax

xuwei-k:
```
Welcome to Scala 3.5.1 (1.8.0_422, Java OpenJDK 64-Bit Server VM).

scala> identity[Int] _
1 warning found
-- Warning: --------------------------------------------------------------------
1 |identity[Int] _
  |^^^^^^^^^^^^^^^
  |The syntax `<function> _` is no longer supported;
  |you can simply leave out the trailing ` _`
val res0: Int => Int = Lambda$1484/1170482099@6233c6c2
```


eed3si9n (approved):
Thanks!

PR #1429, opened 2024-10-05
use `ClassTag` instead of deprecated `Manifest`

xuwei-k:


xuwei-k:
sbt 2 RC released. can't not break binary compatibility

PR #1430, opened 2024-10-05
[2.x] Replace Compilation Timestamp with content hashes

Friendseeker:
This PR replaces compilation timestamp used in `detectAPIChanges` with two types of content hashes. `bytecodeHash` computes content hash of all class files associated with a source file. `extraBytecodeHash` is a hash of all upstream classes's `bytecodeHash`, similar to https://github.com/sbt/zinc/pull/1388#issuecomment-2356974853.

For a non-macro class, we employ `bytecodeHash` as a direct replacement of compilation timestamp.

For a macro class, there are cases for which the macro has same `bytecodeHash` after the macro is recompiled but the macro's upstream dependency is changed hence the macro expansion result would be different. Therefore if a class is macro we also compare by `extraBytecodeHash` to detect such change.


internal/compiler-interface/src/main/contraband/other.json
```diff
@@ -28,6 +28,24 @@
             "Combined with a way to tell if the provenance has changed,",
             "it can be used to short-circuit the 'lookupAnalyzedClass' operation."
           ]
+        },
+        {
+          "name": "bytecodeHash",
+          "type": "int",
```
eed3si9n: We might as well allocate `long` even if we use only the `int` portions initially.

internal/compiler-interface/src/main/contraband/other.json
```diff
@@ -28,6 +28,24 @@
...
+          "default": "0",
+          "since": "2.0.0",
+          "doc": [
+            "A hash of generated bytecode of source file hosting the class"
+          ]
+        },
+        {
+          "name": "extraBytecodeHash",
```
eed3si9n: "extra" often means something supplied by the build user. I suggest `transitiveBytecodeHash` instead.

PR #1432, opened 2024-10-05
[2.x] Handle `DependencyByMacroExpansion` in `ConsistentAnalysisFormat`

Friendseeker:
### Issue

`DependencyByMacroExpansion` is being dropped during analysis serialization / deserialization

### Fix

Handle `DependencyByMacroExpansion` during analysis serialization / deserialization

PR #1434, opened 2024-10-06
[1.x] Handle `DependencyByMacroExpansion` in `ConsistentAnalysisFormat`

Friendseeker:
Backports #1432


eed3si9n (approved):

szeiger:
For future reference: Please make sure to bump the version number when making a breaking change to the format. Trying to read a wrong format will most likely lead to an exception which gets ignored silently, but it makes it harder to perform a gradual migration where you want to be able to read multiple file versions.

eed3si9n:
Good point. Maybe we should add some unit test to pin the version number to sha1 of some files - https://github.com/sbt/zinc/issues/1541

PR #1433, opened 2024-10-06
[2.x] Merge 1.x branch into 2.x

Friendseeker:
Contains changes up to 0787c42

PR #1436, opened 2024-10-06
[2.x] Add 25 minute timeout for CI

Friendseeker:
Ports #1435


Friendseeker:
Shall just include it in #1433 instead.

PR #1438, opened 2024-10-06
[2.x] avoid deprecated collection.mutable.MultiMap

xuwei-k:
https://github.com/scala/scala/blob/v2.13.15/src/library/scala/collection/mutable/MultiMap.scala


eed3si9n (approved):
Danke!

Issue #1442, opened 2024-10-08
CVE-2024-7254 on org.scala-sbt:zinc-persist-core-assembly:1.10.2


gabrieljones:
Sonatype just flagged `org.scala-sbt:zinc-persist-core-assembly:1.10.2` with 

https://nvd.nist.gov/vuln/detail/CVE-2024-7254
https://www.cve.org/CVERecord?id=CVE-2024-7254

Looks like the protobuf lib is shaded into this artifact.

eed3si9n:
@gabrieljones Thanks for letting us know!

Friendseeker:
Thanks for letting us know. Just made a PR to bump protobuf version.

Btw which website did you use to see the sonatype flags? I went on https://ossindex.sonatype.org/component/pkg:maven/org.scala-sbt/zinc-persist_2.12@1.10.2 and saw no flags.

I am thinking if there's some bots we can setup to auto post an issue when some zinc artifact is flagged.

PR #1448, opened 2024-10-09
Submit dependency graph for security alert

adpi2:
Try fixing https://github.com/sbt/zinc/issues/1442#issuecomment-2400768311

### DIFF
--- a/.github/workflows/dependency-graph.yml
+++ b/.github/workflows/dependency-graph.yml
@@ -1,0 +1,16 @@
+# .github/workflows/dependency-graph.yml
+name: Submit Dependency Graph
+on:
+  push:
+    branches: [1.10.x] # default branch of the project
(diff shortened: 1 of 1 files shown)


Friendseeker (approved):
Thanks!

.github/workflows/dependency-graph.yml
```diff
@@ -0,0 +1,16 @@
+# .github/workflows/dependency-graph.yml
+name: Submit Dependency Graph
+on:
+  push:
+    branches: [1.10.x] # default branch of the project
```
Friendseeker, line 5: Btw I am not familiar with how Dependency Graph submission API works, but I guess once the PR is approved, we need to forward port it to develop branch with `branches: [developed]`?

eed3si9n (approved):

PR #1449, opened 2024-10-09
[1.x] Fix CI `sbt: command not found`

Friendseeker:
[Jobs](https://github.com/sbt/zinc/actions/runs/11260318815/job/31311247166) are failing with message

> bin/run-ci.sh: line 5: sbt: command not found

This PR manually installs sbt to address the issue.

### DIFF
--- a/.github/workflows/dependency-graph.yml
+++ b/.github/workflows/dependency-graph.yml
@@ -13,4 +13,5 @@ jobs:
     runs-on: ubuntu-latest # or windows-latest, or macOS-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: sbt/setup-sbt@v1
       - uses: scalacenter/sbt-dependency-submission@v3
--- a/.github/workflows/dependency-graph.yml
+++ b/.github/workflows/dependency-graph.yml
@@ -2,7 +2,7 @@
 name: Submit Dependency Graph
 on:
   push:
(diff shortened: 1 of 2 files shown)


Friendseeker:
Guess Github decided to nuke sbt from their runner images....

----

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

Indeed, they did....

.github/workflows/dependency-graph.yml
```diff
@@ -2,7 +2,7 @@
 name: Submit Dependency Graph
 on:
   push:
-    branches: [1.10.x] # default branch of the project
+    branches: [1.10.x, develop] # default branch of the project
```
jtjeferreira, line 5: why is this needed?

jtjeferreira (approved):

.github/workflows/dependency-graph.yml
Friendseeker: This is not strictly needed but Eugene recommended it when reviewing a similar PR

https://github.com/sbt/sbt/pull/7746#pullrequestreview-2357811045

I guess the motivation is that we can use the same `dependency-graph.yml` across both 1.10.x and develop branch.

SethTisue (approved):

eed3si9n (approved):

PR #1419, opened 2024-10-03
[1.x] Remove the Protocol Buffer usage

Friendseeker:
### Summary

This PR back ports #1388 along with implementing #1412, fixing gaps in consistent analysis format and replacing use of `ParallelGzipOutputStream` with [Snappy](https://github.com/xerial/snappy-java).

### Implementing #1412 (Content Hashing)

This PR replaces timestamp in `detectAPIChanges` with two types of content hashes. `bytecodeHash` computes content hash of all class files associated with a source file. `extraBytecodeHash` is a hash of all upstream classes's `bytecodeHash`, similar to https://github.com/sbt/zinc/pull/1388#issuecomment-2356974853.

For a non-macro class, we employ `bytecodeHash` as a direct replacement of compilation timestamp.

For a macro class, there are cases for which the macro has same `bytecodeHash` after the macro is recompiled but the macro's upstream dependency is changed hence the macro expansion result would be different. Therefore if a class is macro we also compare by `extraBytecodeHash` to detect such change.

### Fixing gaps in consistent analysis format

The following two issues are fixed
- Consistent Analysis Format was not serializing `DependencyByMacroExpansion`. 
- Consistent Analysis Format was not serializing `extraHash`.

### Replacing use of `ParallelGzipOutputStream` with [Snappy](https://github.com/xerial/snappy-java)

There were sporadic CI scripted test timeout failures such as https://github.com/sbt/zinc/actions/runs/11173856057/job/31062605736 and https://github.com/sbt/zinc/actions/runs/11173276915/job/31061098761. These failures commonly happen with (ubuntu-latest, 8, 1) but sometimes happen with (windows-latest, 8, 2). It only happens with Java 8. These failures also cease to happen when we run scripted test one by one.

The CI failures disappear when we use any other compression stream than `ParallelGzipOutputStream`. Hence this PR uses Snappy to compress/decompress Analysis instead. Snappy is slightly faster than `ParallelGzipOutputStream` and is an actively maintained library. c.c. https://github.com/sbt/zinc/pull/1419#issuecomment-2394886677 for raw benchmark result.

PR #1457, opened 2024-10-13
[2.x] Enable scripted test in CI

Friendseeker:
### Issue

When running `scripted` in sbt shell, the following error shows up

```
[error] stack trace is suppressed; run last zincScripted3 / update for the full output
[error] (zincScripted3 / update) sbt.librarymanagement.ResolveException: Error downloading org.scala-sbt:compiler-bridge_3:2.0.0-M2-SNAPSHOT
[error]   Not found
[error]   Not found
[error]   not found: /Users/jiahuitan/.ivy2/local/org.scala-sbt/compiler-bridge_3/2.0.0-M2-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/org/scala-sbt/compiler-bridge_3/2.0.0-M2-SNAPSHOT/compiler-bridge_3-2.0.0-M2-SNAPSHOT.pom
```
### Fix

By removing

```
.configure(
    _.dependsOn(compilerBridge210, compilerBridge211, compilerBridge212, compilerBridge213)
 )
```

sbt would no longer fetch nonexistent `org.scala-sbt/compiler-bridge_3`.

Then by letting `scripted` task depends on `publishBridges`, `ConstantBridgeProvider.fetchCompiledBridge` can then access `target` directory of `compilerBridge210, compilerBridge211, compilerBridge212, compilerBridge213`, which then pick the right directory by calling `ConstantBridgeProvider.bridgeOrBoom` and package the directory with the matching scala version as a jar with naming starting with`scriptedCompilerBridge-bin`, which then gets renamed to `target-bridge-$scalaVersion.jar` in `AbstractBridgeProviderTestkit.getCompilerBridge` and be loaded via `AnalyzingCompiler.getDualLoader`. 

### Comment

I actually don't know if this an actual fix or just a workaround as I still don't know why exactly sbt is fetching nonexistent `org.scala-sbt/compiler-bridge_3`. But nonetheless I feel scripted test is very important and even if this is just a workaround, we should merge it to unblock CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,8 +50,7 @@ jobs:
       if: ${{ matrix.jobtype == 2 }}
       shell: bash
       run: |
-        sbt -v -Dfile.encoding=UTF-8 -Dsbt.supershell=never "Test/compile" "crossTestBridges" "zincRoot/test" 
-      # "zincScripted/Test/run"
+        sbt -v -Dfile.encoding=UTF-8 -Dsbt.supershell=never "Test/compile" "crossTestBridges" "zincRoot/test" "scripted"
     - name: Build and test (3)
       if: ${{ matrix.jobtype == 3 }}
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
 
 jobs:
   test:
+    timeout-minutes: 25
     strategy:
       fail-fast: false
       matrix:
--- a/bin/run-ci.sh
+++ b/bin/run-ci.sh
@@ -11,4 +11,4 @@ sbt -Dfile.encoding=UTF-8 \
   doc \
   crossTestBridges \
   test \
-  # scripted
+  scripted
--- a/build.sbt
+++ b/build.sbt
@@ -653,9 +653,6 @@ lazy val zincScripted = (projectMatrix in internalPath / "zinc-scripted")
     conflictWarning := ConflictWarning.disable,
   )
   .jvmPlatform(scalaVersions = scala3_only)
-  .configure(
-    _.dependsOn(compilerBridge210, compilerBridge211, compilerBridge212, compilerBridge213)
-  )
   .configure(addSbtUtilScripted)
 
 lazy val zincScripted3 = zincScripted.jvm(scala3)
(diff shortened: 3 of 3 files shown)


Friendseeker:
Ah CI is taking forever again...

Actually a good news as CI is catching the [GZIP concurrency issue](https://github.com/sbt/zinc/pull/1419#issuecomment-2394831504).

adpi2:
> I actually don't know if this an actual fix or just a workaround as I still don't know why exactly sbt is fetching nonexistent org.scala-sbt/compiler-bridge_3.

sbt always tries to resolve internal dependencies with Scala binary version. See for instance this:

```scala
ThisBuild / organization := "org.example"
ThisBuild / version := "1.0.0-SNAPSHOT"

lazy val a = project.settings(scalaVersion := "2.13.15")
lazy val b = project.settings(scalaVersion := "3.3.4").dependsOn(a)
```
`b / update` fails with "Error downloading org.example:a_3:1.0.0-SNAPSHOT"

In short, the settings of `zincScripted` looked incorrect and your fix looks good.

adpi2:
@Friendseeker should we merge now or fixing the GZIP concurrency issue first?

Friendseeker:
> @Friendseeker should we merge now or fixing the GZIP concurrency issue first?

I am leaning towards merging the PR now, so the CI scripted test can reject improper fix of GZIP concurrency issue.

@eed3si9n What do you think?

adpi2 (approved):

adpi2:
> I am leaning towards merging the PR now, so the CI scripted test can reject improper fix of GZIP concurrency issue.

I agree, but it's probably not a good idea to freeze a runner for 5 hours.

mzuehlke:
Maybe adding a timeout would help here: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes (default is 360 minutes)

Friendseeker:
> I agree, but it's probably not a good idea to freeze a runner for 5 hours.


> Maybe adding a timeout would help here: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes (default is 360 minutes)

Thanks to both for the feedbacks. Just added a timeout.

mzuehlke (approved):

adpi2 (approved):
Thanks!

PR #1458, opened 2024-10-13
[2.x] Remove compile warnings

Friendseeker:
This PR removes almost all warnings outside of compiler bridge module.


eed3si9n (approved):
We'd be forward/back porting patches between 1.x and 2.x for a while, so we don't want to deviate too far, but I think this is good. Thanks!

PR #1284, opened 2023-11-17
Include extra invalidations in initial validation to fix initial compilation error

Friendseeker:
During initial invalidation, if a class depends on invalidated class, and some invalidated class depends on that class, also invalidate that class.

Fixes https://github.com/sbt/zinc/issues/598.

To validate the fix, minimal reproduction of the issue in https://github.com/sbt/zinc/issues/598#issuecomment-449028234 and https://github.com/sbt/zinc/pull/654 is included as zinc scripted test.

PR #1462, opened 2024-10-16
Revert #1284 Include mutual dependency in initial invalidation

### LINKED ISSUE #1420: Overcompilation on 1.10.x due to potentially incorrect initial invalidation
Follow-up to #1396

## steps

- checkout https://github.com/OndrejSpanel/ZincApiHash
- if using IntelliJ IDE, make sure you have highlighting set to Built-in, not Compiler
- from sbt use `clean;compile`
- change the local variable name in `Entities.f` or change the comment in the same file
- use `compile`

## problem

There are two compilation cycles:

```
[info] compiling 2 Scala sources to C:\Dev\ZincOvercompile\target\scala-3.5.1\classes ...
[info] done compiling
[info] compiling 1 Scala source to C:\Dev\ZincOvercompile\target\scala-3.5.1\classes ...
[info] done compiling
```

## expectation

No more compilation cycles should be needed on a change which does not change the API.

## notes

- I have developed a [SBT plugin](https://gist.github.com/OndrejSpanel/08ff4327c22f05dd0141063b5ef28df1) to help me isolating problems like this. It checks `sbt.internal.inc.Analysis` and prints which files were compiled and what classed have changed their API as a result of compilation. When used on this repro, it shows `apiHash` change on `EntityState` class
- when you repeat modify / `compile` again, there is only one compilation cycle. The additional cycle happens only on the first `compile` after `clean;compile`.

### LINKED ISSUE #1461: Unused import results in unnecessary compilation from 1.10.0-M1 onwards (regression from 1.9.6)
`project/build.properties`
```
sbt.version=1.10.0-M1
```

`build.sbt`
```
scalaVersion := "2.13.15"
```

`src/main/scala/build.scala`
```scala
package build

object outer {
  def foo = { build.subfolder.inner.helperFoo }
}
```

`src/main/scala/subfolder/package.scala`
```scala
package build.subfolder

import build.{outer => unused}

object inner  {
  def helperFoo = { 1 }
}
```

* In SBT/Zinc 1.10.0-M1 onwards, adding a newline to the end of either file results in 2 files being recompiled
* In SBT/Zinc 1.9.6 and before, adding a newline to the end of either file results in only 1 file being recompiled
* Removing the unused `import build.{outer => unused}` results in SBT/Zinc 1.10.0-M1 behaving correctly, only re-compiling 1 file when a newline is added to the end of either of them

It seems to me that Zinc 1.9.6 is correct and Zinc 1.10.0-M1 is wrong. In none of the cases above are any signatures or implementation code changing - I am just adding newlines to the end of the file, so even the line numbers didn't change - and thus we should not need to re-compile any downstream code. But even if we did change implementations (e.g. changing `= { 1 }` to `= { 2 }`) we should not need to recompile downstream code unless signatures change

Noticed in Mill https://github.com/com-lihaoyi/mill/issues/3748

Bisected the Zinc commit log and the misbehavior seems to have been introduced in 8f40c4154e240c3cc3760b79969123eaadda587e. That code seems incorrect to me: 

* We are blindly including mutually-dependent files meaning two files `a <-> b` that depend on one another will be forcibly recompiled even if nothing at all changed (the test case above)

* We are _only_ including _directly_ mutually-dependent files. That means that although a 2-file cycle `a -> b -> a` is considered, a three file cycle `a -> b -> c -> a` is _not_ considered. I have verified this behavior experimentally, and can't imagine any scenarios where this distinction would be something we actually want

### LINKED ISSUE #3748: Investigate why changing the body of a task or method causes downstream files to re-compile
In the `CodeSigSubfolderTests` suite, we can see that modifying the body of task or helper method in the upstream `subfolder/package.mill` causes the downstream `build.mill` to recompile:

https://github.com/com-lihaoyi/mill/blob/5f617f053fbbad32072d69e8967a538bbfbd1e5c/integration/invalidation/codesig-subfolder/src/CodeSigSubfolderTests.scala#L42-L73

This is at odds with how incremental compilation is meant to work, and indeed when testing out simple two-file compilation modules standalone, changing the body of a method in an upstream file without changing the signature does not cause the downstream file to recompile

lihaoyi:
Ref https://github.com/sbt/zinc/pull/1284
Fixes https://github.com/sbt/zinc/issues/1461
Fixes https://github.com/sbt/zinc/issues/1420
Also fixes https://github.com/com-lihaoyi/mill/issues/3748 downstream

Not sure if there's a better way to fix this? Just opening this to spark discussion

The original bugs are fine, but the solution seems incorrect, and is both overly conservative (invalidating everything based on whitespace?) and not conservative enough (doesn't handle cycles with length > 2?). 

I don't know if there's a better answer here, but "invalidate everything in a cycle every time you touch a single file" is problematic, so much so that occasional crashes due to undercompilation seems preferable to universal slowness due to over-compilation. After all, the cost of a compiler crash is just the time spent on a full-clean-recompilation, so the frequency/pain of that needs to be weighed against the time spent re-compiling stuff unnecessarily the other 99% of the time. Maybe we can make it opt-in/out so people can choose the tradeoff that works for themselves?

Now I hate dependency cycles as much as anybody (hence https://github.com/com-lihaoyi/acyclic), but many real-world codebases I have to deal with are not nearly as strict, so I'd like Zinc to remain fast when working with them


Friendseeker:
```
[error] 	source-dependencies/abstract-class-to-trait
[error] 	source-dependencies/false-error
[error] 	source-dependencies/nested-type-params
[error] 	source-dependencies/no-type-annotation
```

@lihaoyi Would you like to rename

```
/zinc/src/sbt-test/source-dependencies/abstract-class-to-trait/test ->/zinc/src/sbt-test/source-dependencies/abstract-class-to-trait/pending
```

(and vice versa for the other 3 tests)?

Since's Mill's next release is [imminent](https://discord.com/channels/632150470000902164/1295786923410984981/1296253344867811338) and Mill build codes heavily feature mutual reference, this is a blocking issue for Mill and I am thinking if we should make a new Zinc release ASAP.

While Mill downgrading Zinc is certainly an option, Zinc fixed a [significant performance regression](https://github.com/sbt/zinc/issues/1311#issuecomment-1868212414) that affected mixed Scala / Java projects so downgrading it would cause a whole different set of slowness problem.

Friendseeker (approved):

eed3si9n (approved):
LGTM. We can always try again on the fix now that we know about the general problem with circular deps.

lihaoyi:
Thanks @eed3si9n ! Could you help me cut a release once this is merged? So I can include it in Mill and close out that issue there

eed3si9n:
Yea. I've been holding off on making a release since we've added a new Java interface and it's good to batch changes. As far as I know we should be able to get something out soon.

lihaoyi:
Thanks @eed3si9n, if you can get it out in the next few days let me know and I'll include it in Mill 0.12.0

eed3si9n:
Yea. I've started the gradual release towards a 1.10.x patch release - https://github.com/sbt/sbt/pull/7776. Hopefully we'll get a new Zinc out today or tomorrow.

eed3si9n:
https://github.com/sbt/zinc/releases/tag/v1.10.3

jackkoenig:
Did anyone ever circle back on this? I ran into related incremental compilation issues 1.10.8 (in a code base with a lot of build units) and have been sticking with 1.10.2 to keep the fix.

PR #1468, opened 2024-10-18
[2.x] Remove `compilerVersionDependentScalacOptions`

### LINKED ISSUE #1467: [2.x] clean up compiler version specific conditions in `build.sbt`
`build.sbt` for develop branch currently contains lots of conditionals based on Scala version. (e.g. `compilerVersionDependentScalacOptions`).

Since we compile Zinc using Scala 3 on develop, we can clean up all these checks and keep them only for `compiler-bridge` module.

We may not want to start working on this right now to avoid 2.x branch diverging too far from 1.x branch.

Friendseeker:
This PR removes `compilerVersionDependentScalacOptions` from `build.sbt` as Zinc 2.x no longer needs to cross build against Scala 2.

As result a bunch more compiler deprecation warnings are surfaced in CI. For example

```
[warn] -- Deprecation Warning: /home/runner/work/zinc/zinc/internal/zinc-classpath/src/main/scala/sbt/internal/inc/ReflectUtilities.scala:38:29 
[warn] 38 |  def fields(clazz: Class[?]) =
[warn]    |                             ^
[warn]    |class OpenHashMap in package scala.collection.mutable is deprecated since 2.13.0: Use HashMap or one of the specialized versions (LongMap, AnyRefMap) instead of OpenHashMap
```
(https://github.com/sbt/zinc/actions/runs/11399506018/job/31718419535)

Is not seen in the latest `develop` [CI run](https://github.com/sbt/zinc/actions/runs/11351881814/job/31573430378).

Closes #1467

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -87,7 +87,6 @@ def baseSettings: Seq[Setting[?]] = Seq(
   testFrameworks += new TestFramework("verify.runner.Framework"),
   compile / javacOptions ++= Seq("-Xlint", "-Xlint:-serial"),
   Test / publishArtifact := false,
-  scalacOptions ++= Seq("-YdisableFlatCpCaching"),
   scalacOptions += {
     scalaBinaryVersion.value match {
       case "2.10" | "2.11" =>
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -103,27 +102,6 @@ def baseSettings: Seq[Setting[?]] = Seq(
...
-          "-Yopt-inline-heuristics:at-inline-annotated"
-        )
-      case _ =>
-        scalacOptions.value.filterNot(
-          Set(
-            "-Xfatal-warnings",
-            "-deprecation",
-            "-YdisableFlatCpCaching",
-          )
-        )
```
eed3si9n, line 122: Don't we have to add this back?

Friendseeker: For `-Xfatal-warnings`, it is never included for Scala 3 projects and Compiler Bridge, hence I was thinking that we don't need to manually filter it out.

https://github.com/sbt/zinc/blob/41dd74e073efc60673cd4d161be1890e5c943685/project/HouseRulesPlugin.scala#L19-L25

https://github.com/sbt/zinc/blob/24012b9c1fdf88196841565cd1477461771140d5/build.sbt#L513-L521

For `-deprecation`, it was suppressing a bunch of valid deprecation warnings, hence we may not want to filter it out.

For `-YdisableFlatCpCaching`, the following logic automatically adds the flag when necessary, leading me to think that we don't need to include it in `build.sbt`

https://github.com/sbt/zinc/blob/41dd74e073efc60673cd4d161be1890e5c943685/zinc/src/main/scala/sbt/internal/inc/IncrementalCompilerImpl.scala#L443-L447

eed3si9n (approved):
Thanks!

build.sbt
eed3si9n: ok. sounds good.

PR #1470, opened 2024-10-18
[2.x] Emit warning for unused

Friendseeker:
Let CI emit warning for unused import, unused variable etc.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -87,6 +87,7 @@ def baseSettings: Seq[Setting[?]] = Seq(
   testFrameworks += new TestFramework("verify.runner.Framework"),
   compile / javacOptions ++= Seq("-Xlint", "-Xlint:-serial"),
   Test / publishArtifact := false,
+  scalacOptions ++= Seq("-Wunused:all"),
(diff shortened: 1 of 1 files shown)


Friendseeker:
Actually this is a good change to have in 1.10.x branch. Maybe do it in 1.10.x first and then forward merge?

In 1.10.x we have `-Xfatal-warnings` on for Scala 2.12 build so may need to do some grunt work.

----

Now I see why "-Ywarn-unused" was not enabled first... It sometimes have false positive
```
[error] D:\Repos\zinc\internal\zinc-classpath\src\main\scala\sbt\internal\inc\ReflectUtilities.scala:52:11: parameter value field in anonymous function is never used
[error]           field <- correspondingFields.get(method.getName)
```

I guess I shall take a break from linting for now. Having two branches & cross building against multiple Scala versions complicate things, and unused is far less important than deprecation warnings.

If anyone else is interested, feel free to takeover the PR.

SethTisue:
Yeah, unused warnings in 2.12 aren't that great. A ton of work (mainly by @som-snytt) went into them later.

PR #1472, opened 2024-10-18
[1.x] Clean compile warnings

Friendseeker:
Backport #1458

Also fixes a bunch more Scala 2.13 deprecation warnings caused by deprecated implicit conversion from Array to Seq. (and with a couple other different deprecation warnings).

```
method copyArrayToImmutableIndexedSeq in class LowPriorityImplicits2 is deprecated (since 2.13.0): implicit conversions from Array to immutable.IndexedSeq are implemented by copying; use `toIndexedSeq` explicitly if you want to copy, or use the more efficient non-copying ArraySeq.unsafeWrapArray
```
Somehow Scala 3 CI build did not complain about them but Scala 2.13 CI build did.


eed3si9n (approved):
Thanks!

PR #1475, opened 2024-10-19
[1.x] avoid deprecated `mutable.OpenHashMap`

Friendseeker:
https://www.scala-lang.org/api/2.13.15/scala/collection/mutable/OpenHashMap.html


eed3si9n (approved):
Thanks!

PR #1479, opened 2024-10-21
[1.x] Make reproducibility toggleable for `ConsistentAnalysisFormat`

Friendseeker:
c.c. https://github.com/sbt/sbt/issues/7790

This PR makes reproducibility of `ConsistentAnalysisFormat` toggle-able via new flag `reproducible` to preserve timestamp information needed for Zinc to conduct certain invalidation decisions. Users who uses reproducible build tool can choose to set `reproducible` flag on (at the expense of less accurate invalidation), and most users who don't use reproducible build tool can set `reproducible` off for more accurate invalidation.

Build tools using Zinc may need to surface the flag to users, and are encouraged to set `reproducible = false` as default option.

### DIFF
--- a/internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/ConsistentAnalysisFormat.scala
+++ b/internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/ConsistentAnalysisFormat.scala
@@ -36,7 +36,7 @@ import Compat._
  * - Faster serialization and deserialization than the existing binary format.
  * - Smaller implementation than either of the existing formats.
  */
-class ConsistentAnalysisFormat(val mappers: ReadWriteMappers, sort: Boolean) {
+class ConsistentAnalysisFormat(val mappers: ReadWriteMappers, reproducible: Boolean) {
   import ConsistentAnalysisFormat._
 
   private[this] final val VERSION = 1100029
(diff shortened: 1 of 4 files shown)


Friendseeker:
Reopening the PR, following discussion in https://github.com/sbt/zinc/pull/1496#issuecomment-2480978181

@ephemerist

internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/ConsistentAnalysisFormat.scala
```diff
@@ -199,14 +204,17 @@ class ConsistentAnalysisFormat(val mappers: ReadWriteMappers, sort: Boolean) {
...
 
   private[this] def writeAPIs(out: Serializer, apis: APIs, storeApis: Boolean): Unit = {
-    def write(n: String, m: Map[String, AnalyzedClass]): Unit =
+    def write(n: String, m: Map[String, AnalyzedClass]): Unit = {
       writeMaybeSortedStringMap(
         out,
         n,
-        m.mapValues(_.withCompilationTimestamp(DefaultCompilationTimestamp))
+        if (reproducible) m.mapValues(_.withCompilationTimestamp(DefaultCompilationTimestamp))
+        else m
```
eed3si9n, line 212: so semantically, this is the only change, and the rest is rename?

ephemerist: @eed3si9n could we have a review of this one please?

eed3si9n (approved):

internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/ConsistentAnalysisFormat.scala
eed3si9n: LGTM.

FYI @szeiger we are gonna break the hermeticity for not sorted version.

PR #1485, opened 2024-10-22
[1.x] Regenerate copyright headers & Add check copyright header consistency job 

Friendseeker:
Some files have missing/incorrect copyright headers. This PR adds all of these headers back along with a CI job to automatically fail when any headers are inconsistent.


eed3si9n (approved):
Thanks!

PR #1487, opened 2024-11-04
[1.x] Surface clear cache method for sbt#7868

Friendseeker:
c.c. https://github.com/sbt/sbt/issues/7432#issuecomment-2453137349

> We should provide some mechanism to wipe out the history for previous compilation

This PR adds the necessary method in Zinc side to wipe in memory previous compilation.

The next step is for sbt side to invoke the new method (probably as a part of `sbt clean`?)

### DIFF
--- a/internal/compiler-interface/src/main/java/xsbti/compile/AnalysisStore.java
+++ b/internal/compiler-interface/src/main/java/xsbti/compile/AnalysisStore.java
@@ -100,6 +100,11 @@ public interface AnalysisStore {
      */
     void set(AnalysisContents analysisContents);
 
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

eed3si9n:
LGTM, but I'll leave this around to see if anyone else wants to chime in.

Friendseeker:
> LGTM, but I'll leave this around to see if anyone else wants to chime in.

Looks like other people are not having concern about the PR judging by a lack of comments in the last 24 days.

Probably good to merge?

PR #1488, opened 2024-11-05
[1.x] Replace deprecated `mutable.AnyRefMap`

Friendseeker:
`mutable.AnyRefMap` is on track top be deprecated in https://github.com/scala/scala/pull/10862. The recommended replacement is `mutable.Hashmap`


eed3si9n (approved):
Thanks!

PR #1497, opened 2024-11-14
[1.x] Print `javac` options in debug log

### LINKED ISSUE #7883: Consider adding the javac call to the sbt output when debug logging is set
Currently, when you enable debug logging via something like
```
set logLevel := Level.Debug
```

you get some nice output on how the _scala compiler_ is invoked. Something like this:
```
[debug] [zinc] The Scala compiler is invoked with:
[debug] 	-deprecation
[debug] 	-encoding
[debug] 	UTF-8
[debug] 	-feature
[debug] 	-language:existentials
[debug] 	-language:higherKinds
[debug] 	-language:implicitConversions
[debug] 	-unchecked
[debug] 	-Xfatal-warnings
[debug] 	-Yno-adapted-args
[debug] 	-Ywarn-dead-code
[debug] 	-Ywarn-numeric-widen
[debug] 	-Ywarn-value-discard
[debug] 	-Ypartial-unification
```

What you don't get is some output on how the _java compiler_ was invoked for projects that also have java files. Adding this might be useful to debug problems in mixed projects.

Friendseeker:
Closes https://github.com/sbt/sbt/issues/7883

### DIFF
--- a/zinc/src/main/scala/sbt/internal/inc/javac/AnalyzingJavaCompiler.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/javac/AnalyzingJavaCompiler.scala
@@ -32,6 +32,7 @@ import xsbti.{
 
 import sbt.util.InterfaceUtil
 import sbt.util.Logger
+import xsbt.Log.debug
 
 /**
  * Define a Java compiler that reports on any discovered source dependencies or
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

zinc/src/main/scala/sbt/internal/inc/javac/AnalyzingJavaCompiler.scala
```diff
@@ -32,6 +32,7 @@ import xsbti.{
 
 import sbt.util.InterfaceUtil
 import sbt.util.Logger
+import xsbt.Log.debug
```
eed3si9n, line 35: I am looking at this again to merge into develop, and I don't know if this is the right logger to use since xsbt package is visible in the compiler bridge implementation only?

Friendseeker: Nice catch! Shall open a new PR that removes the import and calls `log.debug` directly.

Friendseeker: Resolved by https://github.com/sbt/zinc/pull/1516

PR #1501, opened 2024-11-24
[1.x] Let Scala Steward auto-update `com.google.protobuf`

Friendseeker:
Due to popularity of `com.google.protobuf`, CVE reports on `com.google.protobuf` are quite common. Sooner or later Zinc's current version of `com.google.protobuf` will be flagged again.

This needs further discussion but I wonder if we should auto update `com.google.protobuf`.

Previously we decided to opt-out of auto update `com.google.protobuf` due to [worry about regression](https://github.com/sbt/zinc/pull/1236#issuecomment-1677549707), but as we have experienced, sticking with the same `com.google.protobuf` version can also inevitably cause Zinc to be [flagged by security software](https://github.com/sbt/zinc/issues/1442) for depending on a vulnerable dependency.

We also have really strong unit tests for serialization / deserialization that minimizes the risk of regression. For instance our CI caught several issues related to Consistent Analysis Format, including an [intermittent deadlock issue](https://github.com/sbt/zinc/pull/1456).

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -2,9 +2,7 @@
 # either. hopefully this is a reasonable compromise value?
 pullRequests.frequency = "14 days"
 
-updates.ignore = [
-  # as per discussion on sbt/zinc#1236, this is
-  # "if it ain't broke don't fix it" territory
-  { groupId = "com.google.protobuf" },
-  { groupId = "org.eclipse.jgit" }
+updates.pin = [
+  { groupId = "com.google.protobuf", artifactId = "protobuf-java", version="3." },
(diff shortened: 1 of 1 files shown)


.scala-steward.conf
```diff
@@ -5,6 +5,5 @@ pullRequests.frequency = "14 days"
 updates.ignore = [
   # as per discussion on sbt/zinc#1236, this is
   # "if it ain't broke don't fix it" territory
-  { groupId = "com.google.protobuf" },
```
eed3si9n, line 8: If you do this, could you pin protobuf-java to 3.x since the latest is 4.x plz, which is not bincompat?
https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java/

eed3si9n (approved):

PR #1505, opened 2024-11-25
[1.x] Use inclusive language on internal terminology

Friendseeker:
c.c. https://github.com/scala/scala3/issues/21988

### DIFF
--- a/internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala
+++ b/internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala
@@ -266,12 +266,12 @@ object ScalaInstance {
 
   /** Return all the required Scala jars from a path `scalaHome`. */
   def allJars(scalaHome: File): Seq[File] =
-    IO.listFiles(scalaLib(scalaHome)).toIndexedSeq.filter(f => !blacklist(f.getName))
+    IO.listFiles(scalaLib(scalaHome)).toIndexedSeq.filter(f => !excludeList(f.getName))
 
   private[this] def scalaLib(scalaHome: File): File =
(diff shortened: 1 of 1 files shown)


internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala
```diff
@@ -266,12 +266,12 @@ object ScalaInstance {
...
   /** Return all the required Scala jars from a path `scalaHome`. */
   def allJars(scalaHome: File): Seq[File] =
-    IO.listFiles(scalaLib(scalaHome)).toIndexedSeq.filter(f => !blacklist(f.getName))
+    IO.listFiles(scalaLib(scalaHome)).toIndexedSeq.filter(f => !excludeList(f.getName))
 
   private[this] def scalaLib(scalaHome: File): File =
     new File(scalaHome, "lib")
 
-  private[this] val blacklist: Set[String] = Set(
+  private[this] val excludeList: Set[String] = Set(
```
Friendseeker, line 274: Btw I actually don't understand what does the `excludeList` do, but I guess there's probably a better way compared to special casing a bunch of jars...

If someone can decipher what's going on, might be able to remove `excludeList` altogether.

eed3si9n: Likely this code is related to sbt and Zinc's use of layered classloader, where we construct ClassLoader instance using Scala distribution of a specific version (scala-library.jar etc) for performance. In this case, we're trying to exclude non-essential JARs that were historically distributed and/or placed in scalaHome (for scala compiler developers this could be someone on disk)?

eed3si9n (approved):

PR #1508, opened 2024-12-03
Precompile a regex in hot code

retronym:


eed3si9n (approved):
Thanks!

Friendseeker (approved):
Thanks!

I was thinking about the cost of regex overhead in https://github.com/sbt/zinc/pull/1324#issuecomment-1875837199, but this particular optimization slipped my mind...

som-snytt:
github is telling me 500 to view the diff, but from the title I'll say :tada: retronym. I'm so glad retronym is on our side, as they say in the superhero universe.

PR #1516, opened 2024-12-17
[1.x] Use provided logger instead of `xsbt.Log.debug`

Friendseeker:
Addresses https://github.com/sbt/zinc/pull/1497#discussion_r1886437745


eed3si9n (approved):
Thanks!

PR #1517, opened 2024-12-17
[1.x] Add scripted test for #1507

Friendseeker:
Without #1507

```
[error] x source-dependencies/anon-class-dep
[error] sbt.internal.inc.BatchScriptRunner$PreciseScriptedError: {line 2} Command failed: 'B/checkNumberOfLibraries 1'
[error] Caused by: org.scalatest.exceptions.TestFailedException: 2 did not equal 1 analysis.stamps.libraries.size = 2 (expected 1)
```

With #1507
```
About to run tests:
 * source-dependencies/anon-class-dep

[info] + source-dependencies/anon-class-dep
```


eed3si9n (approved):
Thanks!

PR #1521, opened 2024-12-20
[1.x] Delete `.java-version` to not fix java version to 1.8

Friendseeker:
Ports https://github.com/sbt/sbt/pull/7827


.jvmopts
```diff
@@ -1,2 +0,0 @@
--Xmx2G
```
eed3si9n: This was added likely because without it, new contributors run into OOM.

eed3si9n (approved):

PR #1522, opened 2024-12-20
[1.x] Remove `generateVersionFile` from `ZincBuildUtil`

Friendseeker:
### Issue

Currently `compiler-interface` jar on maven central is not reproducible. For example, https://repo1.maven.org/maven2/org/scala-sbt/compiler-interface/1.10.5/compiler-interface-1.10.5.jar contains a file `incrementalcompiler.version.properties`, which contains a build timestamp.

```
version=1.10.5
timestamp=20241130T035052
```

### Fix

We remove the `incrementalcompiler.version.properties` generation logic from Zinc.

### Note

`sbt` does use `incrementalcompiler.version.properties` in https://github.com/sbt/sbt/blob/1.10.x/zinc-lm-integration/src/main/scala/sbt/internal/inc/ZincComponentManager.scala, but since

- Only `sbt` reads from `incrementalcompiler.version.properties`, all other build tools don't
- Creation of `incrementalcompiler.version.properties` uses analysis timestamp, which will be erased by default in zinc 2.x

We should still proceed with the removal of `incrementalcompiler.version.properties` on zinc side, and also remove the usage of  `incrementalcompiler.version.properties` on `sbt` side.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -456,10 +456,6 @@ lazy val compilerInterface = (projectMatrix in internalPath / "compiler-interfac
     libraryDependencies ++= Seq(scalaLibrary.value % Test),
     libraryDependencies ++= Seq(scalatest % Test),
     exportJars := true,
-    Compile / resourceGenerators += Def.task {
-      val a = (Compile / compile).value
-      generateVersionFile(version.value, resourceManaged.value, streams.value, a)
-    }.taskValue,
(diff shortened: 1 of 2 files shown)


eed3si9n:
This is probably a good idea, but we need to remove the usage side first before breaking sbt?

Issue #417, opened 2017-09-27
Inherited trait definitions do not seem to be handled gracefully by Zinc


djspiewak:
See https://github.com/sbt/sbt/issues/3586 for the original report.

Ok, I don't have a nice and minimal test case for you guys, unfortunately, but I do have a reproducer.  It seems like this issue is broadly related to the cake pattern, which is used heavily in some older parts of the quasar codebase (sadface…).  Changing upstream functions (in inherited traits) will cause some downstream compilation units to explode at runtime with `AbstractMethodError` due to Zinc improperly believing the files in question did not need to be invalidated.

This happens intermittently in local development (my steps below point to a changeset which will reproduce the issue) and 100% deterministically in CI.  The reason it is happening deterministically in CI is because we cache our target directories for PR builds in Travis (as discussed with @jvican) in order to bring down our horrendously-long compilation times.  This worked almost flawlessly with SBT 0.13.x for the last eight months (we only needed to cache flush twice), which is quite impressive!  However, on SBT 1.0, we can't even get a single PR build to work, as the slower matrix builds will get the new caches from the faster builds, and themselves run into mutual-invalidation issues.  This is kind of an edge case, but it does demonstrate how much less reliable current Zinc is from its state in 0.13.

## steps

1. Clone [my fork of quasar](https://github.com/djspiewak/quasar)
2. Check out commit `a6b7a91010faf8d06ef678653db53bafdaceaa5e`
3. Run `sbt test:compile`
4. Check out commit `66c97fd9ed3e36eec1a058e5b9a546c80ab6309b`
5. Run `sbt mimir/test`

## notes

sbt version: 1.0.2

eed3si9n:
I was able to reproduce this on the first trial.

## problem

```
[info] StringLibSpecs
[info] for homogeneous sets, the appropriate string function should
[error]   ! determine length
[error]    java.lang.AbstractMethodError: Fatal execution error, caused by quasar.mimir.StringLibSpecs$.buildNonemptyObjects(Lscala/collection/immutable/Map;Lscala/collection/immutable/Map;I)Lscala/collection/immutable/Map; (SliceTransform.scala:488)
[error] quasar.yggdrasil.table.SliceTransforms$SliceTransform$$anonfun$62$$anonfun$apply$17$$anon$14.<init>(SliceTransform.scala:488)
[...]

PR #1528, opened 2025-01-03
[1.x] Local dependency invalidation improvement

rochala:
Hello,
This is an attempt to fix undercompilation which can happen with local dependencies.

This PR is basically trying to solve 2 separate issues:
1. The invariant that `Relations.memberRef` is a superset of `Relations.inheritance` is no longer valid. https://github.com/sbt/zinc/blob/ea947ab20f4660f4e2acec541683bd05b34bd2eb/internal/zinc-core/src/main/scala/sbt/internal/inc/Relations.scala#L193-L194
https://github.com/sbt/zinc/blob/ea947ab20f4660f4e2acec541683bd05b34bd2eb/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala#L110-L125
Basically, since we don't include local dependencies in member references any more, we won't include local inheritance either, so the assumption is no longer true.
This leads to both local inheritance and local member reference to not invalidate correctly, which can be seen in both added test cases.
My attempt at solving this is to actually include all recompiled classes coming from the compilation unit at https://github.com/rochala/zinc/blob/8e6861fc77750b8acf484234ff07082a9785f940/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala#L170-L176
This allows us to include API changes for local dependencies and solves the first issue.
2. Because we don't include local dependencies in `Relations.memberRef` our logic for https://github.com/sbt/zinc/blob/ea947ab20f4660f4e2acec541683bd05b34bd2eb/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala#L502-L508 is not invalidating whole dependency tree. This can be improved a little by computing transitive dependencies based on `firstClassInvalidations` instead of `initialInvalidations`. This will be a bit more heavy to compute, but the result will be something that was in my intended in the first place. 
  Side note: We've actually identified infinite compilation happening, because it turns out that `firstClassTransitiveInvalidations` was not a subset of `firstClassInvalidations`, the reason why is the same as in point 1. The latter includes local inheritance, whereas the former does not.

After some internal testing, we've gone from 12 cycles to 4 in one problematic module. The reason of such high number of cycles was that we did not include local dependencies (and we had a lot of classes defined in a single source) thus they were not considered recompiled, while in fact they were.

Now the question time:
Q1. I see also another solution to those problems, we could try to meet the `membersRef` requirement and include local dependencies. This would also solve the issue with transitive invalidations. Do you know the reason why we ignored these kinds of dependencies ?
Q2. Right now, transitive invalidation may actually result in a lot more compilation happening (it actually recompiles all sources once again). Maybe a whole different approach should be considered, e.g. invalidate everything after 4th cycle or something.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
@@ -506,8 +506,7 @@ private[inc] abstract class IncrementalCommon(
         val firstClassTransitiveInvalidation =
           IncrementalCommon.transitiveDeps(firstClassInvalidation, log)(dependsOnClass)
         log.debug("Invalidate by brute force:\n\t" + firstClassTransitiveInvalidation)
-        firstClassInvalidation ++ firstClassTransitiveInvalidation ++ secondClassInvalidation ++
-          thirdClassInvalidation ++ recompiledClasses
+        firstClassTransitiveInvalidation ++ secondClassInvalidation ++ thirdClassInvalidation ++ recompiledClasses
       } else {
         firstClassInvalidation ++ secondClassInvalidation ++ thirdClassInvalidation
       }
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
@@ -98,7 +98,6 @@ private[inc] abstract class IncrementalCommon(
         invalidatedSources,
         classfileManager,
         pruned,
-        previous,
         classesToRecompile,
         profiler.registerCycle(
           invalidatedClasses,
@@ -149,11 +148,11 @@ private[inc] abstract class IncrementalCommon(
         invalidatedSources: Set[VirtualFile],
         classFileManager: XClassFileManager,
         pruned: Analysis,
-        override val previousAnalysis: Analysis,
         classesToRecompile: Set[String],
         registerCycle: (Set[String], APIChanges, Set[String], Boolean) => Unit
     ) extends IncrementalCallback(classFileManager) {
       override val isFullCompilation: Boolean = allSources.subsetOf(invalidatedSources)
+      override val previousAnalysis: Analysis = previous
       override val previousAnalysisPruned: Analysis = pruned
 
       override def mergeAndInvalidate(
@@ -165,10 +164,12 @@ private[inc] abstract class IncrementalCommon(
             partialAnalysis.copy(compilations = pruned.compilations ++ partialAnalysis.compilations)
           else pruned ++ partialAnalysis
 
-        // Represents classes detected as changed externally and internally (by a previous cycle)
+        // Represents all classes that were compiled as a result of external and internal invalidation (by a previous cycle)
         // Maps the changed sources by the user to class names we can count as invalidated
         val getClasses = (a: Analysis) => initialChangedSources.flatMap(a.relations.classNames)
-        val recompiledClasses = classesToRecompile ++ getClasses(previous) ++ getClasses(analysis)
+        val recompiledClasses = classesToRecompile ++
+          getClasses(previous) ++ getClasses(analysis) ++
+          invalidatedSources.flatMap(previous.relations.classNames)
 
         val newApiChanges =
           detectAPIChanges(recompiledClasses, previous.apis.internalAPI, analysis.apis.internalAPI)
@@ -499,13 +500,14 @@ private[inc] abstract class IncrementalCommon(
       Set.empty
     } else {
       if (invalidateTransitively) {
-        val firstClassTransitiveInvalidation = includeTransitiveInitialInvalidations(
-          initial,
-          IncrementalCommon.transitiveDeps(initial, log)(dependsOnClass),
-          dependsOnClass
-        )
+        // NOTE: As member reference relations do not include local relations, this invalidation will fully propagate
+        // thus we can't rely solely on `firstClassTransitiveInvalidation`. Better bet is to try to find transitive
+        // dependencies from result of `firstClassInvalidation`
+        val firstClassTransitiveInvalidation =
+          IncrementalCommon.transitiveDeps(firstClassInvalidation, log)(dependsOnClass)
         log.debug("Invalidate by brute force:\n\t" + firstClassTransitiveInvalidation)
-        firstClassTransitiveInvalidation ++ secondClassInvalidation ++ thirdClassInvalidation ++ recompiledClasses
+        firstClassInvalidation ++ firstClassTransitiveInvalidation ++ secondClassInvalidation ++
+          thirdClassInvalidation ++ recompiledClasses
       } else {
         firstClassInvalidation ++ secondClassInvalidation ++ thirdClassInvalidation
       }
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Relations.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Relations.scala
@@ -184,13 +184,14 @@ trait Relations {
   private[inc] def externalDependencies: ExternalDependencies
 
   /**
-   * The class dependency relation between classes introduced by member reference.
+   * The class dependency relation between classes introduced by member reference excluding local references
    *
    * NOTE: All inheritance dependencies are included in this relation because in order to
    * inherit from a member you have to refer to it. If you check documentation of `inheritance`
    * you'll see that there's small oddity related to traits being the first parent of a
    * class/trait that results in additional parents being introduced due to normalization.
-   * This relation properly accounts for that so the invariant that `memberRef` is a superset
+   *
+   * Because `inheritance` includes local relations, `memberRef` is not a superset of `inheritance`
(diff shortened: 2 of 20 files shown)


internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -501,13 +502,14 @@ private[inc] abstract class IncrementalCommon(
...
-          dependsOnClass
-        )
+        // NOTE: As member reference relations do not include local relations, this invalidation will fully propagate
+        // thus we can't rely solely on `firstClassTransitiveInvalidation`. Better bet is to try to find transitive
+        // dependencies from result of `firstClassInvalidation`
+        val firstClassTransitiveInvalidation =
+          IncrementalCommon.transitiveDeps(firstClassInvalidation, log)(dependsOnClass)
         log.debug("Invalidate by brute force:\n\t" + firstClassTransitiveInvalidation)
-        firstClassTransitiveInvalidation ++ secondClassInvalidation ++ thirdClassInvalidation ++ recompiledClasses
+        firstClassInvalidation ++ firstClassTransitiveInvalidation ++ secondClassInvalidation ++
```
rochala: Now that I think of it, it may actually be enough to have 
`firstClassTransitiveInvalidation ++ secondClassInvalidation ++ thirdClassInvalidation ++ recompiledClasses` as `firstClassTransitiveInvalidations` is now computed from `firstClassInvalidation`

eed3si9n:
@rochala Thanks for the contribution!
Would you mind opening a bug retroactively, for the future record while you're memory is fresh plz?

I also want to ping @Friendseeker for his views since he's been contributing to Zinc lately.

eed3si9n:
> Q2. Right now, transitive invalidation may actually result in a lot more compilation happening (it actually recompiles all sources once again). Maybe a whole different approach should be considered, e.g. invalidate everything after 4th cycle or something.

Could you elaborate? Zinc already invalidates everything after hitting certain threshold, namely 50% of invalidated sources, or 3 cycle count https://github.com/sbt/zinc/blob/ea947ab20f4660f4e2acec541683bd05b34bd2eb/internal/compiler-interface/src/main/contraband-java/xsbti/compile/IncOptions.java#L12-L17

Are you suggesting that we should increase the threshold to 4 instead?

> Q1. I see also another solution to those problems, we could try to meet the membersRef requirement and include local dependencies. This would also solve the issue with transitive invalidations. Do you know the reason why we ignored these kinds of dependencies ?

It seems like we had similar issue reported in 2017 ('Inherited trait definitions do not seem to be handled gracefully by Zinc' https://github.com/sbt/zinc/issues/417), and I sent https://github.com/sbt/zinc/pull/424 to start tracking previously untracked (?) inheritance within same source. The PR introduced `allowLocal`:

https://github.com/sbt/zinc/blob/ea947ab20f4660f4e2acec541683bd05b34bd2eb/internal/compiler-bridge/src/main/scala/xsbt/Dependency.scala#L124-L125

I guess my thinking back then was that transitive invalidation of inheritance needs to be strict, but transitive invalidation of member ref doesn't matter, which is the basis of name hashing theory. So if the same-source object `B` and object `C` are recompiled together, I didn't see the need to track that member-ref dependencies?

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -165,10 +164,12 @@ private[inc] abstract class IncrementalCommon(
...
           else pruned ++ partialAnalysis
 
-        // Represents classes detected as changed externally and internally (by a previous cycle)
+        // Represents all classes that were compiled as a result of external and internal invalidation (by a previous cycle)
         // Maps the changed sources by the user to class names we can count as invalidated
         val getClasses = (a: Analysis) => initialChangedSources.flatMap(a.relations.classNames)
-        val recompiledClasses = classesToRecompile ++ getClasses(previous) ++ getClasses(analysis)
+        val recompiledClasses = classesToRecompile ++
+          getClasses(previous) ++ getClasses(analysis) ++
+          invalidatedSources.flatMap(previous.relations.classNames)
```
eed3si9n, line 172: This makes sense.

internal/zinc-core/src/main/scala/sbt/internal/inc/Relations.scala
```diff
@@ -184,13 +184,14 @@ trait Relations {
   private[inc] def externalDependencies: ExternalDependencies
 
   /**
-   * The class dependency relation between classes introduced by member reference.
+   * The class dependency relation between classes introduced by member reference excluding local references
```
eed3si9n: In Zinc, the term _local_ often means within the same subproject, i.e. local-to-Analysis, so "excluding same-source references" might be less ambiguous.

internal/zinc-core/src/main/scala/sbt/internal/inc/Relations.scala
```diff
@@ -184,13 +184,14 @@ trait Relations {
...
    *
    * NOTE: All inheritance dependencies are included in this relation because in order to
    * inherit from a member you have to refer to it. If you check documentation of `inheritance`
    * you'll see that there's small oddity related to traits being the first parent of a
    * class/trait that results in additional parents being introduced due to normalization.
-   * This relation properly accounts for that so the invariant that `memberRef` is a superset
+   *
+   * Because `inheritance` includes local relations, `memberRef` is not a superset of `inheritance`
```
eed3si9n: Same here.

eed3si9n:
> Q1. I see also another solution to those problems, we could try to meet the membersRef requirement and include local dependencies.

Also regarding this point, one complication might be the fact that both Scala 3.x and 2.13.x have decided to split out the compiler bridge to a binary artifact published on Maven Central, as opposed to on-the-fly compiled-from-source. This means some range of the Scala version would always embed this characteristics (or issue, depending on how we look at it). So if possible, fixing at at the Zinc level like in IncrementalCommon would be preferable, all else being equal.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
```diff
@@ -149,11 +148,11 @@ private[inc] abstract class IncrementalCommon(
         invalidatedSources: Set[VirtualFile],
         classFileManager: XClassFileManager,
         pruned: Analysis,
-        override val previousAnalysis: Analysis,
```
eed3si9n, line 152: Is this change relevant to the fix?

Friendseeker:
> @rochala Thanks for the contribution! Would you mind opening a bug retroactively, for the future record while you're memory is fresh plz?
> 
> I also want to ping @Friendseeker for his views since he's been contributing to Zinc lately.

Shall review later today. My new school semester is starting in Jan 6 so things got a little busy on my end.

eed3si9n:
@Friendseeker No worries, if you're busy. I think the TLDR of this PR is that in an attempt to fix same-file inheritance invalidation, I managed to create a logical loop hole in Zinc in 2017 via https://github.com/sbt/zinc/pull/424.

internal/zinc-core/src/main/scala/sbt/internal/inc/IncrementalCommon.scala
rochala: Nah, my thinking was to clean up as this parameter that was not necessary and introduced the confusion by duplication of parameters `previous` in `CycleState` vs `previousAnalysis` here.

rochala:
> Could you elaborate? Zinc already invalidates everything after hitting certain threshold, namely 50% of invalidated sources, or 3 cycle count

It seems like entering a transitive invalidation path may actually extend compilation instead of making it faster. Let me explain:
1. If we can't correctly find transitive dependencies for given class and I think that we can't achieve it without same-source relations stored in analysis.
2. This can go into the following state (with my current changes): 
```scala
// source source1
class A
class B:
  val a: A = ???
// source source2
class C:
  val b: B = ???
class D()
  val c: C = ???
// source source3
class E:
  val d: D = ???
class F()
  val e :e = ???
// source sourceN...
class G:
  val f: F = ???
class H()
  val g: G = ???
```
And now if you want to make change in lets say A, we will need to recompile every source `n` (transitive invalidation also includes all recompiled classes)
If we consider for the sake of this example that we're using only transitive invalidations, this would basically do the following:
```
// source1 invalidates source2       // cycle 1 | 2 classes compiled
// source2 invalidates source3       // cycle 2 | 3 clasess compiled (2 recompiled , 1 new class) 
// sourceX invalidates source(X + 1) // cycle X | X + 1 clasess compiled (X recompiled , 1 new class) 
```
3. Basically we have now recompiled same classes over and over again, thus I find it that there are (maybe not that rare) cases that transitive invalidation may have negative impact on your compilation time. Especially that it triggers after 3rd cycle which usually is not a few sources but a dozen.
4. It seems to me like the initial idea of transitive invalidations was to do it all in a single cycle, but that may not be possible right now

I'm not proposing to remove it, I'm rather suggesting that at the current state, we may be better to change it a bit as it is not working as intended. My only worry is that this change will not be possible without including local references.

eed3si9n (approved):

PR #1577, opened 2025-08-30
update mima setting

xuwei-k:


build.sbt
```diff
@@ -8,34 +8,9 @@ def internalPath = zincRootPath / "internal"
...
-      "1.1.0",
-      "1.1.1",
-      "1.1.2",
-      "1.1.3",
-      "1.2.0",
-      "1.2.1",
-      "1.2.2",
-      "1.3.0",
+    val versions: Set[String] = Set(
+      "2.0.0-M6"
```
eed3si9n, line 12: Given that the compiler bridges are now maintained outside, I think we should check the binary compatibility of the compiler interface, though technically since it's an implementation detail, and I don't promise any compatibility of it.

PR #1593, opened 2025-09-20
avoid `private[this]`. fix Scala 3 warnings

xuwei-k:
https://docs.scala-lang.org/scala3/reference/dropped-features/this-qualifier.html


He-Pin:
@xuwei-k Would you like do the same for pekko :)

eed3si9n (approved):
Thanks!

PR #1605, opened 2025-10-02
aggregate all sub projects

xuwei-k:


build.sbt
```diff
@@ -130,7 +130,15 @@ lazy val aggregated: Seq[ProjectReference] = compilerInterface.projectRefs ++
   zincCore.projectRefs ++
   zincPersist.projectRefs ++
   zincTesting.projectRefs ++
-  zinc.projectRefs
+  compilerBridgeTest.projectRefs ++
+  zinc.projectRefs ++ Seq[ProjectReference](
+    classesDep1,
```
eed3si9n, line 135: Why do we need to aggregate this, `jar1`, or `jar2`?

PR #1606, opened 2025-10-06
Increase protobuf recursion limit to 200 (default is 100)

mrdziuban:
As [discussed on discord](https://discord.com/channels/632150470000902164/922600050989875282/1423706375799836753), I ran into an error that causes a project to always be recompiled after loading a new sbt session:

```
com.google.protobuf.InvalidProtocolBufferException: Protocol message had too many levels of nesting.  May be malicious.  Use setRecursionLimit() to increase the recursion depth limit.
```

The [default recursion limit is 100](https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L43), so I chose to double it to 200.


eed3si9n (approved):

mrdziuban:
Hey @eed3si9n, thanks for taking a look at this! Do you have a sense of when you might merge and release a new version of sbt 1.11.x with it included?

eed3si9n:
Thanks for the ping. I'll try to get something out this weekend.

PR #1621, opened 2025-12-17
[2.x] Skip parsing Javadoc info logs

adpi2:
Parsing of Javadoc output fails because of unknown info logs:

```
Building index for all the packages and classes...
Standard Doclet version 21.0.4+7-LTS
Building tree for all the packages and classes...
Generating /foo/bar/MyClass.html...
```

This PR fixes the issue by excluding those logs from parsing in `JavaErrorParser`.


eed3si9n (approved):
Thanks!

PR #1626, opened 2026-01-13
fix: Use strict matching for scala-library jar detection

MkDev11:
The previous check used contains("scala-library") which incorrectly matched any jar with that substring anywhere in the filename, causing user libraries named like "my-scala-library-foo" to be misclassified as the Scala standard library and filtered from the classpath.

Changed to use exact match (scala-library.jar) or prefix match (scala-library-*) to only match the actual Scala library jars.

Ref sbt/sbt#7511


eed3si9n (approved):
Thanks!

PR #1629, opened 2026-01-14
Avoid calling expensive`.toUri.toString` and other micro-optimizations

### LINKED ISSUE #1628: Significant slowdown in Zinc 2.0.0-M11
Running profiles in Mill, almost all the time during Java compilation is spent sorting arrays in `MappedFileConverters`:

Image

This did not used to be a bottleneck in SBT 1.12.0-M2, and is greatly impacting the performance of Mill since upgrading

lihaoyi:
Fixes https://github.com/sbt/zinc/issues/1628

This PR micro-optimizes the top three hotspots i found in the JProfiler profiles compiling the Netty codebase using Javac in Zinc in Mill

## sortBy 

 `.sortBy(x => x.toUri().toString())`  creates a new URI object and String for every comparison during the sort - that's O(n log n) object allocations, and it's blowing up the profile on the linked issue. `.sorted` instead uses Path's native `Comparable<Path>` implementation which compares paths directly without allocating objects.

Both produce deterministic ordering (which is what's needed for reproducible content hashes). The only semantic difference is the comparison method (lexicographic path comparison vs URI string comparison), but for the purpose of consistent hashing, both work.

## isDirectory

`toVirtualFile` called `Files.isDirectory()` on every file from filtered list. Added `toVirtualFileForRegularFile()` that skips the syscall sys we know it is already a regular file

## Caching contentHash

Changed all our `def contentHash`es to `lazy val`s. Looking at the code it seems to be safe - we re-create the file instances every time, and  when file metadata changes `ClasspathCache` creates a new `VirtualFile`. Furthermore, `contentHashStr` is already a `lazy val`, so it seems the code already assumes the files are re-created when changed and so it should be safe to cache the hashes

Manually tested this via `publishLocal` and manually reproducing the original workload in the linked issue, the performance seems to have gone back to previous levels and the profiler no longer shows `toURI` taking up 34% of CPU time, and the `isDirectory`  and `contentHash` hotspots each taking ~5% of CPU time dropped off the profiles as well. `./mill clean &&  time ./mill __.compile ` in the Netty codebase dropped from ~15s to ~8.5s

### DIFF
--- a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/PlainVirtualFile.scala
+++ b/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/PlainVirtualFile.scala
@@ -18,7 +18,7 @@ import java.nio.file.{ Files, Path, Paths }
 import xsbti.{ BasicVirtualFileRef, FileConverter, PathBasedFile, VirtualFileRef, VirtualFile }
 
 class PlainVirtualFile(path: Path) extends BasicVirtualFileRef(path.toString) with PathBasedFile {
-  override def contentHash: Long = HashUtil.farmHash(path)
+  override lazy val contentHash: Long = HashUtil.farmHash(path)
   override def sizeBytes: Long = Files.size(path)
   override lazy val contentHashStr: String = HashUtil.sha256HashStr(input)
   override def name(): String = path.getFileName.toString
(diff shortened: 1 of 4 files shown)


lihaoyi:
CC @eed3si9n for review, looks like CI is green

eed3si9n (approved):
Thanks!

lihaoyi:
Thanks for the merge @eed3si9n. Could I request a milestone release one of these days? I'd like to include this fix in Mill 1.1.0 which should go out Wed 21 Jan

eed3si9n:
Sounds good.

lihaoyi:
Ping on this @eed3si9n, since Scala 3.8.1 is out and a Zinc release with this PR is the only thing blocking Mill 1.1.0 final.

eed3si9n:
Sorry about that. I'll try to get this out today.

eed3si9n:
https://repo1.maven.org/maven2/org/scala-sbt/zinc_3/2.0.0-M13/ is out

lihaoyi:
Thanks eugene!

PR #1630, opened 2026-01-14
[2.x] fix: Support alternative Scala lib directories for scalaHome

### LINKED ISSUE #7477: `scalaHome` support is bitrotted, requires workaround to use with Scala 2.13
(and lord knows whether it works at all with Scala 3!)

If `scalaHome` didn't already exist, I'm not sure we would add it today, as it's not hard to just `publishLocal` from one's Scala 2 checkout (a method that I expect would also work for Scala 3).

I was able to get it partially working with Scala 2.13 by using a workaround, as follows.

I tried:

```
scalaVersion := "2.13.12"
scalaHome := Some(file("/Users/tisue/scala.213"))
```

(the doc is silent about how to handle `scalaVersion`, but we need to get the binary version right, at least; using e.g. `2.13.13-bin-SNAPSHOT"` seems to work just as well)

but then `console` gives:

```
[error] (scalaInstance) sbt.internal.inc.InvalidScalaInstance: Scala instance doesn't exist or is invalid:  (library jar  /Users/tisue/scala.213/lib/scala-library.jar)
```

there is no such `lib` directory in a Scala 2.13 checkout, but after `dist/mkPack`, `build/pack/lib/scala-library.jar` exists, so I thought maybe `ln -s build/pack/lib` would fix it, and it does get past the first error

and it doesn't make `console` work:

```
java.lang.NoSuchMethodError: 'org.jline.utils.AttributedString org.jline.utils.AttributedString.fromAnsi(java.lang.String, java.util.List, java.lang.String, java.lang.String)'
```

but it does make `compile` and `run` work:

```
% cat S.scala
object Main extends App {
  println(s"hi there! ${scala.util.Properties.versionNumberString}")
}
% sbt run
[info] welcome to sbt 1.9.8 (Eclipse Adoptium Java 17.0.9)
[info] loading project definition from /Users/tisue/tmp/scalaHome/project
[info] loading settings for project scalahome from build.sbt ...
[info] set current project to scalahome (in build file:/Users/tisue/tmp/scalaHome/)
[info] running Main 
hi there! 2.13.13-20240118-035212-951cc9b
```

so, the support isn't as fully bitrotted as I'd feared!

anyway, opening this issue to document the current state, but who knows, someone might take an interest in fixing it, and/or perhaps even adding Scala 3 support

MkDev11:
# Fix scalaHome support for Scala 2.13/3 build directories

Fixes sbt/sbt#7477

## The Problem

When using `scalaHome` with a Scala checkout (e.g., for testing local Scala changes), sbt fails with:

```
[error] (scalaInstance) sbt.internal.inc.InvalidScalaInstance: Scala instance doesn't exist or is invalid: (library jar /path/to/scala/lib/scala-library.jar)
```

This happens because zinc expects Scala jars in `$scalaHome/lib/`, but:
- **Scala 2.13** builds from source put jars in `$scalaHome/build/pack/lib/` after running `dist/mkPack`
- **Scala 3** builds may put jars in `$scalaHome/dist/target/pack/lib/`

## The Fix

Update `ScalaInstance.scalaLib()` to check multiple candidate directories:
1. `lib/` (traditional/installed Scala)
2. `build/pack/lib/` (Scala 2.13 after `dist/mkPack`)
3. `dist/target/pack/lib/` (Scala 3 builds)

Also update `libraryJar()` to support both:
- Scala 2: `scala-library.jar`
- Scala 3: `scala3-library_3-*.jar`

## Testing

With this fix, the following now works without the symlink workaround:

```scala
scalaVersion := "2.13.12"
scalaHome := Some(file("/path/to/scala213-checkout"))
```

After running `dist/mkPack` in the Scala checkout, `sbt compile` and `sbt run` work correctly.

### DIFF
--- a/internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala
+++ b/internal/zinc-classpath/src/main/scala/sbt/internal/inc/ScalaInstance.scala
@@ -268,8 +268,23 @@ object ScalaInstance {
   def allJars(scalaHome: File): Seq[File] =
     IO.listFiles(scalaLib(scalaHome)).toIndexedSeq.filter(f => !excludeList(f.getName))
 
-  private def scalaLib(scalaHome: File): File =
-    new File(scalaHome, "lib")
+  private def scalaLib(scalaHome: File): File = {
+    val candidates = Seq(
+      new File(scalaHome, "lib"),
+      new File(scalaHome, "build/pack/lib"),
+      new File(scalaHome, "dist/target/pack/lib")
+    )
+    candidates.find(d => d.isDirectory && hasScalaLibrary(d))
+      .getOrElse(new File(scalaHome, "lib"))
+  }
+
+  private def hasScalaLibrary(dir: File): Boolean = {
+    val files = Option(dir.listFiles()).getOrElse(Array.empty[File])
+    files.exists { f =>
+      val name = f.getName
+      name == "scala-library.jar" || name.startsWith("scala3-library_3")
+    }
+  }
 
   private val excludeList: Set[String] = Set(
     "scala-actors.jar",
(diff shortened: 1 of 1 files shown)


grnlemon:
@eed3si9n please review the push and let me know the result

eed3si9n (approved):
Thanks!

Issue #1540, opened 2025-03-04
Nondeterministic behavior when running zinc multiple times in different folders


lihaoyi:
## steps

* Check out https://github.com/com-lihaoyi/mill/pull/4642 at commit 6398ba3
* Run the reproducibility integration test via

```
./mill -w 'integration.feature[reproducibility].local.server' mill.integration.ReproducibilityTests.diff`
```

* Extract and decompress the `zinc` analysis files via
```
cp out/integration/feature/reproducibility/local/server/test.dest/sandbox/run-1/out/mill-build/compile.dest/zinc zinc-1.txt.gz && cp out/integration/feature/reproducibility/local/server/test.dest/sandbox/run-2/out/mill-build/compile.dest/zinc zinc-2.txt.gz &&  gunzip zinc-2.txt.gz && gunzip zinc-1.txt.gz

```

## problem

This results in two different files 

<img width="1840" alt="Image" src="https://github.com/user-attachments/assets/7753399e-a668-4946-9e61-7dedc419ff9f" />

Attached

[zinc-1.txt](https://github.com/user-attachments/files/19065149/zinc-1.txt)
[zinc-2.txt](https://github.com/user-attachments/files/19065148/zinc-2.txt)

It seems that this `libraryClassName` key in the zinc config picks a random class from each jar to store in the analysis file, and which class gets picked differs every time

## expectation

I would expect the zinc files to contain the same contents every time. I already have custom `ReadWriteMappers` that successfully normalize all the file paths, but this class-based non-determinism doesn't seem related to any files

## notes

It seems like `libraryClassName` should contain a list of every class available in each library, but somehow it's only containing a single class per library, presumably picked arbitrarily. Not sure if we're configuring zinc incorrectly or something

lihaoyi:
@szeiger did you bump into this when you were working on reproducibility at Databricks?

jjudd:
At Lucid we ran into this in our Scala Bazel rule set that uses Zinc. It's been on our list of things to find and fix the root cause of, but we haven't had time yet. We're currently hacking around it setting `libraryClassName` to an empty `libraryClassName`. 

https://github.com/lucidsoftware/rules_scala/blob/lucid-master/src/main/scala/higherkindness/rules_scala/workers/zinc/compile/FilteredRelations.scala

I imagine that may be problematic for incremental compilation. That said, we currently don't use incremental compilation - we're only using the analyzing part of the analyzing caching compiler.

[...]

PR #1638, opened 2026-02-06
Make `libraryClassName` relation deterministic under concurrency

lihaoyi:
Previously, Zinc could produce nondeterministic analysis output because `binaryClassName` used `put(binary, className)` from concurrent `externalLibraryDependency` callbacks. The final value depended on callback scheduling, so `binaryClassName` stores one representative class per binary JAR in a concurrent map, but representative selection was non-deterministic.

This PR replaces `put` with deterministic merge using `updateWith`, choosing the lexicographically smallest class name for each binary. `libraryClassName` only needs a stable representative class per binary for lookup/stamp checks; selecting a deterministic representative preserves behavior while removing scheduler dependence. This removes one source of nondeterminism in Zinc analysis serialization and downstream build cache hashes, which was blocking attempts at reproducible builds in Mill (e.g. https://github.com/com-lihaoyi/mill/pull/4642)

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -744,7 +744,7 @@ private final class AnalysisCallback(
     // representative class name is picked for each binary avoiding non-deterministic output
     binaryClassName.updateWith(binary) {
       case Some(existing) if className.compareTo(existing) >= 0 => Some(existing)
-      case None                                                 => Some(className)
+      case _                                                    => Some(className)
     }
     add(libraryDeps, source, binary)
   }
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
@@ -740,7 +740,14 @@ private final class AnalysisCallback(
       source: VirtualFileRef,
       context: DependencyContext
   ): Unit = {
-    binaryClassName.put(binary, className)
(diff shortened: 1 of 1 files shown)


internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -740,7 +740,12 @@ private final class AnalysisCallback(
       source: VirtualFileRef,
       context: DependencyContext
   ): Unit = {
-    binaryClassName.put(binary, className)
+    // Break ties via lexicographic ordering on the className, which ensures a stable
+    // representative class name is picked for each binary avoiding non-deterministic output
+    binaryClassName.updateWith(binary) {
+      case Some(existing) if className.compareTo(existing) >= 0 => Some(existing)
+      case None                                                 => Some(className)
```
eed3si9n:
```suggestion
      case _                                                    => Some(className)
```

Per the exhaustiveness warning, maybe we need `_` there?

lihaoyi: Thanks for catching it, I messed up my last minute tidying

eed3si9n (approved):
Thanks!

lihaoyi:
@eed3si9n FYI CI looks green

sideeffffect:
This closes

* https://github.com/sbt/zinc/issues/1540 

right?

lihaoyi:
~@sideeffffect it unblocks it, but more work is necessary to get that PR green and mergeable~

I think it actually does close it

PR #1650, opened 2026-02-27
Avoid old varargs splice syntax in build.sbt

xuwei-k:
https://docs.scala-lang.org/scala3/reference/changed-features/vararg-splices.html


eed3si9n (approved):
Thanks!

PR #1680, opened 2026-04-17
Tighter scoping for virtual path replacement

mrdziuban:
This is a follow up to #1545, prompted by sbt/sbt#9101.

It updates `def convertToPath` in `MixedAnalyzingCompiler.scala` to:

1. Only perform virtual path replacement if `config.converter` is a `MappedFileConverter`
2. Use the same logic as `MappedVirtualFile.toPath`, which ensures that a virtual path is replaced only if it is the prefix of the given `String`


eed3si9n (approved):
Thanks!

PR #1687, opened 2026-05-05
Fix warnings. Set explicit scalaVersion

xuwei-k:
```
[warn] scalaVersion for subproject classesDep1 fell back to a default value 2.12.21; declare it explicitly in build.sbt:
[warn]   scalaVersion := "2.12.21"
[warn] scalaVersion for subproject jar1 fell back to a default value 2.12.21; declare it explicitly in build.sbt:
[warn]   scalaVersion := "2.12.21"
[warn] scalaVersion for subproject jar2 fell back to a default value 2.12.21; declare it explicitly in build.sbt:
[warn]   scalaVersion := "2.12.21"
```

### DIFF
--- a/project/ZincBuildUtil.scala
+++ b/project/ZincBuildUtil.scala
@@ -49,6 +49,7 @@ object ZincBuildUtil {
     ) ++ relaxNon212
 
   def relaxNon212: Seq[Setting[?]] = Seq(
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #1689, opened 2026-05-08
[2.x] Update IndexBasedZipFsOpsSpec

xuwei-k:
https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/zip/ZipError.html

> Deprecated, for removal: This API element is subject to removal in a future version.
ZipError is no longer used and is obsolete.


eed3si9n (approved):
Thanks!

PR #1698, opened 2026-05-14
[2.x] fix: Fix NPE in loadInnerClass

### LINKED ISSUE #1697: ClassToAPI.loadInnerClass NPEs on JDK 21+ when a Java class extends Thread (regression in 2.0.0-M17)
Vibe Investigated

## Summary

After upgrading from Zinc 2.0.0-M16 to 2.0.0-M17, compiling a Java-only
project on JDK 21+ fails with a `NullPointerException` in
`sbt.internal.inc.ClassToAPI.loadInnerClass` whenever a user class's
super-type chain contains `java.lang.Thread` (or any other JDK class on
JDK 21+ whose own super-chain transitively reaches `Thread.Builder`).

## Minimal reproducer

- JDK: 21+ (so `java.lang.Thread.Builder` is present)
- Zinc: 2.0.0-M17

```java
// src/main/java/foo/MyThread.java
package foo;
public class MyThread extends Thread {}
```

Compile with Zinc 2.0.0-M17. Compilation fails with:

```
java.lang.NullPointerException: Cannot invoke "java.lang.ClassLoader.loadClass(String)" because "cl" is null
  sbt.internal.inc.ClassToAPI$.loadInnerClass(ClassToAPI.scala:269)
  sbt.internal.inc.ClassToAPI$.innerClassesFromClassfile$$anonfun$2(ClassToAPI.scala:244)
  sbt.internal.inc.ClassToAPI$.innerClassesFromClassfile(ClassToAPI.scala:243)
  sbt.internal.inc.ClassToAPI$.structure(ClassToAPI.scala:216)
  ...
  sbt.internal.inc.javac.AnalyzingJavaCompiler.compile(AnalyzingJavaCompiler.scala:221)
  sbt.internal.inc.MixedAnalyzingCompiler.compile(MixedAnalyzingCompiler.scala:228)
  sbt.internal.inc.IncrementalCompilerImpl.compile(IncrementalCompilerImpl.scala:137)
```

Pre-M17 (M16/M14) compiles the same source successfully.

## Root cause

In M17, `ClassToAPI.structure` switched from `Class.getDeclaredClasses` /
`Class.getClasses` reflection to classfile-based enumeration in
`innerClassesFromClassfile` (sbt/sbt#117):

```scala
// zinc-apiinfo_3 2.0.0-M17, sbt/internal/inc/ClassToAPI.scala
private def innerClassesFromClassfile(c: Class[?], cf: => ClassFile, cmap: ClassMap): Defs = {
  val cl = c.getClassLoader
  val name = c.getName
  // ...
  for {
    parent   <- allSuperTypes(c).collect { case c: Class[?] => c }
    parentCf  = classFileForClass(parent)
    info     <- parentCf.innerClasses if info.outerClassName == parent.getName && info.isPublic
  } {
    loadInnerClass(cl, info, cmap.log).foreach(inheritedClasses += _)
  }
  // merge then calls toDefinitions(cmap) recursively on inheritedClasses
}

private def loadInnerClass(cl: ClassLoader, info: classfile.InnerClassInfo, log: Logger): Option[Class[?]] =
  try Some(cl.loadClass(info.innerClassName))     // <-- NPE when cl is null
  catch {
    case e: (ClassNotFoundException | NoClassDefFoundError) =>
      log.warn(s"Could not load inner class ${info.innerClassName}: $e")
      None
  }
```

The recursion goes:

1. User class `MyThread` is processed by `ClassToAPI.process`. Its lazy
   `structure` is forced (e.g. by `xsbt.api.APIUtil.isAnnotationDefinition`).
2. `innerClassesFromClassfile(MyThread, …)` walks
   `allSuperTypes(MyThread)` = `[Thread, Object, Runnable]`. For
   `parent = Thread`, the loop reads `Thread.classfile`'s `InnerClasses`
   attribute and finds the public entries `Thread$UncaughtExceptionHandler`,
   `Thread$Builder`, `Thread$State` — these are added to
   `inheritedClasses`.
3. `merge` calls `toDefinitions(cmap)` on each of those classes, which
   schedules lazy `structure` evaluations on them via `cmap.lz`.
4. When the lazy for `Thread.Builder` is forced,
   `innerClassesFromClassfile(c = Thread.Builder, …)` is entered with
   `cl = Thread.Builder.getClassLoader = null` (bootstrap-loaded JDK class).
5. On JDK 21+, `Thread.Builder.classfile` has public `InnerClasses`
   entries for `Thread$Builder$OfVirtual` and `Thread$Builder$OfPlatform`
   (introduced for virtual threads), so the direct-inner-class loop
   iterates and calls `loadInnerClass(cl = null, info, …)`.
6. `cl.loadClass(...)` NPEs. The `catch` only handles
   `ClassNotFoundException` / `NoClassDefFoundError`, so the NPE escapes.

The pre-M17 reflection-based path never dereferenced
`Class.getClassLoader`, so this code path didn't exist.

## Why "JDK 21+"

The bug fires only when, somewhere in the recursive walk, a JDK class is
reached that (a) has `getClassLoader == null` (always true for
bootstrap-loaded JDK classes) **and** (b) has at least one public
`InnerClasses` entry where `outerClassName == its own name`.

`java.lang.Thread.Builder`, added in JDK 21 (JEP 444, virtual threads),
is the first such class on the `Thread` super-chain — it declares
`OfVirtual` and `OfPlatform` as public inner classes. On JDK ≤ 20 the
same source compiles, because `Thread.Builder` either doesn't exist
(≤ 18) or its inner classes aren't public.

The same NPE can be triggered by any user class whose super-chain
reaches some JDK class with that shape; `extends Thread` is just the
smallest example.

## Suggested fix

Either of these resolves the NPE — pick whichever matches the intended
semantics:

**Option 1 — short-circuit when `cl` is null.** Matches pre-M17 behavior
(bootstrap-loaded ancestors contribute no inherited inner classes):

```scala
private def loadInnerClass(cl: ClassLoader, info: classfile.InnerClassInfo, log: Logger): Option[Class[?]] =
  if (cl == null) None
  else
    try Some(cl.loadClass(info.innerClassName))
    catch {
      case e: (ClassNotFoundException | NoClassDefFoundError) =>
        log.warn(s"Could not load inner class ${info.innerClassName}: $e")
        None
    }
```

**Option 2 — fall back to the system class loader for bootstrap-loaded
classes**, so the inherited inner classes are actually included in the
API surface:

```scala
private def loadInnerClass(cl: ClassLoader, info: classfile.InnerClassInfo, log: Logger): Option[Class[?]] = {
  val loader = if (cl == null) ClassLoader.getSystemClassLoader else cl
  try Some(loader.loadClass(info.innerClassName))
  catch {
    case e: (ClassNotFoundException | NoClassDefFoundError) =>
      log.warn(s"Could not load inner class ${info.innerClassName}: $e")
      None
  }
}
```

## Verified

- Reproduced with Zinc `2.0.0-M17` + JDK `25` (Zulu 25.0.2).
- Same source compiles cleanly with Zinc `2.0.0-M16` (which still uses
  reflection in `ClassToAPI.structure`).
- Without `extends Thread`, the same project compiles with M17.

lihaoyi:
Fixes https://github.com/sbt/zinc/issues/1697. Vibe coded, but seems to work. In particular I spot checked the upstream PR that surfaced this issue https://github.com/com-lihaoyi/mill/pull/7103 and it seems to fix the original broken test. Covered by a unit test

### DIFF
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -265,13 +265,17 @@ object ClassToAPI {
       cl: ClassLoader,
       info: classfile.InnerClassInfo,
       log: Logger
-  ): Option[Class[?]] =
-    try Some(cl.loadClass(info.innerClassName))
+  ): Option[Class[?]] = {
+    // Bootstrap-loaded classes (e.g. java.lang.Thread.Builder on JDK 21+)
+    // report a null ClassLoader; fall back to the system class loader.
+    val loader = if (cl == null) ClassLoader.getSystemClassLoader else cl
+    try Some(loader.loadClass(info.innerClassName))
     catch {
       case e: (ClassNotFoundException | NoClassDefFoundError) =>
         log.warn(s"Could not load inner class ${info.innerClassName}: $e")
         None
     }
+  }
 
   /** TODO: over time, ClassToAPI should switch the majority of access to the classfile parser */
   private def classFileForClass(c: Class[?]): ClassFile =
--- a/internal/zinc-apiinfo/src/test/scala/sbt/internal/inc/ClassToAPISpecification.scala
+++ b/internal/zinc-apiinfo/src/test/scala/sbt/internal/inc/ClassToAPISpecification.scala
@@ -55,6 +55,26 @@ class ClassToAPISpecification extends UnitSpec {
     assert(apis.keySet === Set("A", "A.B"))
   }
 
+  // Regression: NPE in loadInnerClass when a JDK ancestor is bootstrap-loaded
+  // and has a public InnerClasses entry (e.g. java.lang.Thread.Builder on JDK 21+).
+  it should "not throw NPE when a parent JDK class is bootstrap-loaded" in {
+    IO.withTemporaryDirectory { temp =>
+      val outDir = new File(temp, "out")
+      outDir.mkdir()
+      val src = new File(temp, "MyThread.java")
+      IO.write(src, "public class MyThread extends Thread {}")
+      compileJava(Seq(src), outDir, Seq.empty)
+
+      Using.resource(new java.net.URLClassLoader(Array(outDir.toURI.toURL))) { classloader =>
+        val myThread = classloader.loadClass("MyThread")
+        val (apis, _, _) = ClassToAPI.process(Seq(myThread))
+        // Force the lazy structure so the inner-class walk runs.
+        apis.foreach(_.structure.declared.toIndexedSeq)
+        assert(apis.map(_.name).toSet.contains("MyThread"))
+      }
+    }
+  }
+
   // sbt/sbt#117
   it should "not throw NoClassDefFoundError when inner class references missing type" in {
     IO.withTemporaryDirectory { temp =>


eed3si9n (approved):
Thanks!

lihaoyi:
Thanks @eed3si9n ! May I request a RC/milestone release with this commit? From my own testing this should be enough to let us upgrade Mill to the latest Zinc head so we can help dogfood it

eed3si9n:
Yea, I've started backport to 2.0.x branch to get this out soon.

eed3si9n:
https://github.com/sbt/zinc/releases/tag/v2.0.0-M18 is on its way to Maven Central

PR #1706, opened 2026-05-22
[2.x] fix: Accept jrt: URLs in ClasspathFilter.onClasspath

BrianHotopp:
## Summary

`ClasspathFilter.onClasspath` only recognized `file:` and `jar:` URLs. Since JDK 9 the JDK platform-module classes (`java.sql.*`, `java.scripting.*`, `java.xml.*`, ...) ship in `jrt:/<module>` URLs, so the filter would reject them.

`sbt console` then throws `SecurityException: Prohibited package name: java.sql` because the parent URLClassLoader falls through to `defineClass` on `jrt:` bytes, which the JVM forbids for `java.*` packages from user classloaders.

This change accepts `jrt:` URLs as implicitly on-classpath. Since sbt 2.x mandates JDK 17, the failure case is hit on every `> console` that touches `java.sql`/`java.scripting`/`java.xml`/etc.

Refs sbt/sbt#4328 (the user-visible issue).

## Test plan

- [x] New unit test `ClasspathFilterTest` exercises `loadClass("java.sql.Timestamp")` through a `ClasspathFilter` with an empty user classpath; passes after the change.
- [x] `zincClasspath/test` green.

The end-to-end scripted test in sbt/sbt will follow once this lands and the zinc dep is bumped.

### DIFF
--- a/internal/zinc-classpath/src/main/scala/sbt/internal/inc/classpath/ClassLoaders.scala
+++ b/internal/zinc-classpath/src/main/scala/sbt/internal/inc/classpath/ClassLoaders.scala
@@ -94,7 +94,9 @@ final class ClasspathFilter(parent: ClassLoader, root: ClassLoader, classpath: S
     onClasspath(codeSource.getLocation)
   }
   private def onClasspath(src: URL): Boolean =
-    (src eq null) || (
+    // `jrt:` URLs identify JDK platform-module classes/resources (java.sql, etc.),
+    // which are always implicitly available to user code.
+    (src eq null) || src.getProtocol == "jrt" || (
       ClasspathUtil.asFile(src).headOption match {
         case Some(f) =>
           classpath(f) || directories.exists(dir => ClasspathUtil.relativize(dir, f).isDefined)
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

patrick-nicodemus:
Thank you so much for this fix. I am a new Scala user and I have been trying to use Spark for exploratory data analysis, I have a small pipeline prototype which is configured using sbt and I have been trying to follow a workflow of launching `sbt Test/console`, creating a spark session, and using this to read data from a database. Spark uses reflection at runtime to generate encoders and decoders to convert between its internal representation of datatypes and JVM datatypes, and this triggers the security exception referenced above:

```
ERROR CodeGenerator: failed to compile: org.codehaus.commons.compiler.InternalCompilerException
Caused by: java.lang.SecurityException: Prohibited package name: java.sql
    at java.base/java.lang.ClassLoader.preDefineClass(ClassLoader.java:900)
```

I was able to work around the problem using this fix! However, Spark is still on Scala 2.13 and so I have had to do some tricky stuff to try and get sbt 2.x to cooperate with my older version of sbt. sbt console seems to ignore `fork := true` and so I was getting an `IllegalAccessError` on sun.nio.ch.DirectBuffer because the REPL JVM doesn't receive `--add-opens` flags. I would say this is a very good candidate for backporting as it really makes exploratory data analysis using Spark significantly easier.

PR #1660, opened 2026-03-20
Fix NoClassDefFoundError during analysis of inner classes

BrianHotopp:
## Summary

Fixes sbt/sbt#117 — the oldest open bug in sbt, filed July 22, 2011.

- **Problem**: `ClassToAPI` used Java reflection (`c.getClasses`, `c.getDeclaredClasses`) to discover inner classes during post-compile analysis. When an inner class references types not on the classpath (e.g., optional dependencies like JBoss VFS in Spring), these reflection calls throw `NoClassDefFoundError`, crashing compilation.
- **Fix**: Parse the classfile's `InnerClasses` attribute directly from bytecode instead of using reflection, and gracefully skip any inner classes that can't be loaded via the classloader.
- **Tests**: Added regression test that compiles a class with an inner class extending a type that's then removed from the classpath, verifying `ClassToAPI.process` succeeds. Also added parser-level tests for `InnerClasses` attribute parsing.

This is a revival of #516 by @eed3si9n, rebased onto the current codebase with the tests that were requested at the time.

## Changes

| File | What |
|------|------|
| `ClassFile.scala` | Add `innerClasses` to trait, `InnerClassInfo` case class, `isInnerClasses` on `AttributeInfo` |
| `Parser.scala` | Parse `InnerClasses` classfile attribute into `Array[InnerClassInfo]` |
| `ClassToAPI.scala` | Replace reflection-based inner class discovery with classfile-based approach + try-catch |
| `ParserSpecification.scala` | Tests for `InnerClasses` attribute parsing |
| `ClassToAPISpecification.scala` | Regression test for sbt/sbt#117 |

## Test plan

- [x] All existing `zincClassfile3/test` tests pass (26/26)
- [x] All existing `zincApiInfo3/test` tests pass (42/42), including existing inner class extraction tests
- [x] New regression test verifies `ClassToAPI.process` doesn't throw when inner class references a missing type
- [x] New parser tests verify correct extraction of `InnerClasses` attribute from JDK classes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1711, opened 2026-06-04
[2.x] fix: Classfile fallback for the remaining ClassToAPI reflection sites

rkophs:
Set as a DRAFT first to see if this is the direction the community wants to go.

## Summary

Extends #1660 to the remaining reflection call sites in `ClassToAPI`. Continues chipping away at sbt/sbt#117 (the oldest open sbt bug, July 2011) where `ClassToAPI`'s reflection on a class triggers `NoClassDefFoundError` / `TypeNotPresentException` when a referenced type lives in an optional dep that isn't on the analysis classpath.

#1660 fixed the inner-class case by reading the `InnerClasses` attribute directly. This PR applies the same pattern to the residual sites — methods, constructors, fields, annotations, and exceptions — by keeping reflection as the primary path and falling back to classfile parsing when reflection throws `LinkageError | TypeNotPresentException | ClassNotFoundException`. The reflection path is **unchanged**, so the happy path has zero risk of behavioral regression.

### Why now

Our org is migrating onto Mill + Zinc; #1660 cleared the inner-class crashes but we still hit `ClassToAPI` aborts on classes whose method return / parameter types reference optional deps (e.g. Spring's `compileOnly` + JBoss VFS pattern), and on methods declaring checked exceptions whose types aren't on the classpath.

## Approach

`structure(c, enclPkg, cmap)` becomes a dispatcher:

```scala
try structureReflective(c, enclPkg, cmap)
catch {
  case e: (LinkageError | TypeNotPresentException | ClassNotFoundException) =>
    cmap.log.warn(s"Reflection failed introspecting \${c.getName} (...); falling back to classfile parser")
    structureFromClassfile(c, enclPkg, cmap)
}
```

The same wrap pattern is applied at `c.getAnnotations` and `c.getTypeParameters` in `toDefinitions0`. The fallback emits a `warn` so this rare path is observable in incremental compilation logs.

## Changes

| File | What |
|------|------|
| `ClassFile.scala` | New `ParsedAnnotation` / `ParsedAnnotationArg` case classes. Trait methods `annotations(attrs)`, `parameterAnnotations(attrs)`, `methodExceptions(attrs)`. `AttributeInfo` predicates for `Exceptions` / `RuntimeVisible+Invisible ParameterAnnotations`. `FieldOrMethodInfo` predicates (`isPrivate`/`isProtected`/`isFinal`/`isAbstract`/`isVarArgs`/`isBridge`/`isSynthetic`/`isConstructor`/`isStaticInit`/`isMain`) and the corresponding `ACC_*` constants. |
| `Parser.scala` | Structured parsers for the three new attribute kinds (JVMS §4.7.5, §4.7.16, §4.7.18) reusing the existing constant-pool helpers. `element_value` serializer covers all 13 kinds including nested annotations and arrays. |
| `ClassToAPI.scala` | `structure()` dispatcher + `structureFromClassfile`. `classAnnotationsSafe` and `classTypeParametersSafe` wrappers at the class level. `cf*` helpers convert `FieldOrMethodInfo` to `api.Def`/`Val`/`Var` directly. Inherited members carry the parent's `(Class, ClassFile)` so attribute byte indices resolve against the correct constant pool (a bug caught mid-refactor — without this, reads either return garbage or trip the `entry.tag == ConstantUTF8` assertion in `toUTF8`). Main-class detection moves to the classfile unconditionally (strictly more reliable; no transitive-dep resolution). |
| `DescriptorParser.scala` (new) | Parses JVM field/method descriptors into `xsbti.api.Type` without loading the referenced classes. Mirrors `ClassToAPI.PrimitiveRefs` (`I` → `scala.Int`, etc.) so fallback-produced types hash-compatibly with the reflection path. |
| `ParserSpecification.scala` | +5 tests: `methodExceptions` against `Thread.sleep`, `FieldOrMethodInfo` predicates against JDK, full `element_value` kind coverage via inline-compiled fixture, parameter annotations indexed by position, bridge/synthetic on erasure bridge. |
| `ClassToAPISpecification.scala` | +7 tests: fallback on missing return / parameter / field / class-annotation types (last pins a documented limitation), inherited-member parent-constant-pool regression, `@throws` annotation emission, warn-on-fallback logging. |

## Deferred (TODOs in source)

These are intentional follow-ups to keep this PR scoped:

1. **`Signature` attribute parser (JVMS §4.7.9)** — `classTypeParametersSafe` currently degrades to an empty type-parameter list when reflection on `c.getTypeParameters` throws. A signature parser would also let the method/field fallbacks recover generic return types, generic parameter types, and generic field types. TODO on `classTypeParametersSafe`.
2. **Two-layer supertype recovery** — `allSuperTypesSafe` returns `Seq.empty` on reflection failure, which drops the `cmap.inherited` edges (incremental invalidation cascade) and the parent types in the instance `Structure` (API hash). Right fix: retry with raw `getSuperclass`/`getInterfaces` (no generics), then if that still throws, use `cf.superClassName` + `cf.interfaceNames` directly. TODO on `allSuperTypesSafe`.
3. **Silently-dropped class annotations** — modern JDKs' `c.getAnnotations()` silently drops unresolvable annotation types rather than throwing, so `classAnnotationsSafe` doesn't get a chance to fall back for that specific case. Pinned as a `(limitation)` test. Fix: switch class annotation reading off reflection entirely. TODO on `classAnnotationsSafe`.

### DIFF
--- a/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
+++ b/internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
@@ -16,7 +16,7 @@ package inc
 import java.lang.reflect.{ Array => _, _ }
 import java.lang.annotation.Annotation
 import annotation.tailrec
-import inc.classfile.ClassFile
+import inc.classfile.{ ClassFile, FieldOrMethodInfo, ParsedAnnotation, Constants => CF }
 import xsbti.api
 import xsbti.api.SafeLazyProxy
 import collection.mutable
@@ -132,11 +132,12 @@ object ClassToAPI {
     val enclPkg = packageName(c)
     val mods = modifiers(c.getModifiers)
     val acc = access(c.getModifiers, enclPkg)
-    val annots = annotations(c.getAnnotations)
+    val annots = classAnnotationsSafe(c, cmap)
     val children = childrenOfSealedClass(c)
     val topLevel = c.getEnclosingClass == null
     val name = classCanonicalName(c)
     val tpe = if (Modifier.isInterface(c.getModifiers)) Trait else ClassDef
+    val tparams = classTypeParametersSafe(c, cmap)
     lazy val (static, instance) = structure(c, enclPkg, cmap)
     val cls = api.ClassLike.of(
       name,
@@ -149,10 +150,10 @@ object ClassToAPI {
       emptyStringArray,
       children.toArray,
       topLevel,
-      typeParameters(typeParameterTypes(c))
+      tparams
     )
     val clsDef =
-      api.ClassLikeDef.of(name, acc, mods, annots, typeParameters(typeParameterTypes(c)), tpe)
+      api.ClassLikeDef.of(name, acc, mods, annots, tparams, tpe)
     val stat = api.ClassLike.of(
       name,
       acc,
@@ -172,23 +173,38 @@ object ClassToAPI {
     cmap.memo(name) = defsEmptyMembers
     cmap.allNonLocalClasses ++= defs
 
-    if (
-      c.getMethods.exists(meth =>
-        meth.getName == "main" &&
-          Modifier.isStatic(meth.getModifiers) &&
-          meth.getParameterTypes.length == 1 &&
-          meth.getParameterTypes.head == classOf[Array[String]] &&
-          meth.getReturnType == java.lang.Void.TYPE
-      )
-    ) {
+    if (classFileForClass(c).methods.exists(_.isMain)) {
       cmap.mainClasses += name
     }
 
     defsEmptyMembers
   }
 
-  /** Returns the (static structure, instance structure, inherited classes) for `c`. */
+  /**
+   * Returns the (static structure, instance structure) for `c`.
+   *
+   * Reflection is the primary path. If the JVM can't resolve a referenced type
+   * (typically a transitive dependency marked optional/provided that isn't on
+   * the analysis classpath), we fall back to parsing the classfile directly.
+   * The classfile fallback produces lower-fidelity output (no generic type
+   * parameters; see DescriptorParser for the erased-type substitution).
+   */
   def structure(
+      c: Class[?],
+      enclPkg: Option[String],
+      cmap: ClassMap
+  ): (api.Structure, api.Structure) =
+    try structureReflective(c, enclPkg, cmap)
+    catch {
+      case e: (LinkageError | TypeNotPresentException | ClassNotFoundException) =>
+        cmap.log.warn(
+          s"Reflection failed introspecting ${c.getName} " +
+            s"(${e.getClass.getSimpleName}: ${e.getMessage}); falling back to classfile parser"
+        )
+        structureFromClassfile(c, enclPkg, cmap)
+    }
+
+  private def structureReflective(
       c: Class[?],
       enclPkg: Option[String],
       cmap: ClassMap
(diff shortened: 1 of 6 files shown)


rkophs:
> CLA check for rkophs failed
> Please sign the Scala CLA to contribute.
> Go to https://contribute.akka.io/contribute/cla/scala to sign the Scala CLA
> Error: Process completed with exit code 1.

✅ Signed

eed3si9n:
Thanks for the contribution!

> Our org is migrating onto Mill + Zinc; https://github.com/sbt/zinc/pull/1660 cleared the inner-class crashes but we still hit ClassToAPI aborts on classes whose method return / parameter types reference optional deps (e.g. Spring's compileOnly + JBoss VFS pattern), and on methods declaring checked exceptions whose types aren't on the classpath.

With this change, were you able to get the build going?

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -172,26 +173,41 @@ object ClassToAPI {
...
       enclPkg: Option[String],
       cmap: ClassMap
+  ): (api.Structure, api.Structure) =
+    try structureReflective(c, enclPkg, cmap)
+    catch {
+      case e: (LinkageError | TypeNotPresentException | ClassNotFoundException) =>
+        cmap.log.warn(
+          s"Reflection failed introspecting ${c.getName} " +
+            s"(${e.getClass.getSimpleName}: ${e.getMessage}); falling back to classfile parser"
+        )
```
eed3si9n, line 203: I don't think we should show warning, if the build user can't take actions to correct it.

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -172,26 +173,41 @@ object ClassToAPI {
...
-    if (
-      c.getMethods.exists(meth =>
-        meth.getName == "main" &&
-          Modifier.isStatic(meth.getModifiers) &&
-          meth.getParameterTypes.length == 1 &&
-          meth.getParameterTypes.head == classOf[Array[String]] &&
-          meth.getReturnType == java.lang.Void.TYPE
-      )
-    ) {
+    if (classFileForClass(c).methods.exists(_.isMain)) {
```
eed3si9n, line 176: Does this introduce `classFileForClass(c)` for all classes, which I think is a happy path?

internal/zinc-apiinfo/src/main/scala/sbt/internal/inc/ClassToAPI.scala
```diff
@@ -229,6 +245,305 @@ object ClassToAPI {
...
+   * primary path.
+   */
+  private def classAnnotationsSafe(c: Class[?], cmap: ClassMap): Array[api.Annotation] =
+    try annotations(c.getAnnotations)
+    catch {
+      case e: (LinkageError | TypeNotPresentException | ClassNotFoundException) =>
+        cmap.log.warn(
+          s"Reflection failed reading annotations on ${c.getName} " +
+            s"(${e.getClass.getSimpleName}: ${e.getMessage}); falling back to classfile parser"
+        )
```
eed3si9n, line 356: Same here about warning that user might not take actions.

eed3si9n (changes requested):
I feel like this PR combines a few changes. Would it possible to chunk it into cohesive pieces, both for ease of reviewing, but also so we can revert if we find issues later down the line?

rkophs:
Closing this out as superseded.

After a closer look, [#1730](https://github.com/sbt/zinc/pull/1730) is approaching the same end goal (avoid reflective class loading entirely). But 1730 does so in a more principled way: build `ClassfileToAPI` to feature parity behind an opt-in `IncOptions.classfileJavaApi` flag, with a differential audit against the reflection path. That's the right shape for the long term and I don't want to add competing or divergent work that complicates the path forward.

I begun to split out [#1734](https://github.com/sbt/zinc/pull/1734) out of this PR before realizing the overlap with [#1730](https://github.com/sbt/zinc/pull/1730). To my read, #1734 is structurally orthogonal: it splits the JVMS access-flag constants by context (class / field / method / inner-class) to better follow the JVMS spec.

I'm going to close both out for now, but let me know if you need/want me to resurface 1734 at a later point.

eed3si9n:
Yea, see also 2.0.1 https://github.com/sbt/zinc/releases/tag/v2.0.1 that landed a bunch of Java incremental changes. Once 1730 lands, it might be a good opporunity to see if it can be reorganized etc. Thanks!

PR #1751, opened 2026-07-23
[2.x] Intern directory item conversions in `MappedFileConverter`

### LINKED ISSUE #1750: [2.x] `MappedFileConverter` duplicates directory-entry conversions across consumers, inflating heap on large builds
**Disclaimer**: Investigated with Claude Code

## Summary

`MappedFileConverter.toVirtualFile` expands a *directory* classpath entry into one
`MappedVirtualFile` per contained file. Because every consumer converts its own classpath
independently, a directory that appears on many consumers' classpaths is re-materialized
once per consumer, producing large numbers of value-identical `MappedVirtualFile` instances
(encoded-path `String`s included) held in memory simultaneously. On large multi-project
builds this is a significant and avoidable heap cost.

## Details

When `toVirtualFile` is handed a directory (e.g. a subproject's `classes` output directory on
the compile classpath), `toDirectory` lists the tree and wraps every file in its own
`MappedVirtualFile`
(`internal/zinc-core/src/main/scala/sbt/internal/inc/MappedVirtualFile.scala`):

```scala
def toDirectory(path: Path, encodedPath: String) = {
  val list  = view.list(Glob(path, RecursiveGlob), IsRegularFile && IsNotHidden).map(_._1).sorted
  val items = list.map(toVirtualFileForRegularFile)   // a fresh MappedVirtualFile per file, every call
  MappedDirectory(encodedPath, rootPaths, items.toList)
}
```

So one directory entry containing N files materializes N `MappedVirtualFile` objects — each
carrying its own encoded-path `String` — inside the returned `MappedDirectory`.

In a build tool the `FileConverter` is shared build-wide (e.g. sbt's build-global
`fileConverter`), and each subproject's `CompileOptions` retains its converted classpath.
When M subprojects share a dependency directory on their classpaths, that directory is
converted M times, and the resulting M×N `MappedVirtualFile` instances are value-identical in
groups of M and retained at the same time.

Object graph (M consumers, N files):

```
consumer 1: MappedDirectory#1 → [ItemA#1, ItemB#1, …]   N objects + N strings
consumer 2: MappedDirectory#2 → [ItemA#2, ItemB#2, …]   N objects + N strings
…
consumer M: MappedDirectory#M → [ItemA#M, …]            N objects + N strings
```

`ItemA#1 … ItemA#M` are distinct objects with identical `id`/encoded path — pure duplication.
The duplication is *cross-consumer* (the same path converted by different consumers); items
within a single directory are of course distinct files.

## Impact / magnitude

Illustrative measurement: a 2000-file directory converted by 100 consumers that hold their
conversions simultaneously (a fresh converter per consumer, ~1 KB files) retains on the order
of **~27 MB**, the large majority of which is duplicated `MappedVirtualFile` wrappers and
encoded-path strings that differ only by object identity.

On large monorepos (many subprojects, large shared output/dependency directories) this scales
into hundreds of MB of duplicate `VirtualFile`s held during a build, contributing to GC
pressure and, in the worst case, heap exhaustion.

## Steps to reproduce

1. A multi-project build where many subprojects depend on the same directory-based classpath
   entries (e.g. shared `classes` output directories).
2. Trigger a build that keeps multiple subprojects' `CompileOptions` alive at once (parallel
   compilation).
3. Take a heap dump and count `sbt.internal.inc.MappedVirtualFile` instances grouped by `id` —
   the same `id` appears once per consuming subproject.

## Environment

- zinc `2.0.0-M17-SNAPSHOT` (`develop`)
- `sbt.internal.inc.MappedFileConverter` / `MappedDirectory`

hoangmaihuy:

Fixes #1750

## Problem

`MappedFileConverter.toVirtualFile` expands a directory classpath entry into one
`MappedVirtualFile` per contained file. Because every consumer converts its classpath
independently and the `FileConverter` is shared build-wide, a directory that appears on many
consumers' classpaths is re-materialized once per consumer — producing large numbers of
value-identical `MappedVirtualFile` instances (encoded-path strings included) retained
simultaneously. See #1750 for the full write-up.

## Change

- Cache per-item conversions in `MappedFileConverter`, keyed by the file's
  `(lastModified, size)`, mirroring the metadata-keyed caching already used by
  `ClasspathCache`. Consumers sharing a classpath now share one `MappedVirtualFile` per
  contained file instead of each re-materializing a copy.
- **Directory conversions are still rebuilt from a fresh listing on every call** — only the
  *items* are shared. Directory membership is therefore always current (added/removed files
  are reflected), and each item is revalidated against its `(lastModified, size)` before reuse.
- Items are held through a `SoftReference`: while any live conversion references an item the
  reference cannot clear (simultaneously-held conversions keep sharing one instance); once no
  conversion holds it, the GC may reclaim it under memory pressure, bounding the retention of a
  long-lived shared converter. A reclaimed item is simply rebuilt on the next conversion.
- **Top-level conversions are intentionally left uncached.** Consumers hold a source
  conversion across edits and rely on its lazy `contentHash` reading the content of first
  access; interning those would fuse instances whose laziness must stay independent and break
  incremental source change detection.

## Why it's safe

- **Interned item hashes are never consumed by change detection.** Source stamps read
  `contentHash` only on top-level source files (fresh instances). Classpath/library change
  detection stamps *by path* (`forHashInRootPaths` → `FarmHash.ofPath`), re-resolving class
  files rather than reading `MappedDirectory.items` — even for directory classpath entries.
  So sharing item instances does not affect incremental correctness.
- A cached item is served only while its `(lastModified, size)` is unchanged, so a memoized
  `contentHash`/`contentHashStr` is never served across a detected content change (the same
  heuristic `ClasspathCache` already relies on).
- The cache is an instance field of the converter (not a global singleton), so its retention
  is bounded by the converter's lifetime, and further bounded under pressure by the soft
  references.

## Results

**Micro-benchmark** (`FileConverterBenchmark`, JMH; 2000-file directory, 100 consumers holding
conversions simultaneously): ~12% faster average conversion and ~4.5× less retained heap
(26.8 MB → 5.9 MB) for a shared converter vs a fresh converter per consumer.

**Macro** (real clean compile — fresh empty action cache, wiped target + boot, ~300-subproject
build, 132,362 class files):

| Build                    | wall time     | in-JVM peak | jstat peak | post-GC  | vs stock (peak)   |
|--------------------------|---------------|-------------|------------|----------|-------------------|
| Stock (no interning)     | 968 s         | 17,984 MB   | 16.41 GB   | 3,795 MB | —                 |
| Directory-item interning | 927 s (−4.2%) | 16,944 MB   | 15.48 GB   | 3,884 MB | −1,040 MB (−5.8%) |

`jstat` sample distribution corroborates the ordering (the interning build never exceeded
16 GB; stock touched 16.4 GB).

*Caveat:* one run per build; wall time and momentary heap peaks are both timing-noisy (treat
exact figures as ±a few percent), but the direction — lower peak heap and lower wall time — is
consistent across in-JVM peak, jstat peak, the distribution, and wall clock.

## Tests

`MappedFileConverterInterningSpec` covers:

- items shared across repeated directory conversions;
- an item re-converted fresh once its file is rewritten (same size, bumped mtime);
- directory content changes (add) reflected in a fresh conversion;
- a missing directory that later exists;
- top-level conversions keeping fresh-instance semantics.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/MappedVirtualFile.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/MappedVirtualFile.scala
@@ -13,9 +13,12 @@ package sbt
 package internal
 package inc
 
-import java.io.{ ByteArrayInputStream, InputStream }
+import java.io.{ ByteArrayInputStream, IOException, InputStream }
+import java.lang.ref.SoftReference
 import java.nio.ByteBuffer
 import java.nio.file.{ Files, Path, Paths }
+import java.nio.file.attribute.{ BasicFileAttributes, FileTime }
+import java.util.concurrent.ConcurrentHashMap
 import xsbti.{ BasicVirtualFileRef, FileConverter, PathBasedFile, VirtualFile, VirtualFileRef }
 import sbt.nio.file.{ FileTreeView, Glob, IsNotHidden, IsRegularFile, RecursiveGlob }
 
@@ -50,7 +53,7 @@ object MappedVirtualFile {
 class MappedDirectory(
     encodedPath: String,
     rootPathsMap: Map[String, Path],
-    items: List[VirtualFile]
+    val items: List[VirtualFile]
 ) extends BasicVirtualFileRef(encodedPath)
     with PathBasedFile {
   private def path: Path = MappedVirtualFile.toPath(encodedPath, rootPathsMap)
@@ -141,11 +144,36 @@ class MappedFileConverter(val rootPaths: Map[String, Path], allowMachinePath: Bo
     }
   }
 
+  // Consumers of a shared classpath each convert its entries independently, and a directory entry
+  // wraps every contained file, so without interning a shared classes directory is re-materialized
+  // once per consumer - value-identical MappedVirtualFiles then accumulate on the heap on large
+  // multi-project builds (#1750). Keying by (lastModified, size) keeps a memoized content hash from
+  // being served across a content change, and the SoftReference bounds retention of this long-lived
+  // shared converter. Top-level conversions are deliberately excluded so they stay per-consumer:
+  // their lazy hashes must read content at first access, which incremental source change detection
+  // relies on.
+  private val itemCache =
+    new ConcurrentHashMap[Path, ((FileTime, Long), SoftReference[VirtualFile])]
+
+  private def toItem(path: Path): VirtualFile =
+    try {
+      val attrs = Files.readAttributes(path, classOf[BasicFileAttributes])
+      val metadata = (attrs.lastModifiedTime(), attrs.size())
+      val cached = itemCache.get(path)
+      val hit = if (cached != null && cached._1 == metadata) cached._2.get() else null
+      if (hit != null) hit
+      else {
+        val vf = toVirtualFileForRegularFile(path)
+        itemCache.put(path, (metadata, new SoftReference(vf)))
+        vf
+      }
+    } catch { case _: IOException => toVirtualFileForRegularFile(path) }
+
   def toDirectory(path: Path, encodedPath: String) = {
     val list = view.list(Glob(path, RecursiveGlob), IsRegularFile && IsNotHidden)
       .map(_._1)
       .sorted
-    val items = list.map(toVirtualFileForRegularFile)
+    val items = list.map(toItem)
     MappedDirectory(encodedPath, rootPaths, items.toList)
   }
 }
--- a/internal/zinc-benchmarks/src/test/scala/xsbt/FileConverterBenchmark.scala
+++ b/internal/zinc-benchmarks/src/test/scala/xsbt/FileConverterBenchmark.scala
@@ -1,0 +1,117 @@
+/*
+ * Zinc - The incremental compiler for Scala.
+ * Copyright Scala Center, Lightbend, and Mark Harrah
+ *
+ * Licensed under Apache License 2.0
+ * SPDX-License-Identifier: Apache-2.0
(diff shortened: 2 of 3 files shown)


eed3si9n:
Thanks for the contribution!

internal/zinc-core/src/main/scala/sbt/internal/inc/MappedVirtualFile.scala
```diff
@@ -141,11 +144,45 @@ class MappedFileConverter(val rootPaths: Map[String, Path], allowMachinePath: Bo
...
+  //
+  // The item is held through a SoftReference: while any live conversion still references it the
+  // reference cannot clear, so conversions held simultaneously keep sharing a single instance;
+  // once no conversion holds it, the GC may reclaim it under memory pressure. This bounds the
+  // retention of a long-lived shared converter - a reclaimed item is simply rebuilt on the next
+  // conversion (which re-reads current content, so identity churn is harmless).
+  //
+  // Top-level conversions are deliberately not cached: consumers may hold a conversion across a
+  // file change and rely on its lazy hashes reading the content of first access, so those keep
+  // fresh-instance semantics.
```
eed3si9n: Could we minimize the inline comment plz, esp the ones that essentially transcribes what the code is doing. If we add any comment at all, it would be more useful to add the context / problem description of _why_ this change was introduced, which is something both humans or LLM might forget in the future.

eed3si9n (approved):
Thanks!

PR #1754, opened 2026-07-25
[2.x] Intern analysis values while deserializing

### LINKED ISSUE #1753: Analysis values are duplicated across co-resident analyses
**Disclaimer**: Investigated with Claude Code and human review in the loop

## Summary

In a multi-module build, sbt keeps one `Analysis` resident per subproject for the lifetime of the session. Structurally-equal values are not shared between those analyses, and many are not shared *within* a single analysis either, so a large monorepo pays for millions of redundant objects.

Measured on a 303-module monorepo — 582 persisted analyses, 112 MB gzipped, the complete analysis set of one build — holding every analysis co-resident costs **2233 MB** of heap on `develop`. Roughly a third of that is redundant copies of values that already exist elsewhere in the heap.

That the analyses really are all resident at once is documented on the sbt side in [sbt/sbt#9477](https://github.com/sbt/sbt/issues/9477) ("Memory-retention issues on large monorepo build"): the task engine's `results` map in `Execute.scala` is never pruned, so every task's output — including each subproject's `Analysis` — is retained until the whole evaluation finishes; the `compile` task materializes the full `Analysis` even for no-op builds; and `inMemoryAnalysisCache` (Caffeine, 100 MB cap) weighs entries by *on-disk serialized* size while a deserialized `Analysis` occupies roughly 50× that, so the cap never triggers — that cache alone was measured at 1.5–2.5 GB.

The two problems compound: sbt#9477 governs *how many* analyses are retained and for how long, and this issue is about *how much each retained analysis costs*. Every duplicate below is paid for once per retained copy, and it is what inflates the 50× deserialization multiplier that defeats the cache's weigher.

## Measurements

Per-field walk of the retained object graph over the 582-analysis corpus. `resident` = distinct instances by identity (i.e. what the heap actually holds after the deserializer's own sharing); `Σ distinct` = distinct values by `equals`, summed per analysis; `global` = distinct values across the whole corpus. So `resident − Σ distinct` is redundancy *inside* single analyses, and `Σ distinct − global` is redundancy *across* them.

| bucket | occurrences | resident | Σ distinct | global | within-analysis dup | cross-analysis dup |
|---|---|---|---|---|---|---|
| all strings | 32.5 M | 3.06 M | 3.06 M | 1.01 M | **0** | 2.06 M (~133 MB) |
| `usedNames.UsedName` | 8.22 M | 8.22 M | 1.07 M | 396 K | 7.15 M | 676 K |
| `apis.*.NameHash` | 15.7 M | 15.7 M | 12.6 M | 6.67 M | 3.15 M | 5.90 M |
| `apis.*.Type` | 8.89 M | 8.41 M | 694 K | 259 K | 7.72 M | 435 K |
| `apis.*.Annotation` | 320 K | 320 K | 70 K | 27 K | 250 K | 43 K |

What the numbers say:

1. **Naming is already fully deduplicated within one analysis, and not at all across them.** `withinDup = 0` for strings — the consistent format's per-file string table back-references every repeat to the same instance. But 2.06 M of the 3.06 M resident strings are copies of a string another analysis already holds: ~133 MB of pure cross-analysis redundancy.
2. **`UsedName` is the largest single source of duplication.** 8.22 M instances exist for 396 K distinct values corpus-wide — a 21:1 ratio. Worse, each one carries its own 32-byte `java.util.EnumSet` copy even though only 8 distinct `UseScope` combinations are possible; `RegularEnumSet` alone accounts for 251 MB of the heap.
3. **API tree nodes duplicate overwhelmingly *within* a single analysis** — 7.72 M of the 8.41 M resident `Type` instances are equal to another instance in the same analysis (the same type appears in many signatures). This is redundancy the per-file format does not currently collapse even for a single analysis read in isolation.
4. **`NameHash` has the highest volume but the least favorable shape.** 15.7 M instances, but the internal side has *zero* within-analysis duplication and the distinct population is 6.7 M corpus-wide, so it is the one large type where deduplication is not obviously profitable.

Where the heap actually goes, by class, for the same 582 analyses (`GC.class_histogram`):

| class | instances | bytes |
|---|---|---|
| `xsbti.api.NameHash` (+ arrays) | 15.7 M | 424 MiB |
| `java.util.RegularEnumSet` | 8.22 M | 251 MiB |
| `sbt.internal.inc.UsedName` | 8.22 M | 188 MiB |
| `java.lang.String` + `byte[]` bodies | 3.11 M | 279 MiB |
| `xsbti.api.Projection` + `Singleton` | 8.21 M | 157 MiB |

## Constraint worth noting

Contraband generates identity equality (`this == obj`) for `Structure`, `ClassLike`, `ClassDefinition`, `AnalyzedClass`, `ParameterList`, `MethodParameter` and `AnnotationArgument`, so any equals-keyed deduplication is structurally impossible for those types. The duplication measured above is confined to types that do have value equality, plus strings.

## How this was measured

- Corpus: 582 persisted consistent-binary analyses (format version 1100029) from one monorepo build.
- A per-field walk of the retained graph counting identity-distinct instances, per-analysis value-distinct instances, and corpus-wide value-distinct instances for every field of `Analysis`/`MiniSetup`.
- Heap figures from `GC.class_histogram -all` after loading all analyses co-resident in a 4 GB heap, sampled without an explicit GC (the footprint a build server actually has).
- zinc `develop` (`v2.0.0-M15`+), Scala 3.8.4, JDK 26, macOS/aarch64.

hoangmaihuy:

Fixes https://github.com/sbt/zinc/issues/1753

## Changes

- **Global weak string pool** — interned inline in `BinaryDeserializer.string()`, so the per-read string table and all of its back-references hold the cross-analysis-shared instance.
- **`UsedName` pools** — 8 weak-valued `name → UsedName` pools, one per `UseScope` combination. A `UsedName` is fully determined by its (already canonical) name and 3 scope bits, so probing is keyed by the name string and a pool hit allocates **nothing**; a candidate is constructed only on a miss. `UsedName` also gains a cached `hashCode` (it is re-hashed into a per-class `Set` on every read) which fits in existing object padding.
- **Both pool types are hand-rolled** in `zinc-core/WeakPools.scala` on `ConcurrentHashMap` + `ReferenceQueue`, so this adds **no dependency**. `WeakInterner` keys its map by a weak reference that hashes and compares by its referent's *value*; `WeakValuePool` keeps strong equals-keyed keys with weakly-held values, which is what makes probe-before-construct possible.
- **Per-read node cache** — a plain `HashMap` on `Deserializer`, used through `ConsistentAnalysisFormat.internNode`, dedups value-equality `xsbti.api` nodes within a single read. ~95% of `Type`/`TypeParameter`/`Annotation` duplication is intra-analysis, so this captures nearly all of it with no weak-reference bookkeeping, and it is leak-free by construction: the cache dies with the read. `Structure` and `EmptyType` are excluded (identity equality / singleton).
- **Shared scope `EnumSet`s** — the 8 possible `UseScope` sets are allocated once and shared instead of copied per `UsedName`. Worth 251 MB by itself, independent of interning.
- **Cached `enum.values()`** — `values()` clones its backing array on every call and sits on per-node read paths (`UseScope`, `Severity`, `CompileOrder`, `DefinitionType`, `Variance`, `ParameterModifier`).
- **`readUsedNameSet` interns per name** instead of materializing a `Seq[String]` per scope group and mapping it into fresh `UsedName`s; `writeUsedNameSet` shares the same scope-bit encoding instead of repeating it.
- **The fresh-compilation path interns too** (`AnalysisCallback.usedName`), so compiler-produced names are canonical, not only deserialized ones.

## Leak-freedom

A build server runs for days, so pooling must not retain. Both pools hold their canonical values **weakly**: once the last analysis referencing a value is dropped the value becomes collectable, and every pool operation drains the reference queue so the entry goes with it.

`WeakInterner` (strings) keys the map by a weak reference that hashes and compares by the *value* of its referent, so nothing in the pool strongly references an interned string. `WeakValuePool` (`UsedName`) is keyed strongly by the name and holds the value weakly; the entry remembers its key so that expunging a dead value releases the key too. (Keying that pool weakly instead would leak — the value references its key, pinning it.) `WeakPoolSpec` asserts, for both pools, that values *and* keys are released once unreachable; `AnalysisInternerSpec` asserts the same end to end, plus retention while a reference is held.

Thread safety: the pools are `ConcurrentHashMap`s, and the fresh-compile path interns from many threads at once; a 16-thread test asserts every thread converges on one canonical instance without deadlock.

Transparency: a test asserts api hashes are identical before and after a round trip, so nothing incremental compilation observes changes.

## Why `NameHash` is deliberately not interned

It is the highest-volume type (15.7 M instances, 424 MB incl. arrays) and the obvious candidate, but the shape is wrong: its internal side has **zero** within-analysis duplication (5,999,323 occurrences = 5,999,323 distinct values) and the corpus-wide distinct population is 6.7 M, so a weak pool would need ~6.7 M entries at ~40 B each — more than the duplicates it removes. Measured directly, interning it *increased* retained heap (+12 MB at 100 analyses) and cost 15.6% of read CPU. Unlike `UsedName` there is also no probe-before-construct trick, because identity includes the api hash int, so all 15.7 M reads would have to allocate a candidate first.

It remains the largest analysis-side consumer and is the next thing worth attacking, but it needs a representation change (parallel `String[]`/`int[]` instead of 15.7 M wrappers, or sharing whole `nameHashes` arrays per `(className, apiHash)`) — both breaking changes to a public `xsbti` type, so out of scope here.

## Benchmark results

All numbers below are from one session on one machine, each measurement paired against a stock `develop` worktree running the identical harness back to back.

### Heap: all 582 analyses co-resident

4 GB fixed heap, loading every analysis in the corpus and holding strong references to all of them. Both heap figures come from the same run: first sampled with no explicit GC (what the process actually occupies right after loading, floating garbage included), then after four forced GCs with every analysis still held (the live set — the steady-state floor a long-lived build server sits at).

| metric | stock `develop` | this PR | Δ |
|---|---|---|---|
| used heap after load, **before GC** | 2335 MB | 1738 MB | −597 MB (−26%) |
| **retained heap after GC** | 1906 MB | **1237 MB** | **−669 MB (−35%)** |
| garbage in the pre-GC sample | 429 MB | 501 MB | |
| histogram total bytes (incl. garbage) | 2343 MiB | 1296 MiB | −1047 MiB (−45%) |
| histogram total objects (incl. garbage) | 82.28 M | 43.61 M | −38.7 M (−47%) |

**Judge this by the retained row.** It is stable: stock measured 1906 MB in two sessions a day apart, and a second harness reports 1904 MB (1 thread) / 1901 MB (8 threads) for stock against 1244 / 1238 MB for this PR — agreement to within 0.6%. The pre-GC row is that same live set plus whatever floating garbage the loader happened to leave behind, which depends entirely on when G1 last collected: across repeat runs stock landed between 2233 and 2536 MB and this PR between 1460 and 1738 MB, i.e. a delta anywhere from −26% to −39%. Of the 669 MB retained saving, roughly 256 MB comes from the format-level fixes alone (measured with interning disabled while it was still switchable) and the rest from the pools.

Where the heap goes, class by class (pre-GC histogram, same pair of runs):

| class | stock | this PR | Δ | cause |
|---|---|---|---|---|
| `java.util.RegularEnumSet` | 8,224,311 / 251.0 MiB | **absent** (< 1.7 MiB) | −251 MiB | shared scope sets |
| `sbt.internal.inc.UsedName` | 8,224,298 / 188.2 MiB | **396,197 / 9.1 MiB** | −179 MiB | UsedName pools |
| `byte[]` (string bodies) | 3.33 M / 208.5 MiB | 1.05 M / 80.7 MiB | −128 MiB | string pool |
| `xsbti.api.Projection` | 4,120,137 / 94.3 MiB | 509,459 / 11.7 MiB | −83 MiB | per-read node cache |
| `xsbti.api.Singleton` | 4,092,961 / 62.5 MiB | **absent** | −62 MiB | per-read node cache |
| `readUsedNameSet` wrappers (`JSetWrapper`, `SetHasAsScala`, lambda) | 3.22 M / 57.3 MiB | **absent** | −57 MiB | no `EnumSet.asScala` per name |
| `[Lxsbti.UseScope;` | 1,543,366 / 52.6 MiB | **absent** | −53 MiB | cached `values()` |
| `java.lang.String` | 3,119,293 / 71.4 MiB | 1,042,455 / 23.9 MiB | −48 MiB | string pool |
| `xsbti.api.Annotation` + args + arrays | 960,212 / 22.0 MiB | 210,975 / 4.8 MiB | −17 MiB | per-read node cache |
| `xsbti.api.NameHash` | 15,727,923 / 360.0 MiB | 15,727,923 / 360.0 MiB | 0 | not interned, by design |
| `[Lxsbti.api.NameHash;` | 241,470 / 64.1 MiB | 241,470 / 64.1 MiB | 0 | not interned, by design |
| pool bookkeeping (`WeakValue`, `KeyedWeakValue`, their map nodes) | 0 | 2.81 M / ~85.6 MiB | **+86 MiB** | cost side |

`NameHash` and its arrays are byte-identical between the two columns (15,727,923 instances, 377,470,152 bytes), which is the clearest confirmation that this PR leaves it untouched.

The interned `UsedName` count equals the corpus-global distinct population exactly (396,197) — the pools are perfect. Bookkeeping is what interning costs: one weak reference plus one map node per canonical value — 1,007,113 strings (30.7 MiB of `WeakValue`) and 396,197 `UsedName`s (12.1 MiB of `KeyedWeakValue`), plus ~42.8 MiB of `ConcurrentHashMap` nodes.

### CPU: zinc's own `AnalysisFormatBenchmark`

Read path, the only one that interns, with enough iterations to settle (`-f1 -wi 5 -i 10 …readConsistentBinary`):

| build | `readConsistentBinary` | vs stock |
|---|---|---|
| stock `develop` | 95.546 ± 0.627 ms | — |
| this PR | **99.038 ± 2.274 ms** | **+3.7%** |
| earlier revision of this PR, Guava pools | 102.919 ± 3.540 ms | +7.7% |

The whole benchmark class as a control (`-f1 -wi 3 -i 5 xsbt.AnalysisFormatBenchmark`, same session):

| benchmark | stock `develop` | this PR | Δ |
|---|---|---|---|
| `readConsistentBinary` | 94.332 ± 3.947 ms | 95.571 ± 5.328 ms | +1.3% |
| `writeConsistentBinary` | 136.822 ± 3.699 ms | 134.833 ± 1.539 ms | −1.5% |
| `writeConsistentBinaryNoSort` | 63.941 ± 3.090 ms | 61.687 ± 0.371 ms | −3.5% |
| `writeNull` | 122.548 ± 9.534 ms | 111.511 ± 0.151 ms | −9.0% |
| `writeNullNoSort` | 51.364 ± 0.589 ms | 46.341 ± 0.564 ms | −9.8% |

This PR barely touches serialization, so the four write benchmarks are a control — and they come out *faster* here, which is the point: at five iterations the run-to-run drift on this machine is several percent in either direction, so read cost should be read off the tighter run above (+3.7%), not from this table.

### CPU: loading the whole corpus

582 analyses in a 4 GB heap, median of three runs each:

| workload | stock `develop` | this PR | Δ |
|---|---|---|---|
| sequential | 4110 ms | 4857 ms | +18% |
| 8 threads | 2032 ms | 2383 ms | +17% |

Both builds scale identically from 1 to 8 threads — stock 2.02×, this PR 2.04× — so the pools add per-call work, not lock contention: `ReferenceQueue.poll()` returns on a plain volatile read while the queue is empty, and the maps are `ConcurrentHashMap`s. (The Guava-based revision measured 2387 ms at 8 threads, indistinguishable from the hand-rolled 2383 ms.)

### End to end in a real build

Published this branch as zinc `2.0.0-intern-SNAPSHOT`, built sbt against it, and ran `Test / compileIncremental` over a 303-module monorepo (603 analyses). This run predates the switch from Guava to the hand-rolled pools; the pooled populations it reports are a property of the pooling scheme, which is unchanged.

- **`[success]`, 710 s**, no errors.
- Peak heap 16914 MB, retained after forced GC 4761 MB (project's own `-Xmx20G -Xms8G`, G1).
- In the live server, `UsedName` held **400,158 instances / 9.2 MiB** — within 1% of the corpus-global distinct population — while compiling new modules *and* holding 600+ analyses. So the pools stay canonical under a real parallel compile, not only when reading files.
- `RegularEnumSet` was **absent from the entire histogram**.
- `NameHash` was 704 MiB / 30.8 M instances (+125 MiB of arrays), again the dominant analysis-side cost and untouched here.
- Pool bookkeeping totalled 54.5 MiB (Guava entries at the time; the hand-rolled pools cost about a third more per entry).

## Compatibility

- Analysis format version unchanged (**1100029**) — existing analysis files stay readable, and files written by this build are readable by stock zinc. Verified against released zinc 2.0.0.
- No public API change; `AnalysisInterner` is `sbt.internal.inc`.
- Api hashes unchanged (asserted by test), so incremental invalidation behaves identically.
- No new dependency: the weak pools are hand-rolled on the JDK's `ConcurrentHashMap` and `ReferenceQueue`.

## Testing

- `WeakPoolSpec` — canonicalization and separation of unequal values; 16-thread convergence; release of values once unreachable; for the weak-valued pool, `get`/`putIfAbsent` semantics and release of the *key* along with its dead value.
- `AnalysisInternerSpec` — string canonicalization; one `UsedName` per (name, scope) with probe-before-construct; scope-combination distinction; control-character names pooled under their escaped form; 16-thread convergence; release once unreachable; retention while referenced.
- `ConsistentAnalysisFormatInternerSuite` — strings shared across independent reads; node cache within one read; `UsedName`s shared across independent reads; api hashes preserved.
- Full build green: `zincRoot/testFull` — 221 scalatest tests + 78 property checks, 0 failures, including the e2e `IncrementalCompilerSpec`/`MultiProjectIncrementalSpec` suites that exercise fresh compilation. `scalafmtCheckAll`, `scalafmtSbtCheck`, `headerCheck` and `Test/headerCheck` clean.

### DIFF
--- a/internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisInterner.scala
+++ b/internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisInterner.scala
@@ -1,0 +1,91 @@
+/*
+ * Zinc - The incremental compiler for Scala.
+ * Copyright Scala Center, Lightbend, and Mark Harrah
+ *
+ * Licensed under Apache License 2.0
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package sbt.internal.inc
+
+import java.util.{ EnumSet => JEnumSet }
+
+import xsbti.UseScope
+
+/**
+ * The process-wide canonicalizer for strings and `UsedName`s: fresh
+ * (compiler-produced) and persisted (deserialized) analyses both route through
+ * it, so co-resident analyses share a single instance per distinct value.
+ *
+ * Strings are canonicalized through a weak interner: canonical instances are
+ * held only through weak references, so once every analysis referencing a
+ * string is released the canonical instance becomes eligible for GC.
+ *
+ * `UsedName`s are canonicalized without constructing a candidate first: a
+ * `UsedName` is fully determined by its (already canonical) name string and one
+ * of the eight possible `UseScope` combinations, so each combination gets a
+ * weak-valued name-keyed pool. A pool entry is removed once the canonical
+ * `UsedName` is no longer referenced by any analysis. On a pool hit no
+ * `UsedName` is constructed.
+ *
+ * Both pools make interning leak-free across the lifetime of a long-running
+ * build server. Benchmarks that need an uninterned baseline compare against a
+ * stock zinc checkout; production code always interns.
+ */
+object AnalysisInterner {
+
+  private[inc] final val DEFAULT_SCOPE = 1
+  private[inc] final val IMPLICIT_SCOPE = 2
+  private[inc] final val PAT_MAT_TARGET_SCOPE = 4
+  private final val SCOPE_COMBINATIONS = 8 // every subset of the three scopes above
+
+  private val stringPool = new WeakInterner[String]
+
+  // One weak-valued pool per UseScope combination, keyed by the canonical name.
+  private val usedNamePools: Array[WeakValuePool[String, UsedName]] =
+    Array.fill(SCOPE_COMBINATIONS)(new WeakValuePool[String, UsedName])
+
+  def internString(s: String): String = stringPool.intern(s)
+
+  /**
+   * The canonical `UsedName` for `name` used in the scope combination
+   * `scopeBits`, an or-ing of `DEFAULT_SCOPE`, `IMPLICIT_SCOPE` and
+   * `PAT_MAT_TARGET_SCOPE`. Constructs a `UsedName` only when the value is not
+   * pooled yet.
+   */
+  def usedName(name: String, scopeBits: Int): UsedName = {
+    // UsedName stores the escaped name, so the pool must be probed with it too;
+    // escaping is a no-op unless the name contains control characters.
+    val escaped = UsedName.escapeControlChars(name)
+    val pool = usedNamePools(scopeBits)
+    val existing = pool.get(escaped)
+    if (existing != null) existing
+    else {
+      val fresh = UsedName.make(escaped, scopeSets(scopeBits))
+      val prev = pool.putIfAbsent(fresh.name, fresh)
+      if (prev == null) fresh else prev
+    }
+  }
+
+  /**
+   * The eight possible scope sets, indexed by scope bits. Shared by every
+   * UsedName and treated as immutable -- they must never be mutated.
+   */
+  private[inc] val scopeSets: Array[JEnumSet[UseScope]] =
+    Array.tabulate(SCOPE_COMBINATIONS) { bits =>
+      val scopes = JEnumSet.noneOf(classOf[UseScope])
+      if ((bits & DEFAULT_SCOPE) != 0) scopes.add(UseScope.Default)
+      if ((bits & IMPLICIT_SCOPE) != 0) scopes.add(UseScope.Implicit)
+      if ((bits & PAT_MAT_TARGET_SCOPE) != 0) scopes.add(UseScope.PatMatTarget)
+      scopes
+    }
+
+  /** The inverse of `scopeSets`: also the encoding used by the persisted format. */
+  private[inc] def scopeBits(scopes: JEnumSet[UseScope]): Int =
+    (if (scopes.contains(UseScope.Default)) DEFAULT_SCOPE else 0) |
+      (if (scopes.contains(UseScope.Implicit)) IMPLICIT_SCOPE else 0) |
+      (if (scopes.contains(UseScope.PatMatTarget)) PAT_MAT_TARGET_SCOPE else 0)
+}
--- a/internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/ConsistentAnalysisFormat.scala
+++ b/internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/ConsistentAnalysisFormat.scala
@@ -12,7 +12,7 @@
 package sbt.internal.inc.consistent
 
 import java.nio.file.Paths
-import java.util.{ Arrays, Comparator, EnumSet }
+import java.util.{ Arrays, Comparator }
 import sbt.internal.inc.{ UsedName, Stamp => StampImpl, _ }
 import sbt.internal.util.Relation
 import sbt.util.InterfaceUtil
@@ -192,7 +192,7 @@ class ConsistentAnalysisFormat(val mappers: ReadWriteMappers, reproducible: Bool
       val bh = in.long()
       val ebh = in.long()
       val nhNames = in.readStringArray()
-      val nhScopes = in.readArray[UseScope]() { UseScope.values()(in.byte().toInt) }
+      val nhScopes = in.readArray[UseScope]() { useScopeValues(in.byte().toInt) }
       val nhHashes = in.readArray[Int]() { in.int() }
       val nameHashes = new Array[NameHash](nhNames.length)
       var i = 0
@@ -265,7 +265,7 @@ class ConsistentAnalysisFormat(val mappers: ReadWriteMappers, reproducible: Bool
   private def readSourceInfos(in: Deserializer): SourceInfos = {
     def readProblem(): Problem = in.readBlock {
       val category = in.string()
-      val severity = Severity.values.apply(in.byte().toInt)
+      val severity = severityValues(in.byte().toInt)
       val message = in.string()
       val rendered = Option(in.string())
       def io(): Option[Integer] = in.int() match { case -1 => None; case i => Some(i) }
@@ -339,7 +339,7 @@ class ConsistentAnalysisFormat(val mappers: ReadWriteMappers, reproducible: Bool
       val scalacOptions = in.readArray() { readMapper.mapScalacOption(in.string()) }
       val javacOptions = in.readArray() { readMapper.mapJavacOption(in.string()) }
       val compilerVersion = in.string()
-      val compileOrder = CompileOrder.values()(in.byte().toInt)
+      val compileOrder = compileOrderValues(in.byte().toInt)
       val skipApiStoring = in.bool()
       val extra = in.readArray(2) { InterfaceUtil.t2(in.string() -> in.string()) }
       val outputPath = in.string()
@@ -434,12 +434,7 @@ class ConsistentAnalysisFormat(val mappers: ReadWriteMappers, reproducible: Bool
   private def writeUsedNameSet(out: Serializer, uns: scala.collection.Set[UsedName]): Unit = {
     out.writeBlock("UsedName") {
       val groups0 = uns.iterator.map { un =>
-        val sc = un.scopes
-        var i = 0
-        if (sc.contains(UseScope.Default)) i += 1
-        if (sc.contains(UseScope.Implicit)) i += 2
-        if (sc.contains(UseScope.PatMatTarget)) i += 4
-        (un.name, i.toByte)
+        (un.name, AnalysisInterner.scopeBits(un.scopes).toByte)
       }.toArray.groupBy(_._2)
       val groups = if (reproducible) groups0.toVector.sortBy(_._1) else groups0
       out.writeColl("groups", groups, 2) { case (g, gNames) =>
@@ -452,12 +447,14 @@ class ConsistentAnalysisFormat(val mappers: ReadWriteMappers, reproducible: Bool
   }
 
   private def readUsedNameSet(in: Deserializer): Set[UsedName] = {
-    import scala.jdk.CollectionConverters.*
     in.readBlock {
+      // The name and the scope bits fully determine a UsedName, so the interner
+      // can return the pooled instance without constructing a candidate first.
       val data = in.readColl[Vector[UsedName], Vector[Vector[UsedName]]](Vector, 2) {
-        val i = in.byte().toInt
-        val names = in.readStringSeq()
-        names.iterator.map { n => UsedName(n, useScopes(i).asScala) }.toVector
+        val scopeBits = in.byte().toInt
+        in.readColl[UsedName, Vector[UsedName]](Vector) {
+          AnalysisInterner.usedName(in.string(), scopeBits)
+        }
       }
       data.flatten.toSet
     }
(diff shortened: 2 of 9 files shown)


eed3si9n:
Thanks for the contribution and the analysis.

build.sbt
```diff
@@ -349,7 +349,8 @@ lazy val zincCore = (projectMatrix in internalPath / "zinc-core")
       exclude[MissingClassProblem]("xsbti.*"),
     ),
     libraryDependencies ++= List(
-      "org.scala-lang.modules" %% "scala-parallel-collections" % "1.2.0"
+      "org.scala-lang.modules" %% "scala-parallel-collections" % "1.2.0",
+      "com.google.guava" % "guava" % "33.6.0-jre"
```
eed3si9n: > The pools use `Interners.newWeakInterner()` and `MapMaker().weakValues()`. If a new dependency is unwelcome, a hand-rolled weak pool (`ReferenceQueue` + `ConcurrentHashMap`) is ~60 lines and I am happy to swap it in. 

I'd like to avoid depending on Guava since it's notoriously prevalent, and people use different versions of it. Hand-rolled would be preferable, but Caffeine is ok since sbt already uses it.

internal/zinc-core/src/main/scala/sbt/internal/inc/AnalysisInterner.scala
```diff
@@ -0,0 +1,92 @@
...
+   * UsedName and treated as immutable -- they must never be mutated. The bit
+   * scheme must stay in sync with `ConsistentAnalysisFormat.writeUsedNameSet`
+   * (zinc-persist), which encodes the same bits on the write side.
+   */
+  private[inc] val scopeSets: Array[JEnumSet[UseScope]] =
+    Array.tabulate(8) { i =>
+      val e = JEnumSet.noneOf(classOf[UseScope])
+      if ((i & 1) != 0) e.add(UseScope.Default)
+      if ((i & 2) != 0) e.add(UseScope.Implicit)
+      if ((i & 4) != 0) e.add(UseScope.PatMatTarget)
```
eed3si9n: Consider naming these bits as `final val DEFAULT_SCOPE = 1` etc.

internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/ConsistentAnalysisFormat.scala
```diff
@@ -452,14 +454,25 @@ class ConsistentAnalysisFormat(val mappers: ReadWriteMappers, reproducible: Bool
...
+      // The name string and scope bits fully determine a UsedName, so the
+      // interner can return the pooled instance without constructing one.
+      val nGroups = in.startArray() / 2
+      val b = Set.newBuilder[UsedName]
+      var g = 0
+      while (g < nGroups) {
+        val scopeBits = in.byte().toInt
+        val nNames = in.startArray()
+        var j = 0
+        while (j < nNames) {
```
eed3si9n: Using `var` and `while` isn't idiomatic Scala. I suggest sticking with functional constructs.

build.sbt
hoangmaihuy: Hand-rolled in `WeakPools.scala`

eed3si9n (approved):
Thanks!

PR #1767, opened 2026-08-13
fix(persistence): prevent quadratic API node interning

nguyenyou:
Follow-up to #1754.

The per-read API-node cache uses node values directly as `HashMap` keys. Unequal API nodes can share the generated 32-bit hash, causing recursive tree-bin searches because these nodes are not `Comparable`.

This replaces the map with a compact open-addressed cache using shallow 64-bit fingerprints of already-canonical child nodes. Interning behavior and API hashes remain unchanged.

### Results

For 2,048 colliding `Projection` nodes:

| | Before | After |
|---|---:|---:|
| Time | 20.457 ms | 0.036 ms |
| Allocation | 148,078 B | 98,424 B |

The existing analysis-read benchmark remains neutral:

- Before: 98.739 ± 2.906 ms
- After: 97.565 ± 2.732 ms

Added a deterministic collision benchmark and regression coverage. `zincRoot/testFull`, formatting, and header checks pass.

### DIFF
--- a/internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/NodeCache.scala
+++ b/internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/NodeCache.scala
@@ -70,10 +70,26 @@ private[consistent] final class NodeCache {
 }
 
 private[consistent] object NodeCache {
-  private final val Offset = -3750763034362895579L
-  private final val Prime = 1099511628211L
+  private final val ParameterRefTag = 1L
+  private final val ParameterizedTag = 2L
+  private final val PolymorphicTag = 3L
+  private final val ConstantTag = 4L
+  private final val ExistentialTag = 5L
+  private final val SingletonTag = 6L
+  private final val ProjectionTag = 7L
+  private final val AnnotatedTag = 8L
+  private final val TypeParameterTag = 9L
+  private final val AnnotationTag = 10L
 
-  private def mix(hash: Long, value: Long): Long = (hash ^ value) * Prime
+  // FNV-1a 64-bit offset basis and prime.
+  private final val Fnv1a64OffsetBasis = -3750763034362895579L
+  private final val Fnv1a64Prime = 1099511628211L
+
+  // MurmurHash3 fmix64 multipliers.
+  private final val MurmurHash3Fmix64Multiplier1 = -49064778989728563L
+  private final val MurmurHash3Fmix64Multiplier2 = -4265267296055464877L
+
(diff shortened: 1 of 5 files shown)


eed3si9n:
Thanks for the contribution! Overall looks like a great change.

internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/NodeCache.scala
```diff
@@ -0,0 +1,141 @@
...
+        result = mixReference(result, argument.name())
+        result = mixReference(result, argument.value())
+        index += 1
+      }
+      result
+    }
+  }
+
+  private[consistent] def fingerprint(value: AnyRef): Long = value match {
+    case node: ParameterRef => mixReference(mix(Offset, 1L), node.id())
```
eed3si9n: Could we define a `final val` constant for 1L, 2L etc?

internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/NodeCache.scala
```diff
@@ -0,0 +1,141 @@
...
+      mixArguments(mixReference(mix(Offset, 10L), node.base()), node.arguments())
+    case node => mix(Offset, Integer.toUnsignedLong(node.hashCode()))
+  }
+
+  private def index(hash: Long, length: Int): Int = {
+    var mixed = hash
+    mixed = (mixed ^ (mixed >>> 33)) * -49064778989728563L
+    mixed = (mixed ^ (mixed >>> 33)) * -4265267296055464877L
+    ((mixed ^ (mixed >>> 33)).toInt & Int.MaxValue) & (length - 1)
+  }
```
eed3si9n, line 177: For future reference, could you cite the origin of the prime numbers? e.g. Murmur3

internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/NodeCache.scala
nguyenyou: Done — each node kind now has a named `private final val` tag.

internal/zinc-persist/src/main/scala/sbt/internal/inc/consistent/NodeCache.scala
nguyenyou: Done — the constants are now named and documented as the FNV-1a 64-bit offset basis and prime, and the MurmurHash3 `fmix64` multipliers.

eed3si9n (approved):
Thanks!

PR #1769, opened 2026-08-21
Drop Java serialization from text analysis format

stasimus:
The legacy text analysis format read its companions and compilation blobs with a raw `ObjectInputStream.readObject`, with no class filter. This swaps that for the structural codec the default binary/consistent format already uses, so the text store no longer depends on Java serialization at all. Only `FileAnalysisStore.text` (`useTextAnalysis = true`) is affected; the default format never used it.

I'm treating this as hardening/cleanup, removing an unfiltered deserialization sink, rather than a strong RCE claim (see discussion below). Companions reuse the existing `ClassLike` codec from `ConsistentAnalysisFormat`; `Compilation` encodes its start time and output directly. I bumped the text format version 7 -> 8 so an old cache is rejected cleanly instead of being fed to the new reader, and regenerated the `test-data` fixtures.

Tests: `DeserializationVulnerabilitySpec` (a Java-serialized payload is no longer deserialized) and `CompilationFormatSpec` (single/multiple/empty output round-trip).


eed3si9n:
> If an analysis file can come from somewhere you don't fully trust (a shared or remote build cache, a CI artifact), a crafted file plus a gadget on the classpath gives code execution at compile time.

I'm not against the proposed change itself, but I'm not sure if the premise would hold. If the CI artifact can be compromised, why not just poison the JAR artifacts of sbt, plugins, or the scala-library JAR that would show up on classpath?

stasimus:
Yes, 3 if should happen to use an attack, so the idea I see Java serialization - I see the risk

stasimus:
Fair point, and I agree. If the cache or artifact is writable by an attacker, they can already poison the JARs on the classpath, so this isn't really a privilege escalation on top of that.

I've updated the description to reflect this: it's framed as removing an unfiltered `ObjectInputStream.readObject` sink from the text format (and aligning it with the structural codec the default format already uses) rather than a meaningful RCE vector.

eed3si9n (approved):

PR #1787, opened 2026-09-11
Fix three over-invalidations in trait extraHash

lrytz:
`extraHash` covers the private members of a trait, which leak into the classes that inherit it. Fix three defects that made it invalidate more than it should (over-compilation, never wrong output). One commit each, each turning a pending scripted test into a passing one.

- A companion object's hash was part of the trait's, so an unrelated member added to `object A` recompiled everything inheriting from trait A.
- A trait's parents were taken from the previous analysis, so the same source hashed differently on a cold and a warm build, and a comment-only edit recompiled the inheritance subtree.
- `object B extends A` counted as trait B inheriting A, so a private change in A recompiled everything inheriting from B.

The last adds `AnalysisCallback4`, qualifying the class names in a dependency with their namespace. Bridges that do not report it behave as before.

A fourth defect, merged companion name hashes, stays pending: fixing it would change the analysis format.


eed3si9n:
Thanks for the contribution! Per https://github.com/sbt/zinc/blob/develop/CONTRIBUTING.md#start-with-an-issue, could we create separate issue for these to discuss if they need fixing, and also split the PRs into one fix per PR please?
Specifically around the companion changes, I think they were originally introduced in https://github.com/sbt/zinc/pull/542 for invalidating trait inheritance, but I also wonder if we should keep them around since companions are in the implicit scope.

PR #1789, opened 2026-09-11
fix: pass Java sources to scalac when pipelining

### LINKED ISSUE #9707: usePipelining := true fails to compile JVM sub-projects that have Java sources
## steps

1. `git clone --depth 1 -b use-pipelining https://github.com/plokhotnyuk/jsoniter-scala`
2. `cd jsoniter-scala`
3. `sbt clean +test`

## problem

[error] (jsoniter-scala-coreJVM / Compile / compileIncremental) Compilation failed

## expectation

Successful compilation

## notes

Setting `compileOrder := CompileOrder.JavaThenScala` doesn't help

stasimus:
Fixes sbt/sbt#9707.

With pipelining javac is deferred, so its output is not on the classpath yet. But MixedAnalyzingCompiler only hands Java sources to scalac when compileOrder is Mixed, so JavaThenScala and ScalaThenJava fail to resolve Java symbols from the same project.

The scripted test needed a compileOrder field in build.json.

### DIFF
--- a/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
+++ b/zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
@@ -177,8 +177,11 @@ final class MixedAnalyzingCompiler(
           case _ => scalacOpts1
         }
         JarUtils.withPreviousJar(output) { (extraClasspath: Seq[Path]) =>
+          // Under pipelining javac is deferred, so its output is not on the classpath
+          // yet. Java sources must reach scalac regardless of the configured order,
+          // otherwise Scala sources cannot resolve symbols defined in Java.
           val sources =
-            if (config.currentSetup.order == Mixed) incSrc
+            if (config.currentSetup.order == Mixed || config.incOptions.pipelining) incSrc
             else scalaSrcs
 
           val cp0: Vector[VirtualFile] =
--- a/internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
+++ b/internal/zinc-scripted/src/test/scala/sbt/internal/inc/IncHandler.scala
@@ -65,7 +65,8 @@ final case class Project(
     name: String,
     dependsOn: Option[Vector[String]] = None,
     in: Option[Path] = None,
-    scalaVersion: Option[String] = None
+    scalaVersion: Option[String] = None,
+    compileOrder: Option[String] = None
 )
 
 final case class Build(projects: Seq[Project])
@@ -125,6 +126,7 @@ class IncHandler(directory: Path, cacheDir: Path, scriptedLog: ManagedLogger, co
       val in: Path = p.in.getOrElse(directory / p.name)
       val version = switchScalaVersion(p.scalaVersion)
       val deps = p.dependsOn.toVector.flatten
+      val order = p.compileOrder.fold(CompileOrder.Mixed)(CompileOrder.valueOf)
       val project = ProjectStructure(
         p.name,
         deps,
@@ -134,7 +136,8 @@ class IncHandler(directory: Path, cacheDir: Path, scriptedLog: ManagedLogger, co
         lookupProject,
         version,
         compileToJar,
-        incrementalCompiler
+        incrementalCompiler,
+        order
       )
       buildStructure(p.name) = project
     }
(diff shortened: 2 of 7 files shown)


zinc/src/main/scala/sbt/internal/inc/MixedAnalyzingCompiler.scala
```diff
@@ -177,8 +177,11 @@ final class MixedAnalyzingCompiler(
           case _ => scalacOpts1
         }
         JarUtils.withPreviousJar(output) { (extraClasspath: Seq[Path]) =>
+          // Under pipelining javac is deferred, so its output is not on the classpath
+          // yet. Java sources must reach scalac regardless of the configured order,
+          // otherwise Scala sources cannot resolve symbols defined in Java.
           val sources =
-            if (config.currentSetup.order == Mixed) incSrc
+            if (config.currentSetup.order == Mixed || config.incOptions.pipelining) incSrc
```
eed3si9n, line 184: Pipelined build needs to run Java compiler at a specific timing, so user specifying either JavaThenScala or ScalaThenScala should probably fail immediately?

eed3si9n: Similar comment I made in 2024 apparently - https://github.com/sbt/sbt/issues/7546#issuecomment-2362602654

> https://github.com/sbt/sbt/pull/5703 that I sent shipped pipelining in sbt 1.4.0, though, many people have contributed a bunch of components leading up to it. The diagram attempts to illustrate the potential time-saving that the build pipelining could achieve by starting the downstream Scala compilation earlier than the production of the whole JARs, by emitting a fake early JARs. This currently requires close coordination between the Scala compiler and the Java compiler timing, as my comment apparently said. This likely means that it's fundamentally incompatible with `CompilerOrder.JavaThenScala`, at least in the way pipelining is implemented in sbt and Zinc.

stasimus: Agreed, I'll make it fail instead.

Ran the repro first. On the use-pipelining branch jsoniter-scala-coreJVM gives 100 "not found: value ByteArrayAccess" errors, and switching only JavaThenScala to Mixed compiles clean, downstream macros included. So it's just the explicit order that breaks, which means users have a one line fix today.

Two scope questions before I push:

1. Only JavaThenScala? ScalaThenJava already matches what the pipelined pass does (scalac first, javac deferred) and works today, so failing it would break builds that are currently fine.
2. Gate the error on the subproject actually having Java sources? compileOrder often gets set at ThisBuild, otherwise we'd fail projects with no Java in them at all.

eed3si9n: > 1. Only JavaThenScala?

I guess so.

> 2. Gate the error on the subproject actually having Java sources? 

If we have to run source generator to get this answer, it might be cheaper to just reject JavaThenScala regardless of the presence of Java sources.

stasimus: Implemented the unconditional reject, but it regresses pure Scala subprojects. ThisBuild / compileOrder := JavaThenScala is a normal setting, and with pipelining on the check then kills every subproject in the build. Scripted case with a single Scala file and no Java at all:

    [error] ## Exception when compiling 1 sources to /tmp/sbt_8b21ebc4/use/target/classes
    [error] java.lang.IllegalArgumentException: CompileOrder.JavaThenScala cannot be
            combined with build pipelining, ...
    [error] sbt.internal.inc.MixedAnalyzingCompiler.<init>(MixedAnalyzingCompiler.scala:52)

Those builds compile fine today.

The source generator cost doesn't apply here either. MixedAnalyzingCompiler already splits the resolved sources into javaSrcs and scalaSrcs at the top of compile, so gating on both being non empty is free at the zinc layer.

That trace also shows handleCompilationError printing a config mistake as "## Exception when compiling N sources" with a full stack trace, so it reads like a zinc crash. Better as a CompileFailed.

Unless you'd rather keep it strict, I'll move the check into compile, gate it on the subproject actually mixing Java and Scala, and report it as a compile failure.

eed3si9n: > The source generator cost doesn't apply here either. MixedAnalyzingCompiler already splits the resolved sources into javaSrcs and scalaSrcs at the top of compile, so gating on both being non empty is free at the zinc layer.

Yea, that's true.

> Unless you'd rather keep it strict, I'll move the check into compile, gate it on the subproject actually mixing Java and Scala, and report it as a compile failure.

I'm ok either way, but it is technically not a compilation error since we're saying that the set up is wrong, not the code.

stasimus: Fair point, so not a CompileFailed. Added InvalidCompileSetup, thrown from MixedAnalyzingCompiler and gated on the project having both Java and Scala sources.

eed3si9n (approved):
Thanks!

PR #1799, opened 2026-09-15
fix: Fixes overcompilation of inheritance

### LINKED ISSUE #1794: A comment-only edit to a trait recompiles the classes that inherit it
```scala
// A.scala
trait A

// B.scala
trait B extends A

// D.scala
class D extends B
```

Adding a comment to `B.scala` causes `D` to be recompiled, in an extra invalidation round.

### Cause

`AnalyzedClass.extraHash` for a trait includes the extra-hash values of its parents, so that a private change in `A` reaches `D` (added in sbt/zinc#1289).

The list-of-parents and their hashes were taken from the previous analysis, not from the compilation that is running. On a clean build there is no previous analysis, so nothing is folded into `B.extraHash`. On the next compile `A` is there and gets folded in, so `B.extraHash` changes although `B` did not, and `D` is recompiled.

The same happens when an `extends` clause is edited: the inheritance relation recorded by the current compile is right, but the hash stored next to it is one compile behind.

Fix: take the parents and their hashes from the cycle being compiled. The same source then always produces the same `extraHash`, on a cold build and a warm one.

lrytz:
The `extraHash` of a trait is assembled from its parents `extraHash` values. The list of parents is taken from the previous analysis, which can lead to extra compilation rounds if it's not up to date.

This PR uses up to date parents information from the dependency phase.

Early output (for pipelining) is delayed, now after the dependency phase (instead of the API phase, which runs right before).

Fixes https://github.com/sbt/zinc/issues/1794


internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -1133,25 +1209,7 @@ private final class AnalysisCallback(
...
-            !continue && invalidations.isEmpty,
-            "everything was supposed to be invalidated already"
-          )
-          invalidationResults = Some(CompileCycleResult.empty)
-          writeEarlyArtifacts(merged)
-        }
-      }
-    }
-  }
+  override def apiPhaseCompleted(): Unit = ()
```
eed3si9n, line 1212: So while this change is correct (would prevent over-compilation in the next round), for clean compilation of a pipelined build, we might see slight regression in perf?

eed3si9n (approved):
Thanks!

internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
lrytz: Yes, early output is now after the dependency phase, which runs right after the API phase.

There's a 2016 idea to merge the two phase into one, but it didn't make it so far. https://github.com/adriaanm/scala/commits/sbt-api-consolidate

PR #1801, opened 2026-09-15
fix: Fixes adding a member to companion object invalidating inheritance

### LINKED ISSUE #1793: Adding a member to an object recompiles classes inheriting its companion trait
```scala
// A.scala
trait A
object A { def x: Int = 1 }

// B.scala
trait B extends A

// D.scala
class D extends B
```

Adding `def y: Int = 2` to `object A` causes `D` to be recompiled.

### Cause

`AnalyzedClass.extraHash` is a value that only has meaning for traits. It changes whenever a trait (or one of its parents) changes such that subclasses of the trait need re-compilation. In the example: adding a `private val y: Int = 2` to `trait A` requires re-compiling `D`.

In zinc, there's a single `AnalyzedClass` instance for `A` (trait and object). Its `extraHash` is assembled from the computed extra-hash values for both `trait A` and `object A`. This means `A.extraHash` changes when adding a method to `object A`, causing `B.extraHash` to change, which invalidates `D`.

Fix: `extraHash` for `A` should not include the computed extra-hash value of `object A`. Changes to the `object A` API are still tracked by `AnalyzedClass.apiHash`.

lrytz:
This PR builds on top of https://github.com/sbt/zinc/pull/1799, so that should be merged first.

AnalyzedClass.extraHash is a value that only has meaning for traits.

With this PR, a companion object's hash no longer goes into the `extraHash` of a trait, so changing the members of an `object A` no longer recompiles classes that extend `trait A`.

Fixes https://github.com/sbt/zinc/issues/1793


internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -1043,17 +1043,7 @@ private final class AnalysisCallback(
             hash
         }
 
-    // A companion pair publishes one extraHash, merged from the class and the object side.
     private def compute(className: String): HashAPI.Hash =
-      if (apis.contains(className) || objects.contains(className))
-        (classSide(className), objects.get(className).fold(emptyApiHash)(_.extraHash)).hashCode
```
lrytz, line 1049: The main change is here, from

`(classSide(className), objects.get(className).fold(emptyApiHash)(_.extraHash)).hashCode`

to basically

`classSide(className)`

lrytz:
(Test failure on Windows due to https://github.com/sbt/zinc/pull/1802)

eed3si9n (approved):
Thanks!

PR #1802, opened 2026-09-15
Run each scripted test in its own directory

lrytz:
Tests in a batch shared one directory that was cleared between tests. On Windows the clearing can leave files behind, and a JAR left under `lib/` by one test ended up on the next test's classpath, so `constants` compiled against another test's `object A`.


eed3si9n (approved):
Thanks!

PR #1803, opened 2026-09-15
fix: Fixes compound type in a member signature getting recorded as inheritance

### LINKED ISSUE #1798: A compound type in a member signature is recorded as inheritance
```scala
// A.scala
trait A { def foo: AnyRef with B = ??? }

// B.scala
trait B
```

Adding an unrelated `def baz: Int = 1` to `B` causes `A` to be recompiled. With `def foo: B` instead, it isn't.

```
API change: B: modified names $init$, baz
  transitive inheritance: A, B
```

### Cause

The Scala 2 bridge records `A` as inheriting from `B`. `AnyRef with B` is a `CompoundTypeTree` wrapping a `Template`, and the dependency traverser handles that `Template` like a class body.

Fix: handle `CompoundTypeTree` in the traverser by visiting its parents, self type and body without recording inheritance. The types still produce member-reference dependencies.

lrytz:
`trait A { def foo: AnyRef with B = ??? }`

Zinc records `B` as a parent of `A` on Scala 2. Prevent this by handling the `Template` of a `CompoundType` separately.

Fixes https://github.com/sbt/zinc/issues/1798


eed3si9n (approved):
Thanks!

Issue #1796, opened 2026-09-14
Changing `class A.x` recompiles code that uses `object A.x`


lrytz:
```scala
// A.scala
class  A { def x: Int = 1 }
object A { def x: String = "" }

// B.scala
class B { val v: String = A.x }
```

Changing `x` in `class A` to `Long` causes `B` to be recompiled.

### Cause

Invalidation is name-granular: `B` is recompiled only if it uses a name whose hash changed in `A`. Adding an unrelated `def y` to `object A`, for instance, reports `modified names y` and recompiles nobody.

The class and object `A` share a single `AnalyzedClass`, and `nameHashesForCompanions` merges the name hashes of both halves into one array. So when the hash for `x` changes, the two methods cannot be told apart.

Fix: stop merging in `nameHashesForCompanions`, and teach invalidation which half a use refers to. That needs two things:
  - the namespace of each name passed to `AnalysisCallback.usedName`: a related change is suggested/required for [zinc#1795](https://github.com/sbt/zinc/issues/1795), so if we introduce `AnalysisCallback4`, it should support both tickets
  - separate name hashes per half: that changes the analysis format, `NameHash` and `UsedName` both gain a namespace

### Related, but not addressed here

A related conflation: Zinc also cannot tell whether a referenced name itself (not its owner) is a term or a type. So a class that mentions only the type `A` is invalidated when `object A` changes, and the other way round.

In practice this matters less. The name hash of `A` itself covers only the signature line of `class A` / `object A` (parents, modifiers, type parameters, self type). Every member gets its own name hash.

```scala
class A { def m: Int = 1 }
object A { def n: Int = 2 }

class B { def f(a: A): Int = a.m }   // mentions only the type A
```

Adding `def q` to `object A` leaves `B` alone: the modified name is `q`. `B` is recompiled only when the `object A` line itself changes, for example `object A extends Serializable`.

PR #1807, opened 2026-09-16
Distinguish objects and classes/traits in inheritance dependency tracking

### LINKED ISSUE #1795: A private change in a trait recompiles classes that do not inherit it
```scala
// A.scala
trait A { private var counter: Int = 0 }

// B.scala
trait B
object B extends A

// D.scala
class D extends B
```

Renaming `counter` causes `D` to be recompiled. `D` inherits `trait B`, which is unrelated to `A`.

### Cause

`AnalyzedClass.extraHash` for a trait includes the extra-hash values of its parents, so that a private change in a parent reaches the classes that inherit it (added in https://github.com/sbt/zinc/pull/1289).

The bug is that `trait A` is considered a parent of `trait B`, but in reality it's a parent of `object B`.

`object B extends A` and `trait B extends A` both reach the dependency analysis callback as `classDependency("A", "B", DependencyByInheritance)` with the names as plain strings, so zinc cannot tell the two cases apart and has to assume that `trait B` could extend `trait A`.

The fix requires changing the dependency callback API (introduce `AnalysisCallback4`) so we can distinguish term names (`object B`) from type names (`trait B`).

lrytz:
Dependency tracking currently does not tell apart `object B extends A` from `trait B extends A`, it just records an inheritance edge from `B` to `A`.

Given a plain `trait B` with a companion `object B extends A`, that edge makes subclasses of `B` depend on `A`. So changes in A cause subclasses of `B` to be recompiled, even though they are unrelated.

This PR adds `AnalysisCallback4` with new overloads to the dependency tracking API in `compiler-interface` so that the kind of name (term vs type) can be tracked.

I also included a new overload to `void usedName` that is not yet used in this PR, but preparation for fixing https://github.com/sbt/zinc/issues/1796. That new overload tells apart used name `x` in `(a: A).x` (owner is type A) vs `A.x` (owner is term A).

Fixes https://github.com/sbt/zinc/issues/1795


eed3si9n:
If the type of `object B` is `B.type`, can we fix this by making `B.type` extend `A`?

lrytz:
I like the idea, it would be nice if `AnalysisCallback4` could be avoided.

The problem is: for `object B extends A`, the bridge would emit `classDependency("A", "B.type", DependencyByInheritance)`. When running on an old zinc, this is wrong and leads to under-compilation. The compiler bridge would need a way to know the version of zinc it's talking to. That's what we can solve with the new `AnalysisCallback4`. And once we have that, we can make the representation not depend on strings (```object A { class `type` }``` is a corner case, but it would clash).

The new signature is

```java
    void classDependency(ClassRef onClass,
                         ClassRef sourceClass,
                         DependencyContext context);
```

we could go back to `String onClass`, the current implementation just uses the name, since in zinc there's a single `AnalyzedClass` for `class A` and `object A`. But I think we should keep it, in case the distinction becomes useful in the future.

On the binary side, we have

```java
    void binaryDependency(Path onBinaryEntry,
                          String onBinaryClassName,
                          ClassRef fromClass,
                          VirtualFileRef fromSourceFile,
                          DependencyContext context);
```

The `String onBinaryClassName` allows making the distinction between class and object (`A` vs `A$`). So it seems consistent to allow making it on the source side too.

-----

About the new `usedName` signature (which is in preparation for https://github.com/sbt/zinc/issues/1796):

```java
    void usedName(String className,
                  String name,
                  EnumSet<NameKind> ownerKinds,
                  EnumSet<UseScope> useScopes);
```

The `ownerKinds` allows identifying `(a: A).x` vs `A.x` and fixing 1796.

For the `name` we only have a string, no `NameKind`, so for `Foo.X` we still can't tell if `X` is a term or a type. That is very narrow though: it only bites when the same name exists as both a type and a term in one class, and then only for changes to the declarations themselves. Changes to other members of `object Foo` don't cause a reference to `Foo.X` to re-compile.

`String className` is fine because it doesn't matter if there name use appears in a class or its companion; they are in the same file, so will be both recompiled anyway.

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback.java
```diff
@@ -73,21 +74,19 @@ void binaryDependency(File onBinaryEntry,
...
      *                      knows that <code>onBinaryClassName</code> comes from.
-     * @param onBinaryClassName Dependent binary name.
+     * @param onBinaryClassName Binary name being depended on.
      *                 Binary name with JVM-like representation. Inner classes
      *                 are represented with '$'. For more information on the
      *                 binary name format, check section 13.1 of the Java
      *                 Language Specification.
-     * @param fromClassName Represent the class file name where
-     *                 <code>onBinaryClassName</code> is defined.
-     *                 Binary name with JVM-like representation. Inner classes
```
lrytz, line 83: "Binary name" was wrong here, it's a source class name.

internal/zinc-core/src/main/scala/sbt/internal/inc/Incremental.scala
```diff
@@ -730,11 +745,30 @@ private final class AnalysisCallback(
...
+  override def classDependency(
+      onClass: ClassRef,
+      sourceClass: ClassRef,
+      context: DependencyContext
+  ): Unit = internalClassDependency(onClass.name, sourceClass, context)
+
+  // Bridges that do not report name kinds land here. Type keeps every edge, as before;
+  // over-approximating only costs a recompile.
+  def classDependency(onClassName: String, sourceClassName: String, context: DependencyContext) =
+    internalClassDependency(onClassName, ClassRef.of(sourceClassName, NameKind.Type), context)
```
eed3si9n, line 770: Should the `String` version register for both NameKind.Type and NameKind.Term since we don't know which one?

lrytz: The comment above is unclear, sorry.. "Type keeps every edge, as before".

If you look at `internalClassDependency`, it does

```scala
...
      if (context == DependencyContext.DependencyByInheritance && sourceClass.kind == NameKind.Type)
        add(typeParents, sourceClassName, onClassName)
...
```

That's the only reference to `sourceClass.kind`. So passing `Type` here means "could be a Type".

eed3si9n: I'm wondering if the current version of compiler bridge like Scala 3.9.0 or 2.13.18 emit String `B` to this function.

eed3si9n: > So passing Type here means "could be a Type".

ok never mind then.

eed3si9n (approved):
Thanks!

PR #1809, opened 2026-09-18
Name the used-name kinds after the qualifier

lrytz:
`ownerKinds` reads as the symbol that declares a member, which in a compiler bridge is `sym.owner`. The set records the namespace a name is selected from, so name it that way.

Follow-up for https://github.com/sbt/zinc/pull/1807 while it's not yet released.


internal/compiler-interface/src/main/java/xsbti/AnalysisCallback4.java
```diff
@@ -87,12 +87,12 @@ void binaryDependency(Path onBinaryEntry,
      *
      * @param className The source class name that uses <code>name</code>.
      * @param name The source name used in <code>className</code>.
-     * @param ownerKinds Which side of a companion pair declares the member referred to.
+     * @param qualifierKinds Which side of a companion pair the name is selected from.
```
lrytz: Example:

```scala
trait T1 { def x = 1 }
trait T2 { def x = 2 }
class A extends T1
object A extends T2
```

For `(a: A).x` the qualifier is `Type`, for `A.x` the qualifier is `Term`. The owner of both `x` is a type, but that's not what we want to express here.

internal/compiler-interface/src/main/java/xsbti/AnalysisCallback4.java
```diff
@@ -77,22 +77,17 @@ void binaryDependency(Path onBinaryEntry,
...
-     * both is otherwise one name. Pass both kinds whenever the use does not say which one
-     * it refers to, as for an import selector, an inherited member that either side can
-     * override, or a structural member.
+     * with the namespace it is selected from: a class (Type), an object (Term), or either
+     * when the use does not say (as for an import selector).
      *
      * @param className The source class name that uses <code>name</code>.
      * @param name The source name used in <code>className</code>.
-     * @param ownerKinds Which side of a companion pair declares the member referred to.
+     * @param qualifierKinds The namespaces <code>name</code> may be selected from.
```
lrytz, line 85: Example:

```scala
trait T1 { def x = 1 }
trait T2 { def x = 2 }
class A extends T1
object A extends T2
```

For `(a: A).x` the qualifier is `Type`, for `A.x` the qualifier is `Term`. The owner of both `x` is a type, but that's not what we want to express here.

eed3si9n (approved):
Thanks!

